PR #58, opened 2021-12-19
Code coverge profiling

Dentosal:
Adds per-instruction code coverge profiling, i.e. keeps track of all executed addresses

PR #65, opened 2021-12-29
Fix feature style and improve readme

vlopes11:
* Make the debugger attribute of the interpreter compatible with the
  format proposed by the profiler that avoids using a mock
  implementation.

* Modularize the profiler and split the feature requirements defined by
  `profiler-any` and `profiler-gas` with their respective modules.

* Avoid importing internal preludes related to the library to improve
  source readability.

* Include a changelog.

### DIFF
--- a/src/interpreter/gas.rs
+++ b/src/interpreter/gas.rs
@@ -65,10 +65,13 @@ impl<S> Interpreter<S> {
     pub(crate) fn gas_charge(&mut self, gas: Word) -> Result<(), RuntimeError> {
         let gas = !self.is_predicate() as Word * gas;
 
-        #[cfg(feature = "profile-gas")]
+        #[cfg(feature = "profiler-gas")]
         {
+            use crate::profiler::InstructionLocation;
+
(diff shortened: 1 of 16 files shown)


src/interpreter/gas.rs
```diff
@@ -65,10 +65,13 @@ impl<S> Interpreter<S> {
...
         let gas = !self.is_predicate() as Word * gas;
 
-        #[cfg(feature = "profile-gas")]
+        #[cfg(feature = "profiler-gas")]
         {
+            use crate::profiler::InstructionLocation;
+
             let gas_use = gas.min(self.registers[REG_CGAS]);
-            let location = self.current_location();
+            let location = InstructionLocation::from_vm_context(self);
```
Voxelot, line 73: Why does this need to change?

Voxelot:
I think some of these refactors will be incompatible with the incoming code coverage PR: #58 

We should merge that one before making overlapping structural changes.

vlopes11:
N/A

PR #74, opened 2022-01-20
Change Outputs + `TR` + `TRO`

### LINKED ISSUE #54: Transaction Output Mutation
According to the [specs](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_validity.md#script-execution), the VM needs to modify the transaction in-memory to finalize output amounts for Change and Variable outputs. 

> After execution, the following is extracted:
> 1. The transaction in-memory on VM termination is used as the final transaction which is included in the block.

- [ ] Variable outputs need to be updated on the fly as VM execution progresses
- [ ] The free balances are only moved into change outputs at the very end of execution, once.

Voxelot:
This PR updates `fuel-vm` to mutate the outputs of the processed transaction according to the specs. It also implements the opcodes `TR` and `TRO`.

- Unused gas is refunded into the base asset change output
- Change is appropriately adjusted based on coin and withdrawal outputs
- `TRO` deducts from the input amount / change output in an external context
- `TR` deducts from input amount / change amount in an external context
- Variable outputs are correctly set in both internal and external contexts when using `TRO`
  - The serialized tx in VM is updated
  - The referenced tx returned in the VM state transition object also includes the variable output updates
 

closes #54


Cargo.toml
```diff
@@ -15,24 +15,27 @@ dyn-clone = { version = "1.0", optional = true }
 fuel-asm = "0.1"
 fuel-merkle = "0.1"
 fuel-storage = "0.1"
-fuel-tx = "0.2"
+fuel-tx = { git = "https://github.com/fuellabs/fuel-tx.git", branch = "master"}
```
Voxelot, line 18: This will be updated back to crates io once published.

src/interpreter/gas.rs
```diff
@@ -85,6 +81,7 @@ impl<S> Interpreter<S> {
             Err(PanicReason::OutOfGas.into())
         } else {
             self.registers[REG_CGAS] -= gas;
+            self.registers[REG_GGAS] -= gas;
```
Voxelot, line 84: we actually weren't decrementing global gas before this 😬

Cargo.toml
adlerjohn: It's published, so should change back and test.

src/interpreter/contract.rs
```diff
@@ -44,6 +45,98 @@ where
         self.inc_pc()
     }
 
+    pub(crate) fn transfer(&mut self, a: Word, b: Word, c: Word) -> Result<(), RuntimeError> {
+        let (ax, overflow) = a.overflowing_add(32);
+        let (cx, of) = c.overflowing_add(32);
```
adlerjohn, line 50: Instead of 32 here, can we use something like `ContractId::LEN` and/or `Color::LEN`?

src/interpreter/contract.rs
```diff
@@ -44,6 +45,98 @@ where
...
+            Err(RuntimeError::Recoverable(PanicReason::ExpectedInternalContext)) => None,
+            // bubble up any other kind of errors
+            Err(e) => return Err(e),
+        };
+
+        if let Some(source_contract) = internal_context {
+            // debit funding source (source contract balance)
+            self.balance_decrease(&source_contract, &asset_id, amount)?;
+        } else {
+            // debit external funding source (i.e. UTXOs)
```
adlerjohn, line 84:
```suggestion
            // debit external funding source (i.e. free balance)
```

src/interpreter/executors/main.rs
```diff
@@ -84,13 +84,21 @@ where
...
+                }) {
+                    return Err(InterpreterError::Panic(PanicReason::ContractNotFound));
+                }
+
                 let offset = (VM_TX_MEMORY + Transaction::script_offset()) as Word;
 
                 self.registers[REG_PC] = offset;
                 self.registers[REG_IS] = offset;
-                self.registers[REG_GGAS] = self.tx.gas_limit();
-                self.registers[REG_CGAS] = self.tx.gas_limit();
```
adlerjohn, line 93: Why were these removed?

src/interpreter/initialization.rs
```diff
@@ -40,41 +39,32 @@ where
         self.push_stack(tx.id().as_ref())
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
-        let zeroes = &[0; MAX_INPUTS as usize * (Color::LEN + WORD_SIZE)];
-        let ssp = self.registers[REG_SSP] as usize;
-
-        self.push_stack(zeroes)
-            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
-
-        if tx.is_script() {
```
adlerjohn, line 49: Is there a reason to remove the `is_script()` checks here?

Voxelot: because otherwise we wouldn't have the ability to track free balance of the inputs for create tx, which means they wouldn't be able to return change.

src/interpreter/executors/main.rs
Voxelot: Because it's duplicating logic done in `initialization.rs`

Cargo.toml
adlerjohn: Alright, it's out https://crates.io/crates/fuel-tx/0.3.0

Voxelot: done

adlerjohn (approved):

PR #78, opened 2022-02-04
Implement Create Tx Storage Slot Initialization

### LINKED ISSUE #77: Storage Slot Initialization
Update contract id calculation based on initialized storage slots using SMT as per https://github.com/FuelLabs/fuel-specs/pull/276 

Blocked by: https://github.com/FuelLabs/fuel-tx/issues/74

Voxelot:
This PR updates the VM to initialize contract state when processing a create tx if initial storage values are provided.

closes #77


src/contract.rs
```diff
@@ -31,15 +31,35 @@ impl Contract {
         crypto::ephemeral_merkle_root(root)
     }
 
+    /// Calculate the root of the initial storage slots for this contract
+    pub fn initial_storage_root(storage_slots: &[StorageSlot]) -> Bytes32 {
```
adlerjohn, line 35: Hmmm. Should this be in `fuel-tx` instead? It's something that is entirely independent from the VM. I'm not against keeping it here for now, since there are a lot of things in the VM already that don't really belong and we need to do a big sweep one day.

Also, maybe add a TODO here to switch to SMT when ready?

adlerjohn (approved):

adlerjohn (changes requested):
Wait, I retract my approval. `Output::contract_created` needs to be changed to accept a storage root.

src/contract.rs
Voxelot: That would make sense, however the contract id calculation logic currently lives in the VM so I figured that would be the most logical place for now.

adlerjohn (approved):
Green checkmark means go

Issue #83, opened 2022-02-23
`GM` opcode bug


nfurfaro:
Given the following contract, the example test is passing. It should not be, as the assembly block which calls `gm` should return `true` in this case.

As noted by @adlerjohn on slack:
>If used from a contract that was called directly from a script, then
`Set $rA to true if parent is an external context, false otherwise.`

ref: https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/opcodes.md#gm-get-metadata

## Contract
```
contract;

abi AuthTesting {
    fn is_caller_external(gas_: u64, amount_: u64, color_: b256, value: bool) -> bool;
}

impl AuthTesting for Contract {
    fn is_caller_external(gas_: u64, amount_: u64, color_: b256, value: bool) -> bool {
        asm(r1) {
            gm r1 i1;
            r1: bool
        }
    }
}
```
## Test:
```
#[tokio::test]
async fn is_external_from_sdk() {
    abigen!(AuthContract, "test_artifacts/auth_testing_contract/src/abi-output.json");
    let salt = Salt::from([0u8; 32]);
    let compiled = Contract::compile_sway_contract("test_artifacts/auth_testing_contract", salt).unwrap();
    let (client, auth_id) = Contract::launch_and_deploy(&compiled).await.unwrap();
    let auth_instance = AuthContract::new(compiled, client);

    let result = auth_instance
        .is_caller_external(true)
        .call()
        .await
        .unwrap();

    assert_eq!(result.value, false);
}
```

Voxelot:
Looks like this boolean is the opposite of what it should be: https://github.com/FuelLabs/fuel-vm/blob/master/src/interpreter/metadata.rs#L60

we should be doing 
```
self.registers[ra] = (parent == 0) as Word;
```
instead of 
```
self.registers[ra] = (parent != 0) as Word;
```

PR #84, opened 2022-02-23
Fix GM opcode for returning the opposite result

Voxelot:
GM was returning the opposite of what it should've when checking if the caller was external. Instead of setting rA = true when the caller was external, we set rA = true when the caller was internal.

fixes: https://github.com/FuelLabs/fuel-vm/issues/83


nfurfaro (approved):
LGTM!

PR #90, opened 2022-03-07
Single-stepping support for the debugger

Dentosal:
Adds single-stepping support. When enabled, a pseudo-breakpoint is triggered before each instruction. This could be achieved by simply adding a breakpoint for each instruction manually, which wouldn't require any code changes here. However, this approach is easier to use, and has a smaller performance impact.

Also makes `Interpreter::transaction()` public as accessing it after iterating breakpoints isn't possible with the current API. My understanding is that making it public doesn't have any soundness implications.


src/interpreter.rs
```diff
@@ -92,6 +92,11 @@ impl<S> Interpreter<S> {
         self.registers[REG_FLAG] & 0x02 == 0x02
     }
 
+    /// The current transaction
+    pub fn transaction(&self) -> &Transaction {
```
vlopes11, line 96: Nit: const fn

src/state/debugger.rs
```diff
@@ -10,11 +10,23 @@ use std::collections::{HashMap, HashSet};
...
 pub struct Debugger {
+    /// Single-stepping mode triggers a breakpoint after each instruction
+    single_stepping: bool,
     breakpoints: HashMap<ContractId, HashSet<Word>>,
     last_state: Option<ProgramState>,
 }
 
 impl Debugger {
+    /// Get single-stepping mode
+    pub fn single_stepping(&self) -> bool {
```
vlopes11, line 21: Nit: const fn

src/state/debugger.rs
```diff
@@ -10,11 +10,23 @@ use std::collections::{HashMap, HashSet};
...
+        self.single_stepping
+    }
+
+    /// Set single-stepping mode
+    pub fn set_single_stepping(&mut self, single_stepping: bool) {
```
vlopes11, line 26: Maybe file a good first issue to create a debug builder since we are now adding parameters?

src/state/debugger.rs
```diff
@@ -47,17 +59,22 @@ impl Debugger {
         let contract = contract.copied().unwrap_or_default();
         let last_state = self.last_state.take();
 
+        let current = Breakpoint::raw(contract, pc);
```
vlopes11, line 62: Maybe we move this alloc inside the `if self.single_stepping`? This way we prevent additional alloc for every cycle, regardless of the configuration of the debug

vlopes11 (approved):
Some small optional suggestions, but LGTM overall

Voxelot (dismissed):

src/state/debugger.rs
Voxelot: it seems like this might actually be less allocations than before, since before we were making a breakpoint instance for each configured breakpoint in the set below using the identical pc settings.

src/state/debugger.rs
```diff
@@ -47,17 +59,22 @@ impl Debugger {
...
-            .map(|_| {
-                let breakpoint = Breakpoint::raw(contract, pc);
-
-                match last_state {
-                    Some(s) if s == breakpoint => DebugEval::Continue,
-                    _ => breakpoint.into(),
-                }
+            .map(|_| match last_state {
+                Some(s) if s == current => DebugEval::Continue,
+                _ => current.into(),
```
Voxelot, line 77: What's the point of these breakpoints if we just ignore their locations `map(|_| ..` here? Shouldn't we have something like this?
```suggestion
            .map(|location| match last_state {
                let breakpoint = Breakpoint::raw(contract, location);
                Some(s) if s == breakpoint => DebugEval::Continue,
                _ => breakpoint.into(),
```

Voxelot: Not going to hold up the PR over this since it was already this way before.

Dentosal: I'll make a separate PR to address this

adlerjohn (approved):

PR #93, opened 2022-03-17
Variable output bugfix

Voxelot:
Variable outputs need to be updated directly on the serialized transaction representation in VM memory for IVG purposes. However, there was a bug when writing the variable output changes such that the wrong offset was used, potentially overwriting freebalances.

This PR also makes change outputs consistent between the serialized tx and ref tx. Since if the serialized tx matches the VM referenced tx after post-execution, the tx won't need to be recloned and serialized later in the client for merkleizing.

### DIFF
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -9,8 +9,7 @@ use crate::storage::InterpreterStorage;
 
 use fuel_asm::{InstructionResult, PanicReason};
 use fuel_tx::{Input, Output, Receipt, Transaction};
-use fuel_types::bytes::SerializableVec;
-use fuel_types::Word;
+use fuel_types::{bytes::SerializableVec, Word};
 
 impl<S> Interpreter<S>
 where
@@ -163,7 +162,7 @@ where
         // refund remaining global gas
         let gas_refund = self.registers[REG_GGAS] * self.tx.gas_price();
         let revert = matches!(state, ProgramState::Revert(_));
-        self.update_change_amounts(gas_refund, revert)?;
+        self.finalize_outputs(gas_refund, revert)?;
 
         Ok(state)
     }
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -156,6 +156,7 @@ where
                 crypto::ephemeral_merkle_root(self.receipts().iter().map(|r| r.clone().to_bytes()))
             };
 
+            // TODO: also set this on the serialized tx in memory to keep serialized form consistent
             self.tx.set_receipts_root(receipts_root);
         }
 
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -156,6 +156,7 @@ where
             };
 
             // TODO: also set this on the serialized tx in memory to keep serialized form consistent
+            // https://github.com/FuelLabs/fuel-vm/issues/97
             self.tx.set_receipts_root(receipts_root);
         }
 
--- a/src/interpreter/internal.rs
+++ b/src/interpreter/internal.rs
@@ -181,7 +181,7 @@ impl<S> Interpreter<S> {
             ErrorKind::Other,
             "Invalid output index",
         )))?;
-        if std::mem::discriminant(tx_out) != std::mem::discriminant(&output) {
+        if core::mem::discriminant(tx_out) != core::mem::discriminant(&output) {
             return Err(RuntimeError::Halt(io::Error::new(
                 ErrorKind::Other,
                 "Invalid output type",
--- a/src/interpreter/internal.rs
+++ b/src/interpreter/internal.rs
@@ -166,10 +166,41 @@ impl<S> Interpreter<S> {
         }?;
 
         // update the local copy of the output
-        let mut output = Output::variable(owner_to_set, amount_to_set, asset_id_to_update);
+        let output = Output::variable(owner_to_set, amount_to_set, asset_id_to_update);
+        self.set_output(out_idx, output)
+    }
+
+    pub(crate) fn revert_variable_output(&mut self, out_idx: usize) -> Result<(), RuntimeError> {
+        if out_idx >= self.tx.outputs().len() {
+            return Err(PanicReason::OutputNotFound.into());
+        }
+
+        let reverted_variable_output = Output::variable(Default::default(), 0, Default::default());
+        self.set_output(out_idx, reverted_variable_output)
+    }
+
+    pub(crate) fn set_change_output(&mut self, out_idx: usize, update_amount: Word) -> Result<(), RuntimeError> {
+        let mut output = *self.tx.outputs().get(out_idx).ok_or(PanicReason::OutputNotFound)?;
+        if let Output::Change { amount, .. } = &mut output {
+            *amount = update_amount;
+        } else {
+            // TODO: make a panic reason for ExpectedChangeOutput
+            return Err(PanicReason::OutputNotFound.into());
+        }
+        self.set_output(out_idx, output)
+    }
+
+    /// update an output both in serialized form in-memory and on the referenced tx
+    fn set_output(&mut self, out_idx: usize, mut output: Output) -> Result<(), RuntimeError> {
+        // verify output type matches out_idx type
+        let tx_out = self.tx.outputs().get(out_idx).ok_or(PanicReason::OutputNotFound)?;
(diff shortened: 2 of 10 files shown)


src/interpreter/internal.rs
```diff
@@ -166,10 +166,41 @@ impl<S> Interpreter<S> {
...
 
         // update the local copy of the output
-        let mut output = Output::variable(owner_to_set, amount_to_set, asset_id_to_update);
+        let output = Output::variable(owner_to_set, amount_to_set, asset_id_to_update);
+        self.set_output(out_idx, output)
+    }
+
+    pub(crate) fn revert_variable_output(&mut self, out_idx: usize) -> Result<(), RuntimeError> {
+        if out_idx >= self.tx.outputs().len() {
+            return Err(PanicReason::OutputNotFound.into());
```
adlerjohn, line 175: Hmmm. Is `PanicReason` not for a VM panic? Presumably if this condition is met then the Rust code is buggy, no? The VM itself doesn't have the functionality to revert variable outputs.

This should be an `InterpreterError` rather than a `RuntimeError`.

src/interpreter/internal.rs
```diff
@@ -166,10 +166,41 @@ impl<S> Interpreter<S> {
...
+        self.set_output(out_idx, reverted_variable_output)
+    }
+
+    pub(crate) fn set_change_output(&mut self, out_idx: usize, update_amount: Word) -> Result<(), RuntimeError> {
+        let mut output = *self.tx.outputs().get(out_idx).ok_or(PanicReason::OutputNotFound)?;
+        if let Output::Change { amount, .. } = &mut output {
+            *amount = update_amount;
+        } else {
+            // TODO: make a panic reason for ExpectedChangeOutput
+            return Err(PanicReason::OutputNotFound.into());
```
adlerjohn, line 188: Ditto to above, the VM itself doesn't set change, that's technically done _after_ VM execution.

src/interpreter/internal.rs
```diff
@@ -166,10 +166,41 @@ impl<S> Interpreter<S> {
...
+        self.set_output(out_idx, output)
+    }
+
+    /// update an output both in serialized form in-memory and on the referenced tx
+    fn set_output(&mut self, out_idx: usize, mut output: Output) -> Result<(), RuntimeError> {
+        // verify output type matches out_idx type
+        let tx_out = self.tx.outputs().get(out_idx).ok_or(PanicReason::OutputNotFound)?;
+        if std::mem::discriminant(tx_out) != std::mem::discriminant(&output) {
+            // TODO: need a panic reason for a generic UnexpectedOutputType
+            return Err(PanicReason::OutputNotFound.into());
```
adlerjohn, line 199: Ditto

adlerjohn (changes requested):

src/interpreter/executors/main.rs
```diff
@@ -156,6 +156,7 @@ where
                 crypto::ephemeral_merkle_root(self.receipts().iter().map(|r| r.clone().to_bytes()))
             };
 
+            // TODO: also set this on the serialized tx in memory to keep serialized form consistent
```
vlopes11, line 159: Nit: issue?

src/interpreter/internal.rs
vlopes11: What @adlerjohn said seems to be correct. `PanicReason` is expected to be a result of valid transactions, and in this case it would be a result of an invalid and/or incorrect execution.

src/interpreter/internal.rs
```diff
@@ -166,10 +166,41 @@ impl<S> Interpreter<S> {
         }?;
 
         // update the local copy of the output
-        let mut output = Output::variable(owner_to_set, amount_to_set, asset_id_to_update);
+        let output = Output::variable(owner_to_set, amount_to_set, asset_id_to_update);
+        self.set_output(out_idx, output)
+    }
+
+    pub(crate) fn revert_variable_output(&mut self, out_idx: usize) -> Result<(), RuntimeError> {
```
vlopes11, line 173: This function is called only once `src/interpreter/post_execution.rs`.

Do we really need to generalize this behavior? As a result, we are performing the `out_idx` check more than once.

src/interpreter/executors/main.rs
Voxelot: added here #97

src/interpreter/internal.rs
```diff
@@ -166,10 +166,28 @@ impl<S> Interpreter<S> {
...
+    /// update an output both in serialized form in-memory and on the referenced tx
+    pub(crate) fn set_output(&mut self, out_idx: usize, mut output: Output) -> Result<(), RuntimeError> {
+        // verify output type matches out_idx type
+        // Use halt errors as these should already have been validated
+        let tx_out = self.tx.outputs().get(out_idx).ok_or(RuntimeError::Halt(io::Error::new(
+            ErrorKind::Other,
+            "Invalid output index",
+        )))?;
+        if std::mem::discriminant(tx_out) != std::mem::discriminant(&output) {
+            // TODO: need a panic reason for a generic UnexpectedOutputType
```
adlerjohn, line 182: Could use `PanicReason::ExpectedOutputVariable`?

Voxelot: oops that todo isn't needed anymore, since these are halting errors that should've been prevented earlier.

Voxelot:
I changed the interpreter panic errors into halting runtime errors, since the set_output method should only be called after other validations are performed.

adlerjohn (commented):
LGTM, but I want @vlopes11 to take a look

adlerjohn (approved):

src/interpreter/post_execution.rs
```diff
@@ -7,46 +7,44 @@ impl<S> Interpreter<S>
...
                 let refund = if *asset_id == AssetId::default() {
                     unused_gas_cost
                 } else {
                     0
                 };
 
-                if revert {
-                    *amount = init_balances[asset_id] + refund;
+                let final_amount = if revert {
+                    init_balances[asset_id] + refund
```
vlopes11, line 32: Yellow flag here: we will panic/halt (not talking about well-formed `RuntimeError`, but process crash) the VM if `asset_id` is not in `init_balances`. There is no hard guarantee (except for the transaction validity) that it will be there since the initial balances are calculated from the inputs, and here we are processing the outputs.

Do we want that?

src/interpreter/internal.rs
```diff
@@ -166,10 +166,27 @@ impl<S> Interpreter<S> {
...
+
+    /// update an output both in serialized form in-memory and on the referenced tx
+    pub(crate) fn set_output(&mut self, out_idx: usize, mut output: Output) -> Result<(), RuntimeError> {
+        // verify output type matches out_idx type
+        // Use halt errors as these should already have been validated
+        let tx_out = self.tx.outputs().get(out_idx).ok_or(RuntimeError::Halt(io::Error::new(
+            ErrorKind::Other,
+            "Invalid output index",
+        )))?;
+        if std::mem::discriminant(tx_out) != std::mem::discriminant(&output) {
```
vlopes11, line 181: nit: We don't use fully qualified names elsewhere in the code. In fact, This would make an eventual `no-std` port harder in the future because the imports are not centralized in the header of the file.
```suggestion
        if mem::discriminant(tx_out) != mem::discriminant(&output) {
```

Voxelot: there's going to be a lot of work to port to no-std anyways, since the VM relies so heavily on std::io in many places.

src/interpreter/post_execution.rs
Voxelot: Realized this doesn't validate for coin outputs. Will add safety checks for this.

Voxelot:
While addressing Victor's "yellow flag" issue with array access, I found there were several other potential errors. This led to me creating a new type of `VmValidationError` since the existing error types we have available weren't sufficient. I also converted some of the errors to use thiserror instead since I was lazy and didn't want to write all the formatting and conversion methods by hand.

adlerjohn (approved):
LGTM, though as before deferring to @vlopes11

vlopes11 (approved):
LGTM

The `VmValidationError` overlaps with `TxValidationError`. I think the latter should supersede the former. Can we file an issue to move later on the `VmValidationError` to `TxValidationError`? One of the hard assumptions we have in the VM is that a transaction is validated before execution; if we break that hard assumption and introduce validity errors inside of the VM, we might have some problems

For that reason they were kept in separated domains: the transaction should be atomic and its validation shouldn't depend on runtime.

PR #94, opened 2022-03-23
Address missing functionality in call opcode

Voxelot:
There were several blocking todo's preventing contract calls from working properly, this PR should address all of them.

### DIFF
--- a/src/interpreter/flow.rs
+++ b/src/interpreter/flow.rs
@@ -86,8 +86,8 @@ where
             return Err(PanicReason::ContractNotInInputs.into());
         }
 
-        // TODO validate external and internal context
-        // TODO update asset ID balance
+        // credit contract asset_id balance
+        self.balance_increase(call.to(), &asset_id, b)?;
 
         let mut frame = self.call_frame(call, asset_id)?;
 
@@ -104,9 +104,7 @@ where
 
         self.memory[self.registers[REG_FP] as usize..self.registers[REG_SP] as usize].copy_from_slice(stack.as_slice());
 
-        // TODO set balance for forward coins to $bal
-        // TODO set forward gas to $cgas
-
+        self.registers[REG_BAL] = b;
         self.registers[REG_PC] = self.registers[REG_FP].saturating_add(CallFrame::code_offset() as Word);
         self.registers[REG_IS] = self.registers[REG_PC];
         self.registers[REG_CGAS] = cmp::min(self.registers[REG_GGAS], d);
--- a/src/interpreter/flow.rs
+++ b/src/interpreter/flow.rs
@@ -80,6 +80,9 @@ where
 
         if self.is_external_context() {
             self.external_asset_id_balance_sub(&asset_id, b)?;
+        } else {
+            let source_contract = *self.internal_contract()?;
+            self.balance_decrease(&source_contract, &asset_id, b)?;
         }
 
         if !self.tx.input_contracts().any(|contract| call.to() == contract) {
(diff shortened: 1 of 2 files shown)


src/interpreter/flow.rs
```diff
@@ -104,9 +104,7 @@ where
 
         self.memory[self.registers[REG_FP] as usize..self.registers[REG_SP] as usize].copy_from_slice(stack.as_slice());
 
-        // TODO set balance for forward coins to $bal
-        // TODO set forward gas to $cgas
-
+        self.registers[REG_BAL] = b;
```
vlopes11, line 107: According to the specs, it should be `d`. Maybe typo there? https://github.com/FuelLabs/fuel-specs/blob/a65e14dfe5c0049ebf4f8e0b33b061c60927d7d9/specs/vm/opcodes.md?plain=1#L1098

src/interpreter/flow.rs
```diff
@@ -86,8 +86,8 @@ where
             return Err(PanicReason::ContractNotInInputs.into());
         }
 
-        // TODO validate external and internal context
-        // TODO update asset ID balance
+        // credit contract asset_id balance
```
vlopes11, line 89: Need to add the internal context part:
https://github.com/FuelLabs/fuel-specs/blob/a65e14dfe5c0049ebf4f8e0b33b061c60927d7d9/specs/vm/opcodes.md?plain=1#L1091

vlopes11 (changes requested):

src/interpreter/flow.rs
vlopes11: cc @adlerjohn

adlerjohn: Good point, looks like a typo in the specs!

adlerjohn: https://github.com/FuelLabs/fuel-specs/pull/292

src/interpreter/flow.rs
Voxelot: done and unit tested

Voxelot:
ah thought it was ready but there's one test regression, fix incoming

src/interpreter/flow.rs
```diff
@@ -80,14 +80,17 @@ where
 
         if self.is_external_context() {
             self.external_asset_id_balance_sub(&asset_id, b)?;
+        } else {
```
adlerjohn, line 83: Nit, but would this be a good candidate for a `match` instead of `if`/`else`?

adlerjohn (approved):

tests/contract.rs
```diff
@@ -230,6 +230,169 @@ fn mint_burn() {
...
+
+    // Ensure transfer tx processed correctly
+    assert!(!transfer_tx.should_revert());
+
+    // verify balance transfer occurred
+    let end_balance = test_context.get_contract_balance(&contract_id, &asset_id);
+    assert_eq!(end_balance, call_amount);
+
+    // verify balance register was set
+    assert_eq!(transfer_tx.receipts()[1].val().unwrap(), call_amount);
```
vlopes11, line 296: The logic seems correct, but its always safer to just `LOG` the register and then inspect the receipt. The `CALL` receipt is built with the arguments of the call and by itself isn't a guarantee that the register was properly set.

vlopes11 (approved):
LGTM

src/interpreter/flow.rs
Voxelot: Potentially, although the check for is_external is pretty lightweight itself (`if self.registers[REG_FP] == 0`)

PR #98, opened 2022-03-24
Externalized Validity Checks

### LINKED ISSUE #82: Externalize Validity Checks
Currently the VM performs several validity checks on a transaction in the init/run methods. This includes things like:

- Checking the input vs output amounts
- Verifying predicates (currently only done for create txs)
- Verifying contract outputs have corresponding inputs

Along with many other checks. Since these are internalized to the VM, the only way for the client to perform all these internal validations is to dry-run each tx. While feasible, it may be better if VM provided an interface for purely validating txs, which could be reused in many places such as both tx pool submission or block execution. This would make it easier to prevent blocks with invalid / unspendable utxos from ever being constructed.

Voxelot:
Previously, we didn't have a way to check if a tx was valid for block inclusion without also executing its' script. This introduces a new structure for `CheckedTransaction`'s, which allows a transaction to be validated with VM rules to ensure it's safe to include in a block, without requiring the actual script to be executed. The VM can accept these checked transactions when executing the script to avoid redundant validation as well. This will be useful for validating transactions inside the mempool.

closes #82

### DIFF
--- a/src/interpreter/initialization.rs
+++ b/src/interpreter/initialization.rs
@@ -2,24 +2,21 @@ use super::Interpreter;
 use crate::consts::*;
 use crate::context::Context;
 use crate::error::InterpreterError;
+use crate::interpreter::validation::CheckedTransaction;
 use crate::storage::InterpreterStorage;
 
-use fuel_asm::PanicReason;
 use fuel_tx::consts::MAX_INPUTS;
-use fuel_tx::{Input, Output, Transaction};
 use fuel_types::bytes::{SerializableVec, SizedBytes};
 use fuel_types::{AssetId, Word};
 use itertools::Itertools;
-use std::collections::HashMap;
 use std::io;
 
 impl<S> Interpreter<S>
 where
     S: InterpreterStorage,
 {
-    pub(crate) fn init(&mut self, mut tx: Transaction) -> Result<(), InterpreterError> {
-        tx.validate_without_signature(self.block_height() as Word)?;
-        tx.precompute_metadata();
+    pub(crate) fn init(&mut self, checked_tx: CheckedTransaction) -> Result<(), InterpreterError> {
+        let mut tx = checked_tx.tx;
 
         self.block_height = self.storage.block_height().map_err(InterpreterError::from_io)?;
         self.context = Context::from(&tx);
@@ -40,7 +37,10 @@ where
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
         // Set initial unused balances
-        let free_balances = Self::initial_free_balances(&tx)?;
+        let free_balances = checked_tx.balances;
+        // Clear balance indexes in case vm is being re-initialized
+        self.unused_balance_index.clear();
+        // Put free balances into vm memory
         for (asset_id, amount) in free_balances.iter().sorted_by_key(|i| i.0) {
             // push asset ID
             self.push_stack(asset_id.as_ref())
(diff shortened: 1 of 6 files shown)


src/interpreter/validation.rs
```diff
@@ -0,0 +1,85 @@
+use crate::error::InterpreterError;
+use crate::interpreter::Interpreter;
+use fuel_asm::PanicReason;
+use fuel_tx::{Input, Output, Transaction};
+use fuel_types::{AssetId, Word};
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct CheckedTransaction {
```
vlopes11, line 9: Looks like this code belongs to `fuel-tx` rather than `fuel-vm`. In fact, we already have a data structure to contain transaction [Metadata](https://docs.rs/fuel-tx/0.6.0/fuel_tx/struct.Metadata.html).

I believe we should insert this balances map there.

src/interpreter/internal.rs
```diff
@@ -224,6 +224,8 @@ mod tests {
             vec![vec![].into()],
         );
 
+        let tx = Interpreter::<()>::check_transaction(tx, 0).unwrap();
```
vlopes11, line 227: Consider using `expect` with a message that clarifies what would be the expected behavior here

src/interpreter/validation.rs
Voxelot: The reason I don't think this checked transaction wrapper type should be outside of the VM is based on my proposal in the related issue:

> We could make it so the only way to construct these wrapped transactions is by running the validation method of the VM. Then the interpreter init methods could be modified to only accept these checked transactions.

If CheckedTransaction was in a different crate, it would hold no significance or guarantees for the VM. Anyone could spoof a CheckedTransaction and lead to undefined behavior in the VM.

src/interpreter/initialization.rs
```diff
@@ -40,7 +37,10 @@ where
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
         // Set initial unused balances
-        let free_balances = Self::initial_free_balances(&tx)?;
+        let free_balances = checked_tx.balances;
+        // Clear balance indexes in case vm is being re-initialized
+        self.unused_balance_index.clear();
```
adlerjohn, line 42: Hmmm. Is this necessary? Predicates can't actually modify the free balances since they can't send coins anywhere. Only scripts will, and scripts are only executed once per tx.

src/interpreter/validation.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// the validated transaction
+    pub(crate) tx: Transaction,
+    /// the initial free balances of the transaction
+    pub(crate) balances: HashMap<AssetId, Word>,
```
adlerjohn, line 13: Nit, but maybe an ordered map like a b tree would be better?

Voxelot: Shouldn't be a big deal since we sort this while setting up freebalances in the VM memory.

Voxelot:
blocked by: #93

src/interpreter/validation.rs
Voxelot: based on your comment here: https://github.com/FuelLabs/fuel-vm/pull/93#pullrequestreview-921424472

I'm coming around to the idea that we could actually move all validations the VM needs into fuel-tx, and that this isn't needed anymore. I'm going to close this PR and focus on cleaning up the validation tech debt.

PR #100, opened 2022-03-26
Implement MOVI opcode

Voxelot:
Implement MOVI opcode and replace non-idiomatic usages of `ADDI r 0 imm`.

This unblocks support for transactions with 255 inputs, since their script data offsets are now too large to fit in an ADDI imm12.

blocked by: https://github.com/FuelLabs/fuel-asm/pull/53

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,3 +83,6 @@ required-features = ["random"]
 name = "test-predicate"
 path = "tests/predicate.rs"
 required-features = ["random"]
+
+[patch.crates-io]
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -83,3 +83,6 @@ required-features = ["random"]
 name = "test-predicate"
 path = "tests/predicate.rs"
 required-features = ["random"]
+
+[patch.crates-io]
```
Voxelot, line 87: will fix this patch once https://github.com/FuelLabs/fuel-asm/pull/53 is merged and released.

adlerjohn (approved):

PR #105, opened 2022-04-20
update vm to use newer ScriptExecutionResult

Voxelot:
- Update to fuel-tx 0.8
- Use ScriptExecutionResult instead of InstructionResult in backtraces


adlerjohn (approved):
Looks fine to me, but would like @vlopes11 to confirm.

vlopes11 (approved):
LGTM

PR #119, opened 2022-05-16
Profiler fixes

Dentosal:
Currently the data produced by the profiler cannot be serialized to JSON using Serde, as the data model requires object keys to be strings. Rust doesn't detect this at compile time, and the data was already usable in most binary formats like bincode. This PR makes it possible to serialize the data as JSON, but unfortunately Serde doesn't allow different formats for binary files, so this also increases size of bincode-encoded profile files.

This PR also adds `Send + Sync` trait bounds to the profile receiver, so that it can be used in multithreaded contexts, like in `async` functions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ required-features = ["random", "profile-coverage"]
 [[test]]
 name = "test-serde-profile"
 path = "tests/serde_profile.rs"
-required-features = ["random", "profile-coverage", "profile-gas", "serde-types"]
+required-features = ["random", "profile-coverage", "profile-gas", "serde"]
 
 [[test]]
 name = "test-encoding"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ rand = { version = "0.8", optional = true }
 [dev-dependencies]
 fuel-tx = { version = "0.10", features = ["random"] }
 fuel-vm = { path = ".", default-features = false, features = ["test-helpers"] }
+serde_json = "1.0"
 
 [features]
 debug = []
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -64,6 +65,11 @@ name = "test-code-coverage"
 path = "tests/code_coverage.rs"
 required-features = ["random", "profile-coverage"]
 
+[[test]]
+name = "test-serde-profile"
+path = "tests/serde_profile.rs"
+required-features = ["random", "profile-coverage", "profile-gas", "serde-types"]
```
Voxelot, line 71:
```suggestion
required-features = ["random", "profile-coverage", "profile-gas", "serde"]
```

src/profiler.rs
```diff
@@ -11,14 +11,68 @@ use crate::prelude::*;
...
+    {
+        if let Some(ctx) = self.context {
+            serializer.serialize_str(&format!("{}:{}", ctx, self.offset))
+        } else {
+            serializer.serialize_str(&format!("{}", self.offset))
+        }
+    }
+}
+
+#[cfg(feature = "serde-types")]
```
Voxelot, line 35:
```suggestion
#[cfg(feature = "serde")]
```

src/profiler.rs
```diff
@@ -11,14 +11,68 @@ use crate::prelude::*;
...
+struct InstructionLocationVisitor;
+
+#[cfg(feature = "serde-types")]
```
Voxelot, line 38:
```suggestion
#[cfg(feature = "serde")]
```

src/profiler.rs
```diff
@@ -11,14 +11,68 @@ use crate::prelude::*;
...
+            let offset = r.parse().unwrap();
+            InstructionLocation { context, offset }
+        } else {
+            let offset = value.parse().unwrap();
+            InstructionLocation { context: None, offset }
+        })
+    }
+}
+
+#[cfg(feature = "serde-types")]
```
Voxelot, line 66:
```suggestion
#[cfg(feature = "serde")]
```

Voxelot (approved):

Issue #121, opened 2022-05-19
Arithmetic ops don't set `$of` to overflow value, but overflow state


adlerjohn:
This line https://github.com/FuelLabs/fuel-vm/blob/9f67738c684ba72c5c975808170c72c3fae5bbca/src/interpreter/executors/instruction.rs#L169 sets the `$of` register to the second return value of [`overflowing_mul`](https://doc.rust-lang.org/std/primitive.u64.html#method.overflowing_mul), which is a Boolean indicating whether an overflow occurred or not.

The specs actually say that the overflow _value_ should be placed into `$of` in the case of overflow, no simple a Boolean.

At the very least this needs to be done for add, subtract, multiply, divide, and mod. Pow can be left as-is.

nfurfaro:
>At the very least this needs to be done for add, subtract, multiply, divide, and mod. Pow can be left as-is.

We should add the shifting ops to this list , or else change the spec.

adlerjohn:
Right shift can't overflow. Left shift the specs will change, so I'm anticipation of that there's no need to capture the overflow value.

Issue #104, opened 2022-04-17
Unclear output when using logical shift opcodes


nfurfaro:
While trying to use the `sll` opcode, I'm seeing output that doesn't match my expectations.
For my use case, I want to use the `sll` opcode, and recover any overflow value placed in `$of`. To do so I've created a function `shift_left_with_overflow()` which returns both the shifted value and the value of the `$of` register from an asm block.
The following example script can be run to observe what is currently happening, and the flags can be modified to test different behaviour. 
My assumption was that using the `flag 2` opcode would allow the overflow to be placed in `$of` rather than causing a VM panic, but atm it seems that this logic is not yet implemented (meaning overflows never cause panics,  the overflow is _sometimes_ places in `$of`, and the shifted value _sometimes_ demonstrates wrapping.

```
script;

use std::assert::assert;
use std::chain::log_u64;

const FLAG= 2;

pub fn shift_left_with_overflow(word: u64, shift_amount: u64) -> (u64, u64) {
    let mut output = (0, 0);
    let (shifted, overflow) = asm(out: output, r1: word, r2: shift_amount, r3, r4: FLAG) {
       flag r4;        // set the flag to allow overflow without panic
       sll r3 r1 r2;   // shift 'word' left 'shift_amount' and put result in r3
       sw out r3 i0;   // store the word at r4 in output + 0 bytes
       sw out of i1;   // store the word at r4 in output + 0 bytes
       out: (u64, u64) // return both values
    };

    (shifted, overflow)
}


fn main() -> bool {

    let max_u64 = 18_446_744_073_709_551_615;

    let (shifted, overflow) = shift_left_with_overflow(1, 1);
    // log_u64(shifted);
    // log_u64(overflow);
    assert(shifted == 2);
    assert(overflow == 0);

    let (shifted_2, overflow_2) = shift_left_with_overflow(max_u64, 1);
    assert(shifted_2 == max_u64 - 1);
    assert(overflow_2 == 0); // I would expect this to be 1 !?

    let (shifted_3, overflow_3) = shift_left_with_overflow(1, 63);
    assert(shifted_3 == 9223372036854775808); // (1*2^63)
    assert(overflow_3 == 0);

    let (shifted_4, overflow_4) = shift_left_with_overflow(1, 64);
    assert(shifted_4 == 1); // I would expect this to be 0 !?
    assert(overflow_4 == 1);

[...]

PR #122, opened 2022-05-19
Capture overflow value instead of overflow state into `$of`

Voxelot:
fixes: https://github.com/FuelLabs/fuel-vm/issues/121

Uses u128 math for all the instructions that need to treat the overflow register as the high bits of a u128. This will unblock the stdlib impl of u128 types.

This includes:

- add
- addi
- mul
- muli
- sll
- slli
- srl
- srli
- sub
- subi

The wording of the specs was somewhat confusing for srl & srli, since a logical shift right can't underflow. However, the specs describe the same `$of` behavior as other opcodes which do have under/overflow.

### DIFF
--- a/src/interpreter/alu.rs
+++ b/src/interpreter/alu.rs
@@ -9,23 +9,22 @@ impl<S> Interpreter<S> {
     /// Stores the overflowed wrapped value into REG_OF
     pub(crate) fn alu_capture_overflow<F, B, C>(&mut self, ra: RegisterId, f: F, b: B, c: C) -> Result<(), RuntimeError>
     where
-        F: FnOnce(B, C) -> (Word, bool),
+        F: FnOnce(B, C) -> (u128, bool),
     {
         Self::is_register_writable(ra)?;
 
-        let (result, overflow) = f(b, c);
+        let (result, _overflow) = f(b, c);
 
-        if overflow && !self.is_wrapping() {
+        if result > Word::MAX as u128 && !self.is_wrapping() {
             return Err(PanicReason::ArithmeticOverflow.into());
         }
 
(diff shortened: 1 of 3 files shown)


adlerjohn (commented):
This wasn't documented very nicely, but pow should simply set `$of` to `1` if the result would overflow, i.e. should have Boolean overflow. Right shift shouldn't set `$of` ever. As for left shift, now that I think about it it shouldn't set `$of` ever either, since obviously if you're left shifting even by one bit it'll overflow so setting `$of` isn't useful. Ref: https://github.com/FuelLabs/fuel-vm/issues/104#issuecomment-1120266633

Is it going to be a lot of work to make those changes?

src/interpreter/alu.rs
```diff
@@ -6,7 +6,31 @@ use fuel_asm::PanicReason;
...
+        if result > Word::MAX as u128 && !self.is_wrapping() {
+            return Err(PanicReason::ArithmeticOverflow.into());
+        }
+
+        // set the OF register to high bits of the u128 result
+        self.registers[REG_OF] = (result >> 64) as u64;
+        self.registers[REG_ERR] = 0;
+
+        // set the return value to the low bits of the u128 result
+        self.registers[ra] = (result & Word::MAX as u128) as u64;
```
vlopes11, line 27: If we just truncate to `as u64`, we achieve the same effect.

vlopes11 (dismissed):
looks good after @adlerjohn comments are addressed

Voxelot:
Already handled the pow cases as you described, but forgot to add tests. Will disable overflow behavior for shl and shr.

adlerjohn (approved):

Issue #126, opened 2022-05-23
bug: duplicate `Revert` receipts in reverting transactions


vnepveu:
In some transactions that revert, there are two `Revert` receipts sent back by the VM. A repro is available here: 
https://github.com/FuelLabs/fuels-rs/tree/vnepveu/repro-two-revert-receipts.

PR #127, opened 2022-05-24
Duplicate revert receipt fix

Voxelot:
fixes: #126 

Expected Behavior:
The expected behavior for a revert is to immediately cease execution of opcodes, and only include one revert receipt.

Actual Behavior:
When a contract call reverted, the VM appended two revert receipts.

Summary:
The root cause is that while `run_call` correctly returned a revert state, this result was then discarded by the parent stack and not propagated. Since revert doesn't increase $pc, the same opcodes would run again on the parent stack (callee) causing the receipt to be appended twice.

The fix simply checks the result of `call` and returns with the Revert state if needed.


adlerjohn (approved):

vlopes11 (approved):
LGTM

PR #129, opened 2022-05-24
Update predicates implementation

vlopes11:


src/interpreter/initialization.rs
```diff
@@ -38,26 +38,39 @@ where
...
-            self.push_stack(asset_id.as_ref())
-                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
-            // stack position
-            let asset_id_offset = self.registers[REG_SSP] as usize;
-            self.unused_balance_index.insert(*asset_id, asset_id_offset);
-            // push spendable amount
-            self.push_stack(&amount.to_be_bytes())
-                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
-        }
+        let predicate = !tx.is_script();
```
Voxelot, line 41: this doesn't make sense, predicate behavior is totally unrelated to the transaction type

vlopes11: I didn't review that part yet, its wip

src/interpreter/initialization.rs
```diff
@@ -38,26 +36,38 @@ where
...
+
         // zero out remaining unused balance types
-        for _i in free_balances.len()..(MAX_INPUTS as usize) {
-            self.push_stack(&[0; AssetId::LEN + WORD_SIZE])
-                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
-        }
+        let free_balances = MAX_INPUTS as Word - free_balances;
+        let free_balances = free_balances * (AssetId::LEN + WORD_SIZE) as Word;
+
+        self.reserve_stack(free_balances)?;
```
Voxelot, line 67:
```suggestion
        let unused_balances = MAX_INPUTS as Word - free_balances;
        let unused_balances = unused_balances * (AssetId::LEN + WORD_SIZE) as Word;

        self.reserve_stack(unused_balances)?;
```

src/interpreter/executors/predicate.rs
```diff
@@ -1,26 +1,28 @@
...
+use std::ops::Div;
+
+impl<S> Interpreter<S> {
     pub(crate) fn verify_predicate(&mut self, predicate: &MemoryRange) -> Result<ProgramState, InterpreterError> {
-        // TODO initialize VM with tx prepared for sign
-        // TODO execute should not overflow predicate boundaries. Need to check
-        // internally if a Jump instruction decrements $pc, or if $pc overflows
-        // `end`
-        let (start, _end) = predicate.boundaries(self);
+        debug_assert!(self.is_predicate());
```
Voxelot, line 13: This should be an interpreter error imo

src/interpreter/executors/predicate.rs
```diff
@@ -43,4 +45,338 @@ where
...
+            }
+
+            OpcodeRepr::ORI => {
+                self.gas_charge(GAS_ORI)?;
+                self.alu_set(ra, b | imm)?;
+            }
+
+            OpcodeRepr::SLL => {
+                self.gas_charge(GAS_SLL)?;
+                self.alu_set(ra, b.checked_shl(c as u32).unwrap_or_default())?;
```
Voxelot, line 188: unfortunate that we have to copy the instruction execution implementation 😕 

This creates a possibility for a drift in behavior that could be hard to detect or test for.

src/interpreter/executors/predicate.rs
Voxelot: actually nm, just realized this is internal and only called after the inputs should've been checked

src/interpreter/executors/predicate.rs
vlopes11: Agree, but this change is more delicate since we have different storage backends for these executions. Can we leave this for the next week as debt so I come with a more complete solution that won't duplicate code?

adlerjohn (commented):
CI failing

src/interpreter/executors/predicate.rs
Voxelot: yeah makes sense, we just need to address this before bit-rot sets in

adlerjohn:
The gas limit does matter since each instruction uses a certain amount of gas.

Voxelot:
https://github.com/FuelLabs/fuel-tx/pull/134

Voxelot (approved):

PR #131, opened 2022-05-25
bump to v0.12.0

Voxelot:


adlerjohn (commented):
Why are we skipping v0.11?

Voxelot:
ah good point, v0.11 wasnt released yet!

PR #140, opened 2022-06-06
Update VM memory for receipts root after append

### LINKED ISSUE #125: Have the VM update `receiptsRoot` before the script completes execution
It looks like we currently only set the receipts root in post script execution. The desired behaviour is probably to update that value before then so that contract methods can get it given what receipts have been generated so far.

vlopes11:
The VM memory section for the receipts root was updated only after the
script execution.

It is desirable to have it updated on the fly so the scripts can consume
the generated receipts root.

Closes #125

### DIFF
--- a/src/interpreter/contract.rs
+++ b/src/interpreter/contract.rs
@@ -86,14 +86,16 @@ where
         // credit destination contract
         self.balance_increase(&destination, &asset_id, amount)?;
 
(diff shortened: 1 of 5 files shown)


vlopes11:
Blocked by https://github.com/FuelLabs/fuel-tx/pull/136

Voxelot (approved):

PR #148, opened 2022-06-13
Add gas factor to gas cost and refund calculation

vlopes11:
Related issue: https://github.com/FuelLabs/fuel-tx/pull/141
Closes https://github.com/FuelLabs/fuel-specs/issues/342


vlopes11:
Blocked by https://github.com/FuelLabs/fuel-tx/pull/144

src/interpreter/constructors.rs
```diff
@@ -31,7 +31,7 @@ impl<S> Interpreter<S> {
             #[cfg(feature = "profile-any")]
             profiler: Profiler::default(),
             unused_balance_index: Default::default(),
-            consensus_parameters: params,
+            params: params,
```
Voxelot, line 34: [nit] We _should_ be able to avoid duplicating the assignment vars if they are the same:

```suggestion
            params,
```

src/interpreter/blockchain.rs
```diff
@@ -40,7 +39,7 @@ where
         // Validate arguments
         if ssp + length_to_copy_unpadded > hp
             || (id_addr + ContractId::LEN) as u64 > VM_MAX_RAM
-            || length_to_copy_unpadded > CONTRACT_MAX_SIZE.min(MEM_MAX_ACCESS_SIZE)
+            || length_to_copy_unpadded > self.params.contract_max_size.min(MEM_MAX_ACCESS_SIZE)
```
Voxelot, line 42: good catch

src/interpreter/constructors.rs
vlopes11: true, forgot that

src/interpreter/initialization.rs
```diff
@@ -15,15 +14,12 @@ use std::io;
...
-        &mut self,
-        predicate: bool,
-        block_height: u32,
-        mut tx: Transaction,
-        params: ConsensusParameters,
-    ) -> Result<(), InterpreterError> {
-        tx.validate_without_signature(self.block_height() as Word, &params)?;
-        tx.precompute_metadata();
+    pub fn init(&mut self, predicate: bool, block_height: u32, tx: Transaction) -> Result<(), InterpreterError> {
+        self.tx = tx;
```
Voxelot, line 18: nice cleanup

src/interpreter/initialization.rs
```diff
@@ -109,35 +104,32 @@ impl<S> Interpreter<S> {
...
-            let gas_cost = tx
-                .gas_limit()
-                .checked_mul(tx.gas_price())
-                .ok_or(ValidationError::ArithmeticOverflow)?;
-
-            // add up total amount of required fees
-            let total_fee = byte_balance
-                .checked_add(gas_cost)
-                .ok_or(ValidationError::ArithmeticOverflow)?;
+            // calculate the fee with used witness bytes + gas limit
```
Voxelot, line 107: [nit] Witnesses aren't included in the metered byte size, otherwise validators could add an arbitrary number of witnesses to each tx to max out the fee.

```suggestion
            // calculate the fee with used metered bytes + gas limit
```

src/interpreter/initialization.rs
```diff
@@ -109,35 +104,32 @@ impl<S> Interpreter<S> {
...
+            let bytes = self
+                .tx
+                .byte_price()
+                .saturating_mul(self.tx.metered_bytes_size() as Word);
+            let bytes = (bytes as f64 / factor).ceil() as Word;
+
+            let gas = self.tx.gas_price().saturating_mul(self.tx.gas_limit()) as f64;
+            let gas = (gas / factor).ceil() as Word;
+
+            let fee = bytes.saturating_add(gas);
```
Voxelot, line 119: [nit] better to play it safe with overflow error imo

```suggestion
            let fee = bytes.checked_add(gas).ok_or(ValidationError::ArithmeticOverflow)?;
```

src/interpreter/initialization.rs
```diff
@@ -109,35 +104,32 @@ impl<S> Interpreter<S> {
...
-            let total_fee = byte_balance
-                .checked_add(gas_cost)
-                .ok_or(ValidationError::ArithmeticOverflow)?;
+            // calculate the fee with used metered bytes + gas limit
+            let factor = self.params.gas_price_factor as f64;
+
+            let bytes = self
+                .tx
+                .byte_price()
+                .saturating_mul(self.tx.metered_bytes_size() as Word);
```
Voxelot, line 113:
```suggestion
                .checked_mul(self.tx.metered_bytes_size() as Word).ok_or(ValidationError::ArithmeticOverflow)?;
```

src/interpreter/executors/main.rs
```diff
@@ -211,9 +211,13 @@ where
             self.tx.set_receipts_root(receipts_root);
         }
 
-        // refund remaining global gas
-        let gas_refund = self.registers[REG_GGAS] * self.tx.gas_price();
+        // used bytes are non-refundable
+        let factor = self.params.gas_price_factor as f64;
+        let gas_refund = self.tx.gas_price().saturating_mul(self.registers[REG_GGAS]) as f64;
```
Voxelot, line 216:
```suggestion
        let gas_refund = self.tx.gas_price().checked_mul(self.registers[REG_GGAS]).ok_or(ValidationError::ArithmeticOverflow)? as f64;
```

Voxelot, line 216: saturating is fine here since we already validated the full g_gas deduction during init

tests/validation.rs
```diff
@@ -150,17 +156,14 @@ fn bytes_fee_cant_overflow() {
...
 
-    let mut interpreter = Interpreter::with_memory_storage();
-    let result = interpreter.transact(transaction);
-    assert!(matches!(
-        result,
-        Err(InterpreterError::ValidationError(ValidationError::ArithmeticOverflow))
-    ));
+    Interpreter::with_memory_storage()
+        .transact(transaction)
+        .expect("gas factor won't allow overflow");
```
Voxelot, line 161: is this true if someone sets gas factor to 1?

tests/validation.rs
```diff
@@ -200,10 +200,7 @@ fn total_fee_cant_overflow() {
...
 }
```
Voxelot, line 206: We should have a test to verify that we properly round-up during the initial deduction, and round down for the refund

Voxelot: We may also want a test that uses non-default gas price factors, to verify that the runtime params are being respected.

src/interpreter/executors/main.rs
```diff
@@ -211,9 +211,17 @@ where
             self.tx.set_receipts_root(receipts_root);
         }
 
-        // refund remaining global gas
-        let gas_refund = self.registers[REG_GGAS] * self.tx.gas_price();
+        // used bytes are non-refundable
```
adlerjohn, line 214: ?
```suggestion
        // used gas is non-refundable
```

adlerjohn (commented):
It would be preferable to move the refactoring changes to a separate PR.

src/interpreter/executors/main.rs
Voxelot: I think this comment is explaining why there's no corresponding calculation for bytes when processing the refund?

adlerjohn: Hmmm. Well if that's the case then it should be worded clearer, that's for sure 😂

vlopes11: i'll pick different words, but its not really the gas there. its the price per byte * |witness| that is charged from the base asset, that is non-refundable

tests/validation.rs
vlopes11: Good catch, updated

tests/validation.rs
vlopes11: Added in `tests/gas_factor.rs`

src/interpreter/executors/main.rs
vlopes11: https://github.com/FuelLabs/fuel-vm/pull/148/commits/3f2bf04821fdda979d48cc15c4f32d7b2dd55057

adlerjohn (commented):
Nothing stands out to me, will defer to @Voxelot for final review.

Voxelot (approved):
lgtm, thanks for adding the tests and doing all this cleanup

Issue #142, opened 2022-06-06
[EXP/EXPI] Set `$rA` to zero if overflow occurs


Voxelot:
related to https://github.com/FuelLabs/fuel-specs/pull/335

With EXP/EXPI, we currently set the result register to the wrapped value rust returns from `overflowing_pow`. We should set the result to zero instead.

PR #151, opened 2022-06-21
Set $rA to zero on EXP/EXPI overflow

Dentosal:
Fixes https://github.com/FuelLabs/fuel-vm/issues/142

### DIFF
--- a/src/interpreter/executors/instruction.rs
+++ b/src/interpreter/executors/instruction.rs
@@ -111,17 +111,11 @@ where
             OpcodeRepr::EXP => {
                 self.gas_charge(GAS_EXP)?;
                 self.alu_boolean_overflow(ra, Word::overflowing_pow, b, c as u32)?;
-                if self.registers[REG_OF] != 0 {
-                    self.registers[ra] = 0;
-                }
             }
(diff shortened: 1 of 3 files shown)


src/interpreter/executors/instruction.rs
```diff
@@ -111,11 +111,17 @@ where
             OpcodeRepr::EXP => {
                 self.gas_charge(GAS_EXP)?;
                 self.alu_boolean_overflow(ra, Word::overflowing_pow, b, c as u32)?;
+                if self.registers[REG_OF] != 0 {
```
vlopes11, line 114: Shouldn't this be a behavior of `alu_*_overflow` instead? Also, we don't put concrete implementations here. If exp is a special case of `allu`, then we need to create a new `alu_*` function and call that instead

vlopes11 (changes requested):

src/interpreter/executors/instruction.rs
Dentosal: Fixed

vlopes11 (approved):
LGTM!

PR #154, opened 2022-06-22
fix block height when performing tx validation

Voxelot:
The VM was passing the default block height to tx validation before it was set. 

This PR updates the init logic to use the height passed in to the init function instead of relying on when the block height is set on the interpreter.

fixes: https://github.com/FuelLabs/fuel-core/issues/432

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ repository = "https://github.com/FuelLabs/fuel-vm"
 description = "FuelVM interpreter."
 
 [dependencies]
+anyhow = "1.0"
 dyn-clone = { version = "1.0", optional = true }
 fuel-asm = "0.5"
 fuel-crypto = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,8 @@ tracing = "0.1"
 rand = { version = "0.8", optional = true }
 
 [dev-dependencies]
-fuel-tx = { version = "0.13", features = ["random"] }
+fuel-crypto = { version = "0.5", features = ["random"] }
+fuel-tx = { version = "0.13", features = ["random", "internals", "builder"] }
 fuel-vm = { path = ".", default-features = false, features = ["test-helpers"] }
 serde_json = "1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/FuelLabs/fuel-vm"
 description = "FuelVM interpreter."
 
 [dependencies]
-anyhow = "1.0"
+anyhow = { version = "1.0", optional = true }
 dyn-clone = { version = "1.0", optional = true }
 fuel-asm = "0.5"
 fuel-crypto = "0.5"
@@ -40,7 +40,7 @@ profile-coverage = ["profile-any"]
 profile-any = ["dyn-clone"] # All profiling features should depend on this
 random = ["fuel-types/random", "fuel-tx/random", "rand"]
 serde = ["dep:serde", "fuel-asm/serde", "fuel-types/serde", "fuel-tx/serde"]
-test-helpers = ["random"]
+test-helpers = ["random", "dep:anyhow"]
 
 [[test]]
 name = "test-backtrace"
--- a/src/interpreter/initialization.rs
+++ b/src/interpreter/initialization.rs
@@ -17,8 +17,7 @@ impl<S> Interpreter<S> {
     pub fn init(&mut self, predicate: bool, block_height: u32, tx: Transaction) -> Result<(), InterpreterError> {
         self.tx = tx;
 
-        self.tx
-            .validate_without_signature(self.block_height() as Word, &self.params)?;
+        self.tx.validate_without_signature(block_height as Word, &self.params)?;
         self.tx.precompute_metadata();
 
         self.block_height = block_height;
--- a/src/memory_client.rs
+++ b/src/memory_client.rs
@@ -3,6 +3,7 @@
 use crate::backtrace::Backtrace;
 use crate::storage::MemoryStorage;
 use crate::transactor::Transactor;
+use anyhow::{anyhow, Context};
 
 use fuel_tx::{ConsensusParameters, Receipt, Transaction};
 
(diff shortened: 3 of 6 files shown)


src/interpreter/initialization.rs
```diff
@@ -17,8 +17,7 @@ impl<S> Interpreter<S> {
     pub fn init(&mut self, predicate: bool, block_height: u32, tx: Transaction) -> Result<(), InterpreterError> {
         self.tx = tx;
 
-        self.tx
-            .validate_without_signature(self.block_height() as Word, &self.params)?;
+        self.tx.validate_without_signature(block_height as Word, &self.params)?;
```
vlopes11, line 20: Its odd that we are taking block height as argument here. It should come from the storage, and definitely isn't an arbitrary parameter. Maybe its like this for some outdated compatibility reasons?

We should remove that argument and use the interpreter call, instead.

https://github.com/FuelLabs/fuel-vm/blob/70243e4da9acf1b03f8d4f34ccdf129e3b20c4b0/src/storage/interpreter.rs#L25

`init` is meant to be private. It was moved to public just to speed up some other implementations, but we shouldn't expose these sensible arguments to the user.

src/memory_client.rs
```diff
@@ -60,20 +60,18 @@ impl<'a> MemoryClient<'a> {
...
-            .expect("MemoryStorage implements `Infallible` as storage error. This means panic should be unreachable.");
-
-        if state.should_revert() {
-            self.transactor.as_mut().revert();
+        if let Ok(state) = self.transactor.result() {
+            if state.should_revert() {
+                self.transactor.as_mut().revert();
+            } else {
+                self.transactor.as_mut().commit();
+            }
```
vlopes11, line 68: The memory storage implements infallible as error; its pointless to check the outcome of the result. Also, isn't this outside the scope of the problem?

src/storage/memory.rs
```diff
@@ -77,6 +77,11 @@ impl MemoryStorage {
         self.memory = self.transacted.clone();
         self.persisted = self.transacted.clone();
     }
+
+    /// Set the block height of the chain
+    pub fn set_block_height(&mut self, block_height: u32) {
```
vlopes11, line 82: Ditto; this is responsibility of the storage and shouldn't be exposed to the user via VM.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ repository = "https://github.com/FuelLabs/fuel-vm"
 description = "FuelVM interpreter."
 
 [dependencies]
+anyhow = "1.0"
```
vlopes11, line 14: This is added only for test utils and should be moved to `dev-dependencies`

vlopes11 (changes requested):
Seems the API is over-exposing the internals now. We should move `init` back to a private scope

src/interpreter/initialization.rs
Voxelot: Feel free to make some commits to this PR if you see a better way.

src/memory_client.rs
Voxelot: The problem is that this was panicking if transaction validation failed during init. I needed to capture the error instead of expecting here in order to capture the maturity error.

src/storage/memory.rs
Voxelot: This is on the in-memory storage only used in tests though, users aren't expected to use this. I wanted the ability to preserve storage between transaction - which MemoryStorage via commit and rollback, while modifying the block height during tests.

Cargo.toml
Voxelot: Well utils is not under only test and is publicly exported from the crate. But I should make this optional and only enable when the `test-helpers` feature is enabled.

src/memory_client.rs
vlopes11: Hmm that's very odd. If we panicked with infallible error, then the cause might be deeper. We can quick fix this and I'll investigate later.

src/interpreter/initialization.rs
vlopes11: Added an issue to fix it later; we can for now proceed with the fix

https://github.com/FuelLabs/fuel-vm/issues/156

The fix will potentially involve some tests refactor

src/storage/memory.rs
Voxelot: used `test-helpers` as that seems like the best flag available in the VM.

src/interpreter/initialization.rs
Voxelot: thanks for opening that issue

adlerjohn (commented):
Seems reasonable. Will defer to @vlopes11 for final approval.

segfault-magnet:
I've tested these changes against script transactions and they pass with flying colors but I'm getting 'PredicateFailure' if I set maturity != 0 on a `Create` transaction. 

Is this expected?

adlerjohn:
> then I'll get the predicate error

Do you have a predicate that checks the maturity, or just a predicate in general that returns `false`?

Voxelot:
So this fix only addresses maturity at the txn level. Maturity on inputs is validated in fuel core because it is state dependent on when the coin was originally created. Iirc the specs say that for inputs their maturity is also based on which block the coin was added in.

Cargo.toml
```diff
@@ -27,7 +28,8 @@ tracing = "0.1"
 rand = { version = "0.8", optional = true }
 
 [dev-dependencies]
-fuel-tx = { version = "0.13", features = ["random"] }
+fuel-crypto = { version = "0.5", features = ["random"] }
```
vlopes11, line 31: Was this added as nit? I don't see why we now need `random` from `fuel-crypto`, based in this PR. Besides this, lgtm

vlopes11 (approved):

segfault-magnet:
> > then I'll get the predicate error
> 
> Do you have a predicate that checks the maturity, or just a predicate in general that returns `false`?

This is the test I'm doing:
```rust
#[tokio::test]
async fn contract_deployment_respects_maturity() -> anyhow::Result<()> {
    abigen!(
        MyContract,
        "packages/fuels-abigen-macro/tests/test_projects/transaction_block_height/out/debug/transaction_block_height-abi.json"
    );

    let wallet = launch_provider_and_get_wallet().await;

    let parameters = TxParameters {
        maturity: 1,
        ..Default::default()
    };

    let deploy_the_contract = || {
        Contract::deploy(
            "tests/test_projects/transaction_block_height/out/debug/transaction_block_height.bin",
            &wallet,
            parameters,
        )
    };

    deploy_the_contract()
        .await
        .err()
        .expect("Should have failed because the block height was less than the requested maturity");

    add_blocks(&wallet, 1).await?;

    // it fails here with:
    // TransactionError("Response errors; unexpected block execution error
    // VmExecution { error: PredicateFailure, transaction_id:
    // 387e9cca6bcc1730aa8b0684c26e6cf96c2e86d5673b8db4cb3bf1dd8dde0070 }")' ,
    deploy_the_contract().await.expect(
        "Should have passed because the block height was greater or equal than the requested maturity",
    );

    Ok(())
}
```

Where `add_blocks` will submit a script that does nothing to increase the block height.

I don't know about maturity on inputs -- just seemed at a glance that this should behave just like a script behaves with regards to maturity.

Cargo.toml
Voxelot: hmm maybe that was leftover from before I refactored some tests to use the fuel-tx builder

Voxelot:
Here is where the specs denote the difference in maturity. This allows you to timelock some coins from being spent, whereas tx maturity is mainly just for queuing the tx execute at some time in the future.

> UTXO being spent must have been created at least this many blocks ago.

[InputCoin](https://github.com/FuelLabs/fuel-specs/blob/e0d5d8aea3f6c344ca47776403accc58b67eeb01/specs/protocol/tx_format.md#inputcoin)

PR #158, opened 2022-06-25
Update fuel-asm to 0.6 and fuel-tx to 0.14

vlopes11:

PR #159, opened 2022-06-28
Implement `SMO` instruction

vlopes11:


Cargo.toml
```diff
@@ -12,11 +12,13 @@ description = "FuelVM interpreter."
 
 [dependencies]
 dyn-clone = { version = "1.0", optional = true }
-fuel-asm = "0.5"
+#fuel-asm = "0.6"
+fuel-asm = { git = "https://github.com/FuelLabs/fuel-asm", branch = "vlopes11/gtf" }
```
adlerjohn, line 16: Unblocked?

src/interpreter/executors/instruction.rs
```diff
@@ -240,16 +242,6 @@ where
                 self.alu_set(ra, b ^ imm)?;
             }
 
-            OpcodeRepr::CIMV => {
```
Voxelot, line 243: ideally we remove these in a separate PR to make it easier to track what changes were made in each release.

vlopes11: We have this pr: https://github.com/FuelLabs/fuel-vm/pull/158

But its blocked by the fuel-tx one: https://github.com/FuelLabs/fuel-tx/pull/149

src/interpreter/blockchain.rs
```diff
@@ -308,4 +309,64 @@ where
...
+            if self.is_external_context() {
+                self.base_asset_balance_sub(amount)?;
+            } else {
+                self.external_asset_balance_sub(amount)?;
+            }
+        }
+
+        let fp = self.registers[REG_FP] as usize;
+        let txid = self.tx_id();
+        let data = a as usize + Address::LEN;
```
pixelcircuits, line 357: Maybe this should use `Bytes32::LEN` like on line 316 for consistency.

src/interpreter/executors/instruction.rs
```diff
@@ -36,11 +36,13 @@ where
         // Should execute `lo` only if there is no rupture in the flow - that means
         // either a breakpoint or some instruction that would skip `lo` such as
         // `RET`, `JI` or `CALL`
-        if self.registers[REG_PC] == pc && state.should_continue() {
+        let result = if self.registers[REG_PC] == pc && state.should_continue() {
```
pixelcircuits, line 39: Why was `result` added?

src/interpreter/blockchain.rs
```diff
@@ -308,4 +309,64 @@ where
...
+        }
+
+        // validations passed, perform the mutations
+
+        // subtract the balance only if there is an amount set
+        if amount > 0 {
+            if self.is_external_context() {
+                self.base_asset_balance_sub(amount)?;
+            } else {
+                self.external_asset_balance_sub(amount)?;
```
pixelcircuits, line 351: The SMO amount only ever refers to the base asset. In ether context, `self.base_asset_balance_sub` should be used

src/interpreter/internal.rs
```diff
@@ -155,6 +155,20 @@ impl<S> Interpreter<S> {
         Ok(())
     }
 
+    /// Reduces the unspent balance of the asset defined in $fp
+    pub(crate) fn external_asset_balance_sub(&mut self, value: Word) -> Result<(), RuntimeError> {
+        let fp = self.registers[REG_FP] as usize;
+        let asset_id = &self.memory[fp..fp + AssetId::LEN];
```
pixelcircuits, line 161: Isn't the `asset_id` the second set of 32 bytes on the [call frame](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/main.md#call-frames)?

adlerjohn: There should be a test for this is what this means!

src/interpreter/blockchain.rs
vlopes11: True, they should be the same. The concrete type is in fact `Address`, instead of raw bytes. Better to use that so its clear what is expected of this register:

https://github.com/FuelLabs/fuel-vm/pull/159/commits/7bd578949cb2d81e1bd84388532d33e523473636

src/interpreter/executors/instruction.rs
vlopes11: This is to facilitate debug and readability. If we just return complex expressions without assigning to a variable, then we need to inspect the frame pointers in gdb - having an assigned variable is zero runtime overhead and will give us that readability for free

src/interpreter/blockchain.rs
vlopes11: https://github.com/FuelLabs/fuel-vm/pull/159/commits/4398216d539b52c63b2098ea4ddf38245b480924

Do you think we should rephrase here to simplify the description since we only update the base asset?

https://github.com/FuelLabs/fuel-specs/blob/bd6ec935e3d1797a192f731dadced3f121744d54/specs/vm/instruction_set.md?plain=1#L1417

src/interpreter/internal.rs
vlopes11: This function was added as a specific use-case for SMO and is no longer present

src/interpreter/blockchain.rs
adlerjohn: That line does say `balanceOfStart(0)` and `asset ID 0`, so it's explicit and unambiguous if you know what you're looking for. More generally, the specs aren't approachable _unless_ you've read them almost entirely, and that probably should be improved.

src/interpreter/initialization.rs
```diff
@@ -88,20 +88,25 @@ impl<S> Interpreter<S> {
 
     // compute the initial free balances for each asset type
     pub(crate) fn initial_free_balances(&self) -> Result<HashMap<AssetId, Word>, InterpreterError> {
+        let base_asset = AssetId::default();
```
adlerjohn, line 91: Orthogonal to this PR, but I'm thinking the base asset's ID should be a parameter in the VM. It's not necessarily guaranteed to be `AssetId::default()` always. Both the specs and the implementation should be modular in this regard. Can you file an issue to track changing?

src/interpreter/initialization.rs
```diff
@@ -124,19 +129,20 @@ impl<S> Interpreter<S> {
...
                     })?;
+
+            *base_asset_balance = deducted_balance;
         }
 
-        // reduce free balances by coin and withdrawal outputs
+        // reduce free balances by coin
         for (asset_id, amount) in self.tx.outputs().iter().filter_map(|output| match output {
             Output::Coin { asset_id, amount, .. } => Some((asset_id, amount)),
-            Output::Message { .. } => todo!(),
```
adlerjohn, line 139: Wait, message outputs still count as "withdrawals," or rather, they should still reduce the free balance. What is this change doing? Is the output amount accounted for?

src/interpreter/transaction.rs
```diff
@@ -52,7 +52,7 @@ impl<S> Interpreter<S> {
             .tx
             .witnesses()
             .get(b as usize)
-            .ok_or(PanicReason::OutputNotFound)
+            .ok_or(PanicReason::WitnessNotFound)
```
adlerjohn, line 55: This looks like an orthogonal bugfix. Ideally make a separate PR for this.

adlerjohn (commented):
minor comments

src/interpreter/executors/instruction.rs
pixelcircuits: cool!

src/interpreter/initialization.rs
pixelcircuits: OutputMessage amount is 0 or the whole transaction is invalid.

adlerjohn: Ah right they're variable.

adlerjohn (approved):
Looks fine to me, would like @pixelcircuits and @Voxelot to give a review as well.

Dentosal (approved):

pixelcircuits (approved):

src/interpreter/transaction.rs
vlopes11: Thats true, I just spotted this nit while debugging

PR #160, opened 2022-06-29
Disable faulty predicate checks during create tx execution

Voxelot:
disable duplicative predicate verification checks when submitting create txs. Predicate verification should be executed by the client / txpool, not when a tx is run by the vm.

partially adresses: https://github.com/FuelLabs/fuel-core/issues/449


vlopes11 (approved):
lgtm

PR #165, opened 2022-07-04
Enable clippy

### LINKED ISSUE #163: Deny clippy::dbg_macro lint in CI
Lint description: https://rust-lang.github.io/rust-clippy/master/#dbg_macro
 
Deny only in CI. This automatically makes sure that `dbg!` macro calls are detected before PR review, but can still be used when developing. I'm still uncertain should it be allowed in tests.

Dentosal:
Fixes all clippy lint issues and closes #163. This makes `cargo clippy` usable form the command line, and also does some additional checks on the CI.

Currently a bit ugly, as we have to disable `clippy::iter_cloned_collect` in some files due to a clippy bug: https://github.com/rust-lang/rust-clippy/issues/9119 If this is an issue, we can wait until the bug is resolved.


adlerjohn (dismissed):
lgtm, but deferring to @vlopes11 and @Voxelot for decision

adlerjohn (commented):
merge conflicts

adlerjohn (approved):
same comment as before

src/interpreter/executors/main.rs
```diff
@@ -30,6 +30,7 @@ impl Interpreter<PredicateStorage> {
             return false;
         }
 
+        #[allow(clippy::needless_collect)] // TODO: the collect could probably be removed
```
vlopes11, line 33: again; this exists only for code readability and will cause zero runtime overhead. With the current code, its unambiguous that we are extracting memory ranges that are representations of predicate programs.

src/interpreter/memory.rs
```diff
@@ -9,6 +9,7 @@ use std::{ops, ptr};
 
 #[derive(Debug, Clone, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[allow(clippy::derive_hash_xor_eq)] // Manual impl doesn't violate Hash guarantees
```
vlopes11, line 12: of all clippy lints, this is the one I dislike the most 😂

what if we allow this as global/per default?

tests/blockchain.rs
```diff
@@ -415,11 +417,11 @@ fn can_read_state_from_initial_storage_slots() {
         Opcode::LW(0x10, 0x10, 0),
         // alloc 32 bytes stack space
         Opcode::ADDI(0x11, REG_SP, 0),
-        Opcode::CFEI(32 as Immediate24),
+        Opcode::CFEI(32),
```
vlopes11, line 420: Mixed feelings in this one as well. I wanted to make it clear there that we are using immediate values - and not register references.

tests/blockchain.rs
```diff
@@ -518,12 +520,14 @@ fn smo_instruction_works() {
         let txid = tx.id();
         let receipts = client.transact(tx);
 
-        let success = receipts.iter().any(|r| match r {
-            Receipt::ScriptResult {
-                result: ScriptExecutionResult::Success,
-                ..
-            } => true,
-            _ => false,
+        let success = receipts.iter().any(|r| {
```
vlopes11, line 523: with this style, we get an extra level of indentation. Is that an improvement?

tests/encoding.rs
```diff
@@ -27,8 +27,8 @@ where
         buffer = vec![0u8; read_size];
 
         // Minimum size buffer assertion
-        d.read(buffer.as_mut_slice()).expect("Failed to read");
-        d_p.write(buffer.as_slice()).expect("Failed to write");
+        let _ = d.read(buffer.as_mut_slice()).expect("Failed to read");
```
vlopes11, line 30: maybe allow this one per default? this one deteriorates the readability imo, and doesn't really make any sense for buffer contexts such as this one.

vlopes11 (approved):
LGTM

Apart from the changes that are unmistakable readability improvements, a couple of others causes me mixed feelings on whether or not we are actually improving the code.

imo, the stablished clippy culture in the rust community came with a fundamental flaw: to add long code remarks to satisfy lints that shouldn't be general rules.

We could consider a global `clippy.toml` for the entirety of Fuel stack, but that is a quite naive approach from the tool: we might have trouble differentiating between production code and test code; the latter must be extra verbose while the former is sometimes better if succinct - this means the same ruleset wouldn't be applicable for both.

src/interpreter/memory.rs
Dentosal: I personally don't hate it, as it's a common error and I don't mind having to disable the lint by hand. Still, I'm not against disabling this globally.

tests/blockchain.rs
Dentosal: It's more readable to me, but a personal preference of course. I always code with clippy enabled, so my preferences often match with clippy.

src/interpreter/executors/instruction.rs
```diff
@@ -18,6 +18,7 @@ where
     S: InterpreterStorage,
 {
     /// Execute the current instruction pair located in `$m[$pc]`.
+    #[allow(clippy::let_and_return)]
```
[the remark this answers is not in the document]
Dentosal: I'd like to have an org-wide `.clippy.toml`, any ideas on how to implement this without manual copying? I think we also want separate files for CI vs local development.

src/interpreter/executors/main.rs
Dentosal: > will cause zero runtime overhead

I'm not sure if this is true, although I'd think the compiler is cabable of omitting the extra allocation. Anyway, I still agree that the current code is more readable.

Voxelot (approved):

Dentosal:
Closing as outdated. I'll open a new PR if this still needs to be done.

PR #168, opened 2022-07-13
Fix CGAS overflow bug

Dentosal:
I'm not sure if this is actually the correct approach, so please review carefully.

Also fortifies some gas arithmetics using `checked_*` methods, as per our discussion with @Voxelot.

### DIFF
--- a/src/interpreter/flow.rs
+++ b/src/interpreter/flow.rs
@@ -46,7 +46,9 @@ impl<S> Interpreter<S> {
 
     pub(crate) fn return_from_context(&mut self, receipt: Receipt) -> Result<(), RuntimeError> {
         if let Some(frame) = self.frames.pop() {
-            self.registers[REG_CGAS] += frame.context_gas();
+            self.registers[REG_CGAS] = self.registers[REG_CGAS]
+                .checked_add(frame.context_gas())
+                .ok_or(RuntimeError::halt_on_bug("CGAS would overflow"))?;
 
             frame
                 .registers()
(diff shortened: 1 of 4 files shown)


src/interpreter/flow.rs
```diff
@@ -158,6 +160,13 @@ where
         // credit contract asset_id balance
         self.balance_increase(call.to(), &asset_id, b)?;
 
+        let forward_gas_amount = cmp::min(self.registers[REG_GGAS], d);
+
+        // subtract gas
```
Voxelot, line 165: this is the fix I had in mind 👍🏻

adlerjohn (commented):
CI failing

Dentosal:
This seems to introduce a new bug, caught by the `profile-gas` tests only. Not sure if the bug is in tests or code, looking at it now.

src/interpreter/gas.rs
```diff
@@ -75,13 +75,19 @@ impl<S> Interpreter<S> {
         }
 
         if gas > self.registers[REG_CGAS] {
-            self.registers[REG_GGAS] -= self.registers[REG_CGAS];
+            self.registers[REG_GGAS] = self.registers[REG_CGAS]
+                .checked_sub(gas)
+                .ok_or(RuntimeError::halt_on_bug("gas invariant violation"))?;
```
Voxelot, line 80:
```suggestion
            self.registers[REG_GGAS] = self.registers[REG_GGAS]
                .checked_sub(self.registers[REG_CGAS])
                .ok_or(RuntimeError::halt_on_bug("gas invariant violation"))?;
```

Voxelot (approved):
lgtm

PR #171, opened 2022-07-14
Allow cgas to fall below requested forward amount

Voxelot:
When decrementing the cgas before making a contract call, we took the `min(forward_amount, current_global_gas)`. This meant that once cgas was less than the forward amount, an error would occur. 

This PR fixes the behavior to make cgas decrement using `min(forward_amount, current_context_gas)`. This ensures cgas will still allow calls to occur when less than the max forward amount.


Voxelot:
yep, that's why I left it in draft for now

Dentosal (approved):

PR #173, opened 2022-07-18
Add checked tx

vlopes11:


src/interpreter/balances.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        value
+    }
+
+    /// Attempt to add the balance of an asset, updating the VM memory in the appropriate
+    /// offset
+    pub fn checked_balance_add(&mut self, memory: &mut [u8], asset: &AssetId, value: Word) -> Option<Word> {
+        self.state
+            .get_mut(asset)
+            .and_then(|b| b.checked_sub(value))
```
Voxelot, line 106: How did this not get caught in tests?

```suggestion
            .and_then(|b| b.checked_add(value))
```

src/interpreter/balances.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub fn checked_balance_sub(&mut self, memory: &mut [u8], asset: &AssetId, value: Word) -> Option<Word> {
+        self.state
+            .get_mut(asset)
+            .and_then(|b| b.checked_sub(value))
+            .map(|balance| Self::_set_memory_balance(balance, memory))
+            .or((value == 0).then(|| 0))
+    }
+
+    /// Write all assets into the VM memory.
+    pub fn to_vm<S>(self, vm: &mut Interpreter<S>) {
```
Voxelot, line 122: This seems like a very impure and leaky encapsulation of logic, as the correctness is heavily reliant on the external / global state of the VM and the order of operations when initializing the stack outside of this call as well as manipulating REG_SP directly.

Instead, I think it would be cleaner to pass a mutable slice to write the balances state into, rather than a mutable reference to the whole VM. That way, the logic for setting reserving the stack and managing REG_SP is consolidated into one place rather than delegated away.

src/interpreter/balances.rs
```diff
@@ -0,0 +1,191 @@
...
+impl Index<&AssetId> for RuntimeBalances {
+    type Output = Word;
+
+    fn index(&self, index: &AssetId) -> &Self::Output {
+        &self.state[index].value
+    }
+}
+
+#[test]
+fn runtime_balances_writes_to_memory_correctly() {
```
Voxelot, line 150: we should also have unit tests that verify checked_add and checked_sub update the in-memory values properly.

src/interpreter/executors/main.rs
```diff
@@ -85,34 +90,36 @@ where
 {
     // TODO maybe infallible?
```
Voxelot, line 91: We could fail due to I/O issues or problems that trigger a `halt_on_bug`, so I don't think this TODO is needed.

```suggestion
```

src/interpreter/memory.rs
```diff
@@ -420,10 +420,12 @@ mod tests {
     use crate::prelude::*;
 
     #[test]
+    #[ignore]
```
Voxelot, line 423: why are we ignoring these tests now?

Voxelot:
Please add PR description giving a high-level breakdown of the changes be done here

tests/cgas_overflow.rs
```diff
@@ -0,0 +1,96 @@
+use fuel_vm::consts::*;
```
Voxelot, line 1: This file was moved to `tests/cgas.rs` and should be removed.

Voxelot (changes requested):
Requesting the following fixes:

1. found a code dupe bug in balances
2. merge issue with duplicate cgas tests
3. ignored tests without any explanation

src/interpreter/balances.rs
vlopes11: The reason this nit wasn't caught in tests is that a balance add never happen during VM runtime. We initialize the balances from the available free assets, and only decrease them.

But you're right, we should have this covered. I fixed it here and added unit tests to this new struct:

https://github.com/FuelLabs/fuel-vm/pull/173/commits/902e894408a347987f157371eea947322ecb6352

src/interpreter/balances.rs
vlopes11: We don't have a use-case where we update a slice of bytes with the balances - except when this slice of bytes happens to be the memory representation of the VM. This means we don't really update slices, but instead, update the VM itself. Keep the function like this unmistakably translates into that - while receiving slices allows multiple incorrect interpretations of what this function is supposed to do

Incrementing the stack pointer here is also an exception since we don't really do that anywhere else but during assets initialization. Also, its worthy to keep in mind that this balances struct definition is part of the interpreter module - therefore, it is ok to access the VM internals.

src/interpreter/balances.rs
vlopes11: Agreed, done here

https://github.com/FuelLabs/fuel-vm/pull/173/commits/902e894408a347987f157371eea947322ecb6352

src/interpreter/executors/main.rs
vlopes11: This was created a long time ago, when I was centralizing the I/O exceptions into `std::io::Error`. Its not really related to this PR, but won't hurt to remove it here

https://github.com/FuelLabs/fuel-vm/pull/173/commits/26b3663740c1064a3baa47434d2355d9f90c6289

tests/cgas_overflow.rs
vlopes11: https://github.com/FuelLabs/fuel-vm/pull/173/commits/62af3eeb75ef71b79f1f61e21341690021a16d6d

vlopes11:
Adding some relevant description to the PR and commit; re-requesting review to be done in parallel

src/interpreter/balances.rs
Voxelot: > this balances struct definition is part of the interpreter module - therefore, it is ok to access the interpreter internals

In that case, it seems like it may be a good idea to make this struct `pub(crate)` instead of `pub` so that it doesn't get misused outside the intended context. But It's not worth blocking the PR on this point.

src/interpreter/balances.rs
```diff
@@ -0,0 +1,323 @@
...
+    assert_eq!(val, 150);
+    assert_eq!(bal, 150);
+
+    let val = balances
+        .checked_balance_add(&mut memory, &asset, 75)
+        .expect("failed to add balance");
+    let bal = balances.balance(&asset).expect("failed to fetch balance");
+
+    assert_eq!(val, 225);
+    assert_eq!(bal, 225);
```
Voxelot, line 283: We should also test that the call to `_set_memory_balance` is working as expected. Currently, we only verify the VM memory access for initialization, although the VM memory is also modified whenever these balances change.

Voxelot (approved):
I think this is pretty good for now, although it would be good to see additional testing for the VM memory modifications triggered by `RuntimeBalances::checked_balance_sub`.

PR #174, opened 2022-07-21
Implement GM::GetVerifyingPredicate metadata instr

vlopes11:
The context of the execution should contain more info such as the index
of the current predicate so we decouple that from the vm initialization.

Also, its worthy to create a type that will encapsulate the behavior of
the runtime predicates since it needs to store not only the program to
be executed - so it can't be represented just as a memory slice - but
the original input index that created the predicate.

The input index is not required for the predicate execution since the
program memory offset is a constant function of the consensus
parameters, but it might be consumed by the user - as in the specs.

### DIFF
--- a/src/interpreter/initialization.rs
+++ b/src/interpreter/initialization.rs
@@ -13,10 +13,8 @@ use std::io;
 
 impl<S> Interpreter<S> {
     /// Initialize the VM with a given transaction
-    pub fn init(&mut self, block_height: u32, tx: CheckedTransaction) -> Result<(), InterpreterError> {
+    pub fn init(&mut self, tx: CheckedTransaction) -> Result<(), InterpreterError> {
         self.tx = tx;
-
-        self.block_height = block_height;
 
         self.frames.clear();
         self.receipts.clear();
@@ -55,13 +53,11 @@ impl<S> Interpreter<S> {
 
     /// Initialize the VM for a predicate context
     pub fn init_predicate(&mut self, tx: CheckedTransaction) -> bool {
-        let block_height = 0;
-
         self.context = Context::Predicate {
             program: Default::default(),
         };
 
-        self.init(block_height, tx).is_ok()
+        self.init(tx).is_ok()
     }
 }
 
@@ -74,10 +70,8 @@ where
     ///
     /// For predicate verification, check [`Self::init`]
     pub fn init_with_storage(&mut self, tx: CheckedTransaction) -> Result<(), InterpreterError> {
-        let block_height = self.storage.block_height().map_err(InterpreterError::from_io)?;
-
         self.context = Context::Script;
 
-        self.init(block_height, tx)
+        self.init(tx)
     }
 }
--- a/src/interpreter/initialization.rs
+++ b/src/interpreter/initialization.rs
@@ -70,7 +70,9 @@ where
     ///
     /// For predicate verification, check [`Self::init`]
     pub fn init_with_storage(&mut self, tx: CheckedTransaction) -> Result<(), InterpreterError> {
-        self.context = Context::Script;
+        let block_height = self.storage.block_height().map_err(InterpreterError::from_io)?;
+
+        self.context = Context::Script { block_height };
 
         self.init(tx)
     }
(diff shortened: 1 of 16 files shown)


src/interpreter/initialization.rs
```diff
@@ -13,11 +13,10 @@ use std::io;
 
 impl<S> Interpreter<S> {
     /// Initialize the VM with a given transaction
-    pub fn init(&mut self, predicate: bool, block_height: u32, tx: CheckedTransaction) -> Result<(), InterpreterError> {
+    pub fn init(&mut self, block_height: u32, tx: CheckedTransaction) -> Result<(), InterpreterError> {
```
vlopes11, line 16: Block height seems to be something that belongs to the context, and not to VM initialization. Should we tackle that in this PR?

src/interpreter/initialization.rs
```diff
@@ -64,9 +74,10 @@ where
     ///
     /// For predicate verification, check [`Self::init`]
     pub fn init_with_storage(&mut self, tx: CheckedTransaction) -> Result<(), InterpreterError> {
-        let predicate = false;
         let block_height = self.storage.block_height().map_err(InterpreterError::from_io)?;
 
-        self.init(predicate, block_height, tx)
+        self.context = Context::Script;
```
Voxelot, line 79: [nit] Should we also have a `Create` context?

src/predicate.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Create a new runtime predicate from a transaction, given the input index
+    ///
+    /// Return `None` if the tx input doesn't map to an input coin with a predicate
+    pub fn from_tx<T>(params: &ConsensusParameters, tx: T, idx: usize) -> Option<Self>
+    where
+        T: Borrow<Transaction>,
+    {
+        tx.borrow()
+            .input_coin_predicate_offset(idx)
```
Voxelot, line 37: Hmm it seems like we are missing support for `MessagePredicate` verification as well. Will this require changes to fuel-tx?

Voxelot: created an issue here: https://github.com/FuelLabs/fuel-tx/issues/166

Voxelot: also filed an issue here in fuel-vm, so we can add message predicate verification as a followup: https://github.com/FuelLabs/fuel-vm/issues/175

src/interpreter/initialization.rs
Voxelot: It seems like we shouldn't be setting block_height on the VM at all, especially now that we moved the transaction validation logic out of VM initialization I don't think it would be used. The block height opcode should be able to just fetch it from interpreter storage as needed.

vlopes11: https://github.com/FuelLabs/fuel-vm/pull/174/commits/3716e92f8a45823387c8fc380bc7fdcdbd3eb287

src/interpreter/initialization.rs
vlopes11: Most definitely. Making some tests now to see if its a small impact change

src/interpreter/initialization.rs
vlopes11: The previous implementation of the context isn't good. Also, the block height belongs to the context and should be fetched from the storage a single time (on script init since block height isn't valid for predicates).

We shouldn't fetch the block height from the storage during runtime because this is a fallible implementation from the storage:

https://github.com/FuelLabs/fuel-vm/pull/174/commits/2f781e30e190c82ce1c0ca194992435d22f6e363

src/interpreter/initialization.rs
```diff
@@ -53,6 +50,15 @@ impl<S> Interpreter<S> {
 
         Ok(())
     }
+
+    /// Initialize the VM for a predicate context
+    pub fn init_predicate(&mut self, tx: CheckedTransaction) -> bool {
+        self.context = Context::Predicate {
+            program: Default::default(),
```
Voxelot, line 57: why use a default value here instead of just passing through the idx of the predicate?

Voxelot (approved):

PR #177, opened 2022-07-27
Implement `GTF` instruction

vlopes11:
Implements the `GTF` instruction.

The tests are being developed in parallel. This PR is released so the accumulated breaking changes can be unblocked for `fuel-core`.

The tests are expected to be added only incrementally, so they can be released as patch (with eventual bugfixes, if applicable)

### DIFF
--- a/tests/metadata.rs
+++ b/tests/metadata.rs
@@ -1,8 +1,10 @@
 use fuel_crypto::Hasher;
+use fuel_tx::TransactionBuilder;
 use fuel_vm::consts::*;
-use fuel_vm::prelude::*;
 use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
+
+use fuel_vm::prelude::*;
 
 #[test]
 fn metadata() {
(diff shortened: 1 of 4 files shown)


vlopes11:
Blocked by https://github.com/FuelLabs/fuel-tx/pull/169

tests/metadata.rs
```diff
@@ -154,3 +156,84 @@ fn metadata() {
...
+
+    #[rustfmt::skip]
+    let cases = vec![
+        inputs_bytes[0].clone(),
+    ];
+
+    #[rustfmt::skip]
+    let script = vec![
+        Opcode::MOVI(0x20, 0x01),
+        Opcode::gtf(0x30, 0x00, GTFArgs::ScriptData),
```
Dentosal, line 196: Why is gtf here lowercase?

Dentosal (changes requested):

tests/metadata.rs
vlopes11: We should probably split and build a dedicated dsl for the tx scripts, but its currently centralized in `Opcode` that that one is a function: [Opcode::gtf](https://docs.rs/fuel-asm/latest/fuel_asm/enum.Opcode.html#method.gtf)

Dentosal: Oh so that's just for converting from GTFArgs to Imm12

Dentosal (approved):

PR #178, opened 2022-07-27
Add tests for `GTF` instruction

vlopes11:


bvrooman (approved):
LGTM but will defer to others with VM and opcode knowledge

Voxelot (approved):
it'd be nice if we parameterized the test cases at the harness level so that failures would be associated with specific cases. Otherwise, if there's a failure anywhere inside this script, you essentially have to start bisecting the assembly to find the issue.

vlopes11:
> It'd be nice if we parameterized the test cases at the harness level so that failures would be associated with specific cases. Otherwise, if there's a failure anywhere inside this script, you essentially have to start bisecting the assembly to find the issue.

True, we have some `TODOs` there to cover the predicate and create cases. We should go into that direction as well. We also need to include some sort of negative tests since we are testing some naive basic cases, and not protecting against panic, potential ub, etc

Issue #167, opened 2022-07-11
Remove the `XIS`, `XOS`, `XWS`, `XIL`, `XOL` & `XWL` instructions from the vm.


nfurfaro:
These have now been replaced by the new `gtf` instruction.

### Important:

We should wait until gtf is fully functional before removing these!

Dentosal:
https://github.com/FuelLabs/fuel-vm/pull/179

PR #180, opened 2022-08-02
Update fuel-tx to v0.17

vlopes11:
This upgrade includes the tx pointer as part of the input structure, the
ability to perform predicate verification for input messages, and an API
to prepare transactions for either predicate verification or script
execution.

It also removes the `X**` instructions since they are supersede by `GTF`


Dentosal (approved):

Dentosal:
Closes #167

Voxelot (approved):
lgtm, although we should probably have a test to verify MessagePredicate inputs are actually verified

PR #184, opened 2022-08-03
Fix gas subtraction overflow

vlopes11:
As pointed by @mediremi , we might overflow on the post-execution gas
deduction.

That operation must be safe, so we have to use `checked_sub`

### DIFF
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -135,7 +135,11 @@ where
                 // TODO set tree balance
 
                 let program = self.run_program();
-                let gas_used = self.transaction().gas_limit() - self.registers[REG_GGAS];
+                let gas_used = self
+                    .transaction()
+                    .gas_limit()
+                    .checked_sub(self.registers[REG_GGAS])
+                    .ok_or(InterpreterError::Panic(PanicReason::ArithmeticOverflow))?;
 
                 // Catch VM panic and don't propagate, generating a receipt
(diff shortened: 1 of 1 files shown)


src/interpreter/executors/main.rs
```diff
@@ -135,7 +135,11 @@ where
                 // TODO set tree balance
 
                 let program = self.run_program();
-                let gas_used = self.transaction().gas_limit() - self.registers[REG_GGAS];
+                let gas_used = self
+                    .transaction()
+                    .gas_limit()
+                    .checked_sub(self.registers[REG_GGAS])
+                    .ok_or(InterpreterError::Panic(PanicReason::ArithmeticOverflow))?;
```
Dentosal, line 142: Why is the panic reason `ArithmeticOverflow`? It doesn't feel right as this is a gas-related issue

Dentosal (approved):

Dentosal (approved):

src/interpreter/executors/main.rs
vlopes11: True. This error should not happen in normal cases for sound transactions, so its debatable on whether we should use canonical errors for this exception, or just classify it by what it is: an underflow.

But then it might be better to stick to the easier to track option that is out of gas: https://github.com/FuelLabs/fuel-vm/pull/184/commits/9cf81732a0ac3c2b96932858bf86d69a79223f37

PR #185, opened 2022-08-04
Assert non-writable property of reserved registers

vlopes11:
All instructions that writes data to a register must be shielded by the
assertion that the register is writable

### DIFF
--- a/src/interpreter/blockchain.rs
+++ b/src/interpreter/blockchain.rs
@@ -233,6 +233,8 @@ where
     }
 
     pub(crate) fn state_read_word(&mut self, ra: RegisterId, b: Word) -> Result<(), RuntimeError> {
+        Self::is_register_writable(ra)?;
+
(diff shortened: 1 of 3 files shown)


Dentosal (approved):

Voxelot:
When fixing bugs like these we should include test cases to verify repro and prevent the issue from happening again

Dentosal:
Maybe we could have one big test that checks that all instructions honor the read-only properties of reserved registers?

vlopes11:
We need some serious test coverage for a lot of other stuff we aren't checking now. Covering these micro edges is a project by itself and can't be solved here - unless we just add tests for these instructions, which will neither grant the assertion of this rule nor cover the other edges

Voxelot:
I think it's a good practice to capture each bug discovered into a test case. Otherwise we're just eyeballing and manually testing fixes. Since bugs expose edge cases, that's the perfect time to add coverage for the edge case that was uncovered.

PR #186, opened 2022-08-04
Add tests for some write-only registers checks

Dentosal:
Adds test cases for recently merged #185

PR #187, opened 2022-08-04
Quickcheck reserved register access

Voxelot:
add some light quickcheck fuzzing to ensure reserved registers can't be written to. It uses a match over opcode variants to explicitly specify which opcodes are expected to write to rA. This will help with future proofing against future opcodes, as this match (`writes_to_ra`) will have to be updated.

may replace: https://github.com/FuelLabs/fuel-vm/pull/186

### DIFF
--- a/src/interpreter/executors/instruction/tests/reserved_registers.rs
+++ b/src/interpreter/executors/instruction/tests/reserved_registers.rs
@@ -1,0 +1,140 @@
+use super::*;
+use quickcheck::TestResult;
+use quickcheck_macros::quickcheck;
+
+// Ensure none of the opcodes can write to reserved registers
+#[quickcheck]
+fn cant_write_to_reserved_registers(opcode: u8, fuzz: u16) -> TestResult {
+    // opcode || REG_ONE || random fuzz
+    let attempt_reserved_reg_access = ((opcode as u32) << 24) | (1 << 18) | ((fuzz >> 4) as u32);
+    let instruction = Instruction::new(attempt_reserved_reg_access);
+    let opcode = Opcode::new(instruction);
+    // skip undefined opcodes
+    if matches!(opcode, Opcode::Undefined) {
+        return TestResult::discard();
+    }
+
+    let mut vm = Interpreter::with_memory_storage();
+
+    vm.init_script(CheckedTransaction::default())
+        .expect("Failed to init VM");
+    let res = vm.instruction(instruction);
+
+    if writes_to_ra(opcode) {
+        // if this opcode writes to $rA, expect an error since we're attempting to use a reserved register
+        if !matches!(
+            res,
+            Err(InterpreterError::PanicInstruction(r)) if r.reason() == &ReservedRegisterNotWritable
+        ) {
+            return TestResult::error(format!(
+                "expected ReservedRegisterNotWritable error {:?}",
+                (opcode, &res)
+            ));
+        }
+    } else {
+        // Ensure REG_ONE wasn't changed.
+        // While not a perfect guarantee against the opcode writing a value
+        // to an invalid register, this increases the likelihood of detecting
+        // erroneous register access.
+        if vm.registers[REG_ONE] != 1 {
+            return TestResult::error("reserved register was modified!");
+        }
+
+        // throw err if a ReservedRegisterNotWritable err was detected outside our writes_to_ra check
+        // This would likely happen if the opcode wasn't properly marked as true in `writes_to_ra`
+        if matches!(
+            res,
+            Err(InterpreterError::PanicInstruction(r)) if r.reason() == &ReservedRegisterNotWritable
+        ) {
+            return TestResult::error(format!(
+                "unexpected ReservedRegisterNotWritable, test configuration may be faulty {:?}",
+                (opcode, &res)
+            ));
+        }
+    }
+    TestResult::passed()
+}
+
+// determines whether a given opcode stores a value into $rA
+fn writes_to_ra(opcode: Opcode) -> bool {
+    match opcode {
+        Opcode::ADD(_, _, _) => true,
+        Opcode::ADDI(_, _, _) => true,
+        Opcode::AND(_, _, _) => true,
+        Opcode::ANDI(_, _, _) => true,
+        Opcode::DIV(_, _, _) => true,
+        Opcode::DIVI(_, _, _) => true,
+        Opcode::EQ(_, _, _) => true,
+        Opcode::EXP(_, _, _) => true,
+        Opcode::EXPI(_, _, _) => true,
+        Opcode::GT(_, _, _) => true,
+        Opcode::LT(_, _, _) => true,
+        Opcode::MLOG(_, _, _) => true,
+        Opcode::MROO(_, _, _) => true,
+        Opcode::MOD(_, _, _) => true,
+        Opcode::MODI(_, _, _) => true,
+        Opcode::MOVE(_, _) => true,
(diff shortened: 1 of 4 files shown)


src/interpreter/executors/instruction/tests/reserved_registers.rs
```diff
@@ -0,0 +1,140 @@
...
+                "unexpected ReservedRegisterNotWritable, test configuration may be faulty {:?}",
+                (opcode, &res)
+            ));
+        }
+    }
+    TestResult::passed()
+}
+
+// determines whether a given opcode stores a value into $rA
+fn writes_to_ra(opcode: Opcode) -> bool {
```
Dentosal, line 59: It might be a good idea to instead expose this property from `fuel-asm` itself as `Opcode::writes_to_ra()`. Although I cannot imagine it being used for anything else than this test, I think it might help us to track breaking changes in there. Or is the intent that `fuel-asm` only concerns itself with representation, leaving the semantics to `fuel-vm`?

src/interpreter/executors/instruction/tests/reserved_registers.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    let mut vm = Interpreter::with_memory_storage();
+
+    vm.init_script(CheckedTransaction::default())
+        .expect("Failed to init VM");
+    let res = vm.instruction(instruction);
+
+    if writes_to_ra(opcode) {
+        // if this opcode writes to $rA, expect an error since we're attempting to use a reserved register
+        if !matches!(
```
Dentosal, line 25: nit: This requires that write-permission for $rA is checked before verifying any other properties of the instruction that might cause an error. This is how it's currently done, but maybe adding a comment explaining that wouldn't hurt.

src/interpreter/executors/instruction/tests/reserved_registers.rs
```diff
@@ -0,0 +1,140 @@
...
+            ));
+        }
+    } else {
+        // Ensure REG_ONE wasn't changed.
+        // While not a perfect guarantee against the opcode writing a value
+        // to an invalid register, this increases the likelihood of detecting
+        // erroneous register access.
+        if vm.registers[REG_ONE] != 1 {
+            return TestResult::error("reserved register was modified!");
+        }
```
Dentosal, line 41: We should probably do this check irregardless of `writes_to_ra(opcode)`, as modifying REG_ONE to have any other value would be incorrect in either case.

Dentosal (changes requested):

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: ah good point, i originally had these two checks flipped

src/interpreter/executors/instruction/tests/reserved_registers.rs
```diff
@@ -0,0 +1,140 @@
+use super::*;
+use quickcheck::TestResult;
+use quickcheck_macros::quickcheck;
+
+// Ensure none of the opcodes can write to reserved registers
+#[quickcheck]
+fn cant_write_to_reserved_registers(opcode: u8, fuzz: u16) -> TestResult {
+    // opcode || REG_ONE || random fuzz
+    let attempt_reserved_reg_access = ((opcode as u32) << 24) | (1 << 18) | ((fuzz >> 4) as u32);
```
vlopes11, line 9: It will be maybe too much coupled if we operate the internals of an instruction here. Maybe we benefit from [Instruction::from(opcode)](https://docs.rs/fuel-asm/0.7.0/fuel_asm/struct.Instruction.html#impl-From%3COpcode%3E-for-Instruction)?

src/interpreter/executors/instruction/tests/reserved_registers.rs
```diff
@@ -0,0 +1,140 @@
...
+                "expected ReservedRegisterNotWritable error {:?}",
+                (opcode, &res)
+            ));
+        }
+    } else {
+        // Ensure REG_ONE wasn't changed.
+        // While not a perfect guarantee against the opcode writing a value
+        // to an invalid register, this increases the likelihood of detecting
+        // erroneous register access.
+        if vm.registers[REG_ONE] != 1 {
```
vlopes11, line 39: Shouldn't we try to write to other registers as well? Anything below 

https://github.com/FuelLabs/fuel-vm/blob/a5a6759fbb7804bb4375ca299d9f68322a47db4c/src/consts.rs#L61-L62

Dentosal: I think it's enough to test any single register, since we can assume that the write-permission check is done using the same function everywhere

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: The difficulty is that there's not an easy way to programmatically generate randomized Opcode's without handling each enum variant separately. The Instruction type would be a little easier to arbitrarily muck with for fuzzing, but we'd likely need  a `test-helpers` feature to allow its internals to be set.

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: REG_ZERO or REG_ONE are simple to test because they are expected to be constants that should never change. Detecting illegal writes of other reserved registers seemed much more difficult because we'd have to distinguish between valid and invalid changes for each register, based on the randomized instruction that was executed.

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: The advantage of moving this to Opcode is that the VM could generalize the checks for reserved register access across any instruction. The disadvantage is that we end up with extra branching hits for opcodes that don't need the reserved register access check.

I generally prefer to keep utilities like this close to where they are used, until we see a clear shared need across the board. Since fuel-vm is entirely responsible for deciding which opcodes need writable register checks, I think it makes sense to keep this util in the VM for now. There are likely a lot of other shared validation rules that apply to specific subcategories of opcodes, but since they are enforced by fuel-vm it may be easier to manage it like this for now.

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: Updated the test to try writing to other non-writable registers as well. However, it only checks for REG_ZERO and REG_ONE since detecting invalid modifications for other registers is more complicated.

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: Changed this test to avoid the bitmath, and now it generates instructions from a random u32 instead. This required broadening the scope of the test to consider writing to other registers besides REG_ONE.

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: done

src/interpreter/executors/instruction/tests/reserved_registers.rs
Voxelot: realized this comment was meant to go on a different thread. This is good to point out that the test relies on the validation order of instructions.

vlopes11 (approved):
LGTM!

Maybe we add an issue to check the remainder registers later on

PR #189, opened 2022-08-05
Add traceable bug id to unexpected behavior

vlopes11:
Some states of the VM are ultimately invalid and contradict the
specifications constraints. These maps to implementation errors that
must be fixed.

In order to make it easier for the user to spot these errors, we must
have a reliable reporting system.

As a first step, its important to map the origin of the error. This
commit introduces `Bug`, that will provide meaningful descriptions to
circumstances that violates the specifications constraints, and `BugId`,
that is guaranteed by tests to map to a specic - and unique - location
of the code.

### DIFF
--- a/tests/bug-id.rs
+++ b/tests/bug-id.rs
@@ -11,10 +11,15 @@ use fuel_vm::prelude::*;
 fn check_bug_id_unique() {
     let mut matches = HashSet::new();
     let re = Regex::new(r"BugId::ID\d{3}").expect("failed to create regex");
+    let re_use = Regex::new(r"BugId::\*").expect("failed to create regex");
 
     for source in glob("src/**/*.rs").expect("Failed to read glob pattern") {
         let source = source.expect("failed to fetch source from glob");
         let source = fs::read_to_string(source).expect("failed to read source");
+
+        if re_use.is_match(&source) {
+            panic!("BugId isn't supposed to be required as '*'")
+        }
 
         re.find_iter(&source).map(|m| m.as_str().to_string()).for_each(|s| {
             if !matches.insert(s.clone()) {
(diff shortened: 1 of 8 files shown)


tests/bug-id.rs
```diff
@@ -11,11 +11,16 @@ use fuel_vm::prelude::*;
 fn check_bug_id_unique() {
     let mut matches = HashSet::new();
     let re = Regex::new(r"BugId::ID\d{3}").expect("failed to create regex");
+    let re_use = Regex::new(r"BugId::\*").expect("failed to create regex");
```
Dentosal, line 14: Why is this a regex? A string with `.contains` should do the job here.

Voxelot:
I'd prefer using backtraces instead of managing these bug ids, thiserror makes it pretty easy https://gist.github.com/mikeando/c03fd62e552ccabcdb894251b59038c7

tests/bug-id.rs
vlopes11: True. Guess I got carried away from using regex haha

https://github.com/FuelLabs/fuel-vm/pull/189/commits/9f6df1d4abc50a3242132e7af506993cad677591

vlopes11:
> I'd prefer using backtraces instead of managing these bug ids, thiserror makes it pretty easy https://gist.github.com/mikeando/c03fd62e552ccabcdb894251b59038c7

Added here as a feature. Does that sound good?

https://github.com/FuelLabs/fuel-vm/pull/189/commits/d291bd85d91e5c70e4c0eb4e1c61016a10a08050

Its a feature because it depends on debug symbols - and if the user opts to not enable it, he can just not activate backtrace

Dentosal:
The code itself looks good to me, but I'm not sure if this is the correct approach here. Deferring to other reviewers.

Voxelot (approved):
i think we should revisit the approach to tracking bugId's once we've spent more time fine tuning our use of backtraces.

PR #193, opened 2022-08-08
Add `TIME` instruction implementation

vlopes11:


src/storage/interpreter.rs
```diff
@@ -24,6 +24,14 @@ pub trait InterpreterStorage:
     /// executed.
     fn block_height(&self) -> Result<u32, Self::DataError>;
 
+    /// Return the timestamp of a given block
+    ///
+    /// This isn't optional because the VM is expected to panic if an invalid block height is
+    /// passed - under the assumption that the block height is consistent, the storage should
+    /// necessarily have the timestamp for the block, unless some I/O error prevents it from
+    /// fetching it.
+    fn timestamp(&self, height: u32) -> Result<Word, Self::DataError>;
```
Voxelot, line 33: If tai64 is important for the spec, would it make sense to make it part of the interface here?

```suggestion
    fn timestamp(&self, height: u32) -> Result<Tai64, Self::DataError>;
```

Voxelot (approved):

src/storage/interpreter.rs
adlerjohn: Hmmm. The fact that it's a TAI-64 time specifically could change in the future, it's not really a hard requirement that must hold for all time.

Issue #170, opened 2022-07-13
Remove any unchecked arithmetic


Voxelot:
In order to prevent overflows and unpredictable behaviors, all arithmetic (esp gas or free balances) should use `checked_*` operations rather than std ops like `AddAssign`. In cases when we know that it is not possible to fail or the context around is valid, then we can ignore or `expect` the operation.

The main idea is to make the code safe and keep API as simple as possible. If the input from the user may invalidate the state/behavior, then we need to return an `Option` or `Error`.

It is better to do this change crate by crate to avoid conflicts with other PRs:
- [ ] `fuel-merkle` #729
- [x] `fuel-crypto`: #725
- [x] `fuel-tx`: #725
- [x] `fuel-types`: #725
- [x] `fuel-vm`: #725

Dentosal:
Just to elaborate on this: there are `checked_*`, `overflowing_*`, `saturating_*` and multiple other variants. The appropriate operation should be used. A comment (e.g. spec link) on why a specific variant was selected with `overflowing_*` or `saturating_*` operations wouldn't hurt either.

For optimization, there are also unstable and unsafe `unchecked_*` operations, which assume overflow cannot occur.

Voxelot:
Looks like clippy has a lint for unchecked math, `cargo clippy -- -D clippy::integer_arithmetic` which results in 135 errors currently.

Once we've addressed this issue, we can add `clippy::integer_arithmetic` to our set of disallowed lints.

MitchTurner:
After spending a couple of days on this, it's not clear the best approach. In terms of capturing the error effectively, it makes sense to return `Result` from any function that has checked arithmetic. Theoretically there is internal logic that could never go awry during healthy behavior, and for those functions we could `#[allow(clippy::arithmetic_side_effects)]`, but in reality its very hard to tell which those are. For the rest of the functions, making them fallible (returning `Result`) clutters up the code a lot.

It was when those `Result`s started bleeding into infallible functions in other crates (from `fuel-merkle` into `fuel-tx`) I really started questioning the value of these changes.

It seems more valuable to have thorough fuzzing/prop testing than be able to catch the error telling us the chain is bricked.

With all that in mind, I'm going to put this story back and work on higher-priority issues.

PR #195, opened 2022-08-16
Implement safe arithmetic for blockchain mod

vlopes11:
Addresses `blockchain.rs` for issue #170

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -226,6 +226,9 @@ pub enum BugId {
     ID004,
     ID005,
     ID006,
+    ID007,
+    ID008,
+    ID009,
 }
 
 /// Traceable bug variants
@@ -239,6 +242,12 @@ pub enum BugVariant {
 
     /// Global gas subtraction has underflow
     GlobalGasUnderflow,
+
+    /// The stack point has overflow
+    StackPointerOverflow,
+
+    /// The frame point has overflow
+    FramePointerOverflow,
 }
 
 impl fmt::Display for BugVariant {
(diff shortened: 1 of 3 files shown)


src/interpreter/blockchain.rs
```diff
@@ -30,66 +31,81 @@ where
...
+            || length > MEM_MAX_ACCESS_SIZE as usize
+            || length > self.params.contract_max_size as usize
         {
             return Err(PanicReason::MemoryOverflow.into());
         }
 
-        if ssp != sp {
-            return Err(PanicReason::ExpectedUnallocatedStack.into());
-        }
+        // compiler will optmize to memset
```
mitch-fuel, line 57: optimize

vlopes11: https://github.com/FuelLabs/fuel-vm/pull/195/commits/d559692b84e3bb6c1840a67806c71b2e92cea9d9

src/error.rs
```diff
@@ -226,6 +226,9 @@ pub enum BugId {
     ID004,
     ID005,
     ID006,
+    ID007,
+    ID008,
+    ID009,
```
adlerjohn, line 231: Why are you adding 3 IDs here but only 2 errors further down?

vlopes11: This ID is a fallback for when the debug symbols are not available for the backtrace. They map to unique locations of the source, and we have 3 new instances:

https://github.com/FuelLabs/fuel-vm/blob/d559692b84e3bb6c1840a67806c71b2e92cea9d9/src/interpreter/blockchain.rs#L93-L99

https://github.com/FuelLabs/fuel-vm/blob/d559692b84e3bb6c1840a67806c71b2e92cea9d9/src/interpreter/blockchain.rs#L102-L105

https://github.com/FuelLabs/fuel-vm/blob/d559692b84e3bb6c1840a67806c71b2e92cea9d9/src/interpreter/blockchain.rs#L420-L422

vlopes11: We are being over-cautious now, even with checks that logically can't panic. The idea is to reduce these checks when we enter the optimization phase

adlerjohn (commented):
looks fine to me, but will defer to client team

mitch-fuel (approved):
I have tested all of these checks in my associated PR, approved

Issue #96, opened 2022-03-23
Add log `tracing`


vlopes11:
The consumers of `fuel-vm` [use](https://github.com/FuelLabs/fuel-core/blob/cca6ac155b7a7c85ea6cd6043d8f9cd64d4fa76d/fuel-core/Cargo.toml#L61) [tracing](https://crates.io/crates/tracing) as log mechanism

We should add tracing with `DEBUG` and `TRACE` levels to track the VM execution

Use the following criteria:
- after any mutation, log with `debug`
- upon any request, log with `trace`

PR #196, opened 2022-08-24
Add tracing to op codes and show register values

freesig:
Adds tracing to the vm's `instruction` call.
Output looks like:
image

Related to #96


vlopes11 (changes requested):
Looks good overall. A couple of last dets:

- Add related issue: https://github.com/FuelLabs/fuel-vm/issues/96 (note we are not solving the issue, just incrementally adding steps to that direction)
- Protect the tracing via feature

freesig:
@vlopes11 I wonder if we can use the tracing [compile time filters](https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters) to disable tracing with a feature. Then we don't have to put `#[cfg(feature = "tracing")]` all through the code.

Voxelot (approved):

Voxelot (approved):

xgreenx (approved):

PR #199, opened 2022-08-25
Remove transmute

freesig:
Removes a transmute that is never actually used.
Also this transmute crates a self reference that is UB when the struct is moved.

### DIFF
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -267,6 +267,7 @@ where
         let mut interpreter = Interpreter::with_storage(storage, params);
         interpreter
             .transact(tx)
+            .map(|state| ProgramState::from(state))
             .map(|state| StateTransition::new(state, interpreter.tx.into(), interpreter.receipts))
     }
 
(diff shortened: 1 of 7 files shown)


src/interpreter/executors/main.rs
```diff
@@ -264,25 +264,27 @@ where
...
 
-        let state = state_result?;
-
-        let transition = StateTransitionRef::new(state, self.transaction(), self.receipts());
+        state_result
+    }
 
-        Ok(transition)
+    /// Get the state transaction from a new program state.
+    pub fn state_transition<'a>(&'a self, state: &'a ProgramState) -> StateTransitionRef<'a> {
```
xgreenx, line 287: Now the API looks not consistent=( First, you need to call `transact`, and after with the output, you need to call `state_transition`.

It seems that `StateTransitionRef` still can own `ProgramState` as before, but `Transactor` should work with `ProgramState`.

In this case, API will be consistent, and we will remove `unsafe`=)
But you need to add a separate method like `transact_internal` that returns `ProgramState`. 

Also maybe `StateTransitionRef` can be generic like `StateTransitionRef<'a, Generic>` where in some cases it is `&'a PrgoramState` in some `ProgramState`(it will remove unnecessary `clone` in some cases).

freesig: I don't think that's possible though because the `ProgramState` will be created in the `transact_internal` and we can't return a reference to `ProgramState` when it's a local variable in the function.

xgreenx: `StateTransitionRef<'a, ProgramState>` will own `ProgramState` inside=) `transact_internal` it is for `StateTransaciton`

freesig: @xgreenx I just pushed some changes, does that fix your concern?

xgreenx (approved):

PR #200, opened 2022-08-25
Refactored usage of storage traits

xgreenx:
Applied changes regarding new storage traits from https://github.com/FuelLabs/fuel-storage/pull/15.
Moved `storage.rs` to the `storage` folder and renamed it to `mod.rs`(did the same for `state` and `interpreter`)
Use of new syntax to work with `Storage`.
Defined tables(`ContractsRawCode`, `ContractsState`, `ContractsInfo`, `ContractsAssets`) to simplify and unify work with them in `fuel-core`. It prevents us from using wrong types during work with storage and makes code readeable.


src/storage/mod.rs
```diff
@@ -0,0 +1,56 @@
+//! Storage backend implementations.
```
vlopes11, line 1: Almost every crate uses a different mod style. Since this is an issue of its own, let split the scope of changes into different PRs and keep the current project structure in this one - focusing only in the storage changes

vlopes11 (changes requested):

src/storage/mod.rs
Voxelot: mod.rs is rust 2015 style which we are trying to avoid.

xgreenx: Usage the `mod.rs` doesn't mean that you are using an outdated style=) Rust provides an alternative way with `module_name.rs` because it can be easy for Rust newcomers.

I prefer `mod.rs`, because it allows the encapsulation of all the module's stuff in one folder. It is very helpful in cases when you have many modules. Because it removes "dust" files from the top of the crate.

As an example, `fuel-core/fuel-core/src`:

image

It can be simplified into:
image

You can see that top of the `src` folder with the `mod.rs` style contains files only with custom logic not related to any modules(they belong only to `src`). It helps you better understand the relationship of files/modules with one look at the folder tree.

Of course, it is only my preference, and if everybody likes the approach with `module_name.rs`, I will use it=) But I don't see pros using `module_name.rs` for crates with many modules.

Voxelot: This is bikeshedding and we've already adopted the rust 2018 style, I don't really want to debate all these preferences constantly.

Voxelot (approved):
looks cleaner to me, :shipit:

src/storage.rs
```diff
@@ -1,9 +1,58 @@
...
+    /// `Salt` - is the salt used during creation of the contract for uniques.
+    /// `Byte32` - is the root hash of the contract's code.
+    type SetValue = (Salt, Bytes32);
+    type GetValue = Self::SetValue;
+}
+
+/// The storage table for contract's assets balances.
+///
+/// Lifetime is for optimization to avoid `clone`.
+pub struct ContractsAssets<'a>(PhantomData<&'a ()>);
```
vlopes11, line 39: Can't this be a tuple of generics that implements either `AsRef<ContractId>` or `Borrow<ContractId>`? Introducing lifetime requirements to the signature might increase the complexity of the implementation and I think it should be avoided whenever possible so we don't confuse the compiler about ownership requirements

src/storage/predicate.rs
```diff
@@ -16,104 +15,34 @@ use fuel_types::{Address, AssetId, Bytes32, ContractId, Salt};
 #[derive(Debug, Default, Clone, Copy)]
 pub struct PredicateStorage;
 
-impl Storage<ContractId, Contract> for PredicateStorage {
+impl<Type: Mappable> StorageInspect<Type> for PredicateStorage {
```
vlopes11, line 18: The storage predicate is a special case of a storage that shouldn't provide a functionality. Since we are refactoring the storage, do you think we can come with a more robust solution to that problem? Its ofc not a blocker for this PR, but its a tech-debt that will have to be addressed either now or later

vlopes11 (commented):
Couple of questions/suggestions

src/storage.rs
xgreenx: It can be, but in this case, the definition will look like this:

image

And you need to propagate those generics everywhere. Instead of `StorageMutate<ContractsAssets<'_>>`, you need to write `StorageMutate<ContractsAssets<C, A>>`, and `C` and `A` should be generics with bounds.

Plus compiler will not force you to use reference, and sometimes you can use `Copy` in places where you don't want to do that. 

But opposite it, you can have cases where you want to pass by reference and pass by value. In these cases, generic with `Borrow` will be better.

The usage of an inferred lifetime(`'_`) almost in all cases will work for you(I guess in all, but who knows). The same way references in the methods work. So you will not have errors related to it. Also, you can write `self.storage::<ContractsAsset>()`, and it works.

So, if we want to support pass by reference and value, I can use generics(but it means propagation of them). If not, the usage of a lifetime `'_` will not harm us=)

src/storage/predicate.rs
xgreenx: But I guess the bytecode of the predicate may contain the opcodes with access to the storage? (If the user compiled it without using or modified sway?) And in these cases, we need to return an error. So the current implementation is correct.

bvrooman (approved):

bvrooman (approved):

Voxelot (approved):

src/storage.rs
vlopes11: I'm fine going with the reference, but I'd rather go with the generic+borrow because this is more canonical to the language & more flexible (we will be able to use anything that can be a contract id; from Cow to references).

You're right, the `'_` should cover most of the cases

src/storage/predicate.rs
vlopes11: Yes, the proposed change extends the previous implementation so we end up with the same behavior (e.g. returning an error if an access to the storage is attempted).

However, that implementation was a quick&dirty way to have a storage that will also "support predicates" because originally the storage was a requirement for the interpreter. But then we have a mode of the interpreter that shouldn't implement anything related to the storage (i.e. predicate mode) so it will basically just error anything that comes.

If a predicate ends up calling a storage function, then that is a bug in the VM. We should halt any instance of a contract opcode instead of attempting to call the storage (and failing). In other words; yes, it is correct to say that we should error for any predicate verification, but it was implemented like that just to avoid the increased work of making the storage optional for the interpreter.

But as mentioned, this is not really a blocker for this PR, it was just mentioned as window of opportunity to improve things further.

vlopes11 (approved):

src/storage/predicate.rs
xgreenx: Do you have a separate check in the `fuel-vm` that the byte code of the predicate doesn't contain any storage opcode?

Voxelot (approved):

Issue #201, opened 2022-09-01
Test Arithemetic Overflow Exceptions in Blockchain.rs


mitch-fuel:
Test the conditionals for safe arithmetic checks in `Blockchain.rs` under issue #170

PR #202, opened 2022-09-01
Test safe arithmetic checks in blockchain.rs

mitch-fuel:
Closes issue #201 

Depends on issue #195

### DIFF
--- a/src/interpreter/blockchain.rs
+++ b/src/interpreter/blockchain.rs
@@ -91,6 +91,7 @@ where
         memory.copy_from_slice(code);
 
         self.registers[REG_SP]
+            //TODO this is looser than the compare against [REG_HP,REG_SSP+length]
             .checked_add(length as Word)
             .map(|sp| {
                 self.registers[REG_SP] = sp;
--- a/src/interpreter/blockchain.rs
+++ b/src/interpreter/blockchain.rs
@@ -54,7 +54,7 @@ where
             return Err(PanicReason::MemoryOverflow.into());
         }
 
-        // compiler will optmize to memset
+        // compiler will optimize to memset
         self.memory[memory_offset..memory_offset_end]
             .iter_mut()
             .for_each(|m| *m = 0);
--- a/src/interpreter/blockchain.rs
+++ b/src/interpreter/blockchain.rs
@@ -150,7 +150,11 @@ where
         let (cd, of) = c.overflowing_add(d);
         let overflow = overflow || of;
 
-        if overflow || d > MEM_MAX_ACCESS_SIZE || a > VM_MAX_RAM.saturating_sub(d) || bx > VM_MAX_RAM || cd > VM_MAX_RAM
+        if overflow
+            || d > MEM_MAX_ACCESS_SIZE
+            || a > VM_MAX_RAM.saturating_sub(d)
+            || bx > VM_MAX_RAM
+            || cd > VM_MAX_RAM
         {
             return Err(PanicReason::MemoryOverflow.into());
         }
--- a/src/interpreter/blockchain.rs
+++ b/src/interpreter/blockchain.rs
@@ -217,7 +217,9 @@ where
         let (bx, of) = b.overflowing_add(ContractId::LEN as Word);
         let overflow = overflow || of;
 
-        if overflow || ax > VM_MAX_RAM || bx > VM_MAX_RAM {
+        if overflow
+            || ax > VM_MAX_RAM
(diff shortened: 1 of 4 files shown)


tests/blockchain.rs
```diff
@@ -1,14 +1,15 @@
 use fuel_crypto::{Hasher, SecretKey};
-use fuel_tx::{StorageSlot, TransactionBuilder};
+use fuel_tx::TransactionBuilder;
 use fuel_types::bytes;
-use fuel_vm::script_with_data_offset;
-use fuel_vm::util::test_helpers::TestBuilder;
+// use fuel_vm::script_with_data_offset;
```
vlopes11, line 4: Idk if it was the intent, but we need to indeed remove these manual offsets since now we can trivially fetch them with `GTF`. Anyway, I'm very happy to see the overall coverage improvement

mitch-fuel: I certainly didn't intend to have this comment left behind. I removed the commented uses.

tests/blockchain.rs
```diff
@@ -410,69 +409,700 @@ fn load_external_contract_code() {
     }
 }
 
-#[test]
-fn can_read_state_from_initial_storage_slots() {
-    // the initial key and value pair for the contract
-    let key = Hasher::hash(b"initial key");
-    let value = [128u8; 32].into();
+fn ldc_reason_helper(cmd: Vec<Opcode>, expected_reason: PanicReason, should_patch_jump: bool) {
```
vlopes11, line 412: nit: `#[test]` directive is missing

vlopes11 (changes requested):
looks good overall; small nit

tests/blockchain.rs
mitch-fuel: this isn't a test, it is a helper function used by a test. Is there a way to mark something as FOR a test but not to be executed independently?

Dentosal: `#[cfg(test)]` makes a function only available when running tests

Voxelot: although, since this function is defined inside the `tests` directory, it will only ever be available within a testing context. Cargo ignores anything under `#[cfg(test)]`, `#[test]` or in the `/tests` dir when publishing a crate.

Voxelot: It would be cleaner without the annotations IMO, they aren't needed here. While not a big deal, we probably don't want this to become a pattern repeated throughout the codebase.

Voxelot (approved):

Voxelot (approved):

src/interpreter/blockchain.rs
```diff
@@ -100,9 +102,7 @@ where
 
         // update frame pointer, if we have a stack frame (e.g. fp > 0)
         if fp > 0 {
-            let fpx = fp
-                .checked_add(CallFrame::code_size_offset())
-                .ok_or_else(|| Bug::new(BugId::ID008, BugVariant::FramePointerOverflow))?;
+            let fpx = checked_add_usize(fp, CallFrame::code_size_offset())?;
```
vlopes11, line 105: This can't overflow unless we have a bug. Consider using `add_usize` instead. If the `$fp` will always point to a valid memory address, and we can't overflow `usize` or `u64` with a valid memory address + frame offset.

Consider using `add_usize` instead.
https://github.com/FuelLabs/fuel-vm/blob/2053b2fd2f2e874670a11cac4158c7ec88397576/src/arith.rs#L73

 otherwise, might be worthy to add a new function to `arith.rs` that will fail only if we have a bug, using the same signature as `add_word`

https://github.com/FuelLabs/fuel-vm/blob/2053b2fd2f2e874670a11cac4158c7ec88397576/src/arith.rs#L39

But then this function will contain a hint that, whenever it is reached, it will necessarily represent a bug in the code - so we output an according error message

vlopes11 (changes requested):
last nit

src/interpreter/blockchain.rs
vlopes11: Could be smth like `add_usize_infallible(usize, usize) -> Result<usize, RuntimeError>`

tests/blockchain.rs
vlopes11: Agreed; if it isn't a test case, then further directives aren't required

src/interpreter/blockchain.rs
mitch-fuel: I agree and have modified this to use add_usize

vlopes11 (approved):

PR #203, opened 2022-09-06
Apply safe arith to call and crypto

vlopes11:
This commit applies safe arithmetic to call and crypto interpreter
modules.

The deprecated implementation of secp256k1 was not using safe
arithmetic. Instead of updating it, the module was deprecated in favor
of fuel-crypto.

### DIFF
--- a/src/call.rs
+++ b/src/call.rs
@@ -53,7 +53,7 @@ impl Call {
 
 impl SizedBytes for Call {
     fn serialized_size(&self) -> usize {
-        WORD_SIZE.saturating_mul(2).saturating_add(ContractId::LEN)
+        ContractId::LEN + 2 * WORD_SIZE
     }
 }
 
@@ -148,34 +148,22 @@ impl CallFrame {
 
     /// Contract code memory offset.
     pub const fn code_offset() -> usize {
-        Self::code_size_offset().saturating_add(WORD_SIZE)
+        Self::code_size_offset() + WORD_SIZE
     }
 
     /// Contract code size memory offset.
     pub const fn code_size_offset() -> usize {
-        VM_REGISTER_COUNT
-            .saturating_add(2)
-            .saturating_mul(WORD_SIZE)
-            .saturating_add(ContractId::LEN)
-            .saturating_add(AssetId::LEN)
+        ContractId::LEN + AssetId::LEN + WORD_SIZE * (2 + VM_REGISTER_COUNT)
     }
 
     /// `a` argument memory offset.
     pub const fn a_offset() -> usize {
-        VM_REGISTER_COUNT
-            .saturating_add(1)
-            .saturating_mul(WORD_SIZE)
-            .saturating_add(ContractId::LEN)
-            .saturating_add(AssetId::LEN)
+        ContractId::LEN + AssetId::LEN + WORD_SIZE * (1 + VM_REGISTER_COUNT)
     }
 
     /// `b` argument memory offset.
     pub const fn b_offset() -> usize {
-        VM_REGISTER_COUNT
-            .saturating_add(2)
-            .saturating_mul(WORD_SIZE)
-            .saturating_add(ContractId::LEN)
-            .saturating_add(AssetId::LEN)
+        ContractId::LEN + AssetId::LEN + WORD_SIZE * (2 + VM_REGISTER_COUNT)
     }
 
     /// Registers prior to the called execution.
@@ -211,7 +199,7 @@ impl CallFrame {
 
 impl SizedBytes for CallFrame {
     fn serialized_size(&self) -> usize {
-        Self::code_offset().saturating_add(bytes::padded_len(self.code.as_ref()))
+        Self::code_offset() + bytes::padded_len(self.code.as_ref())
     }
 }
 
--- a/src/call.rs
+++ b/src/call.rs
@@ -53,7 +53,7 @@ impl Call {
 
 impl SizedBytes for Call {
     fn serialized_size(&self) -> usize {
-        ContractId::LEN + 2 * WORD_SIZE
+        WORD_SIZE.saturating_mul(2).saturating_add(ContractId::LEN)
     }
 }
 
@@ -148,22 +148,34 @@ impl CallFrame {
 
     /// Contract code memory offset.
     pub const fn code_offset() -> usize {
-        Self::code_size_offset() + WORD_SIZE
+        Self::code_size_offset().saturating_add(WORD_SIZE)
     }
 
     /// Contract code size memory offset.
     pub const fn code_size_offset() -> usize {
-        ContractId::LEN + AssetId::LEN + WORD_SIZE * (2 + VM_REGISTER_COUNT)
+        VM_REGISTER_COUNT
+            .saturating_add(2)
+            .saturating_mul(WORD_SIZE)
+            .saturating_add(ContractId::LEN)
+            .saturating_add(AssetId::LEN)
     }
 
     /// `a` argument memory offset.
     pub const fn a_offset() -> usize {
-        ContractId::LEN + AssetId::LEN + WORD_SIZE * (1 + VM_REGISTER_COUNT)
+        VM_REGISTER_COUNT
+            .saturating_add(1)
+            .saturating_mul(WORD_SIZE)
+            .saturating_add(ContractId::LEN)
+            .saturating_add(AssetId::LEN)
     }
 
     /// `b` argument memory offset.
     pub const fn b_offset() -> usize {
-        ContractId::LEN + AssetId::LEN + WORD_SIZE * (2 + VM_REGISTER_COUNT)
+        VM_REGISTER_COUNT
+            .saturating_add(2)
+            .saturating_mul(WORD_SIZE)
+            .saturating_add(ContractId::LEN)
+            .saturating_add(AssetId::LEN)
     }
 
     /// Registers prior to the called execution.
@@ -199,7 +211,7 @@ impl CallFrame {
 
 impl SizedBytes for CallFrame {
     fn serialized_size(&self) -> usize {
-        Self::code_offset() + bytes::padded_len(self.code.as_ref())
+        Self::code_offset().saturating_add(bytes::padded_len(self.code.as_ref()))
     }
 }
 
@@ -249,7 +261,10 @@ impl io::Write for CallFrame {
         let (b, buf) = unsafe { bytes::restore_word_unchecked(buf) };
 
         let (bytes, code, _) = bytes::restore_raw_bytes(buf, code_len)?;
-        n += bytes;
+
+        n = n
+            .checked_add(bytes)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "io bytes overflow"))?;
 
         self.to = to.into();
         self.asset_id = asset_id.into();
--- a/src/call.rs
+++ b/src/call.rs
@@ -7,6 +7,7 @@ use fuel_tx::Contract;
 use fuel_types::bytes::{self, SizedBytes};
 use fuel_types::{AssetId, ContractId, Word};
 
+use crate::arith::checked_add_usize;
 use std::io::{self, Write};
 use std::mem;
 
@@ -250,9 +251,7 @@ impl io::Write for CallFrame {
 
         let (bytes, code, _) = bytes::restore_raw_bytes(buf, code_len)?;
 
-        n = n
-            .checked_add(bytes)
-            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "io bytes overflow"))?;
+        n = checked_add_usize(n, bytes)?;
 
         self.to = to.into();
         self.asset_id = asset_id.into();
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -16,7 +16,7 @@ where
 
 #[test]
 #[cfg(feature = "random")]
-fn ephemeral_merkle_root_works() {
+fn ephemeral_merkle_root_returns_the_expected_root() {
     use fuel_crypto::Hasher;
     use rand::rngs::StdRng;
     use rand::{Rng, SeedableRng};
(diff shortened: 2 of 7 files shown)


Voxelot (approved):

Dentosal:
Many `saturating_addr` operations are only performed on saturating `usize`. I feel like we should really panic, since the in inputs in such cases are surely incorrect. I'm afraid that using `saturating_add` in such cases might hide critical security bugs.

Voxelot:
> Many `saturating_addr` operations are only performed on saturating `usize`. I feel like we should really panic, since the in inputs in such cases are surely incorrect. I'm afraid that using `saturating_add` in such cases might hide critical security bugs.

Makes sense, especially in methods that already have a Result, there's no reason not to use checked math and raise a bug code. Defensive coding like this helps protect against future refactoring where some of the expected validation checks are incorrectly altered.

Voxelot (changes requested):
revoking approval in favor of using the more defensive checked_add instead (when a result type is available).

src/interpreter/crypto.rs
```diff
@@ -59,12 +61,15 @@ impl<S> Interpreter<S> {
...
+            || b > VM_MAX_RAM.saturating_sub(c)
+        {
             return Err(PanicReason::MemoryOverflow.into());
         }
 
         let (a, b, c) = (a as usize, b as usize, c as usize);
-        let bc = b + c;
+        let bc = b.saturating_add(c);
 
         self.try_mem_write(a, Hasher::hash(&self.memory[b..bc]).as_ref())?;
```
Voxelot, line 74: Consider additionally using `.get()` instead of directly slicing to avoid potential rust level panics.

```suggestion
        self.try_mem_write(a, Hasher::hash(&self.memory.get(b..bc).ok_or(.. some bug error type ..)).as_ref())?;
```

vlopes11: We need to keep in mind that 95% of these cases are not runtime, such as https://github.com/FuelLabs/fuel-vm/pull/203/files#diff-df812721f376b9b2f9599469028b29653b7341d65fa823c5b4d8bfd2127d5032R11

In fact, using checked arith there is maybe even overkill. Its just applied because clippy isn't smart enough to differentiate between what is fixed by the compiler and what is resolved at runtime. I'll recheck the ones that are runtime and might overflow

vlopes11:
Seems checked add can't be used everywhere.

This code for example can't be compiled because `expect` is not a part of the `const` set of the language (it is supposed to be add somewhere in the near future tho)

```
    /// Contract code memory offset.
    pub const fn code_offset() -> usize {
        Self::code_size_offset()
            .checked_add(WORD_SIZE)
            .expect("compile-time arithmetic")
    }
```

Seems we have two options here:

- Leave the unchecked arith as it is exclusively for compile-time resolved numbers until `expect` is part of the `const` subset. This means we won't be able to add checked arith clippy lint in the CI (I don't want to pollute everywhere in the code with clippy allows)

- use saturating_add/sub/mul until we have the support of expect in stable

Voxelot:
I agree that checked everywhere is overkill in some cases. Especially when we are just adding other constant numbers together. 

For me, the checked arithmetic is mainly a concern when we are dealing with dynamic variables that could be altered either by user input or other parts of the VM at runtime (e.g. like our cgas calculation bug). Generally speaking (although maybe not comprehensive) this would involve using checked math when operating on data from the following sources:

- length of predicates, scripts, script data, transaction size, witnesses
- asset balances
- any time we access register values
- gas usage

Generally, these operations will all be done in a non-const context and performed at runtime during init, instruction exec, or post-execution. Since all of these areas are considered to be fallible, it will be low impact on the code (i.e. no refactor or complex error handling). Since the cost is minimal, I'd rather err on the side of checked arithmetic during these kinds of runtime operations. That way we don't risk saturating a gas refund amount for example.

src/call.rs
```diff
@@ -53,7 +53,7 @@ impl Call {
 
 impl SizedBytes for Call {
     fn serialized_size(&self) -> usize {
-        ContractId::LEN + 2 * WORD_SIZE
+        WORD_SIZE.saturating_mul(2).saturating_add(ContractId::LEN)
```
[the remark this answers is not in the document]
Voxelot: I think the risk is fairly low here, since these values are all known at compile time and are constant. These constants translate into `8 * 2 + 32`, which seems a little overkill to add checked math error handling around.

vlopes11:
@Voxelot sounds good! We just need to keep in mind also that saturated gas is potentially unfeasible in terms of available money in the world. In other words, no one can setup a transaction with a base asset with `u64::MAX` gas with a normal price factor, unless he severely undercharge the gas via a super big inverse gas factor conversion (and this might lead to no block validator picking this tx from the pool). But then it doesn't hurt to add this guarantee and in that case we are effectively increasing the robustness of the system, so the approach sounds good!

The original intent was to enable the check on the CI, but then we might need some tweaking from clippy side to achieve that (e.g. safe arith check only for operations that are runtime resolved)

src/interpreter/crypto.rs
```diff
@@ -18,20 +17,25 @@ impl<S> Interpreter<S> {
 
         let (a, b, c) = (a as usize, b as usize, c as usize);
 
-        let bx = b + Bytes64::LEN;
-        let cx = c + Bytes32::LEN;
+        let bx = b
+            .checked_add(Bytes64::LEN)
+            .ok_or_else(|| Bug::new(BugId::ID010, BugVariant::CheckedRegisterOverflow))?;
```
mitch-fuel, line 22: in my testing I realized that this and subsequent similar checks are only reachable if VM_MAX_RAM is nearly MAX. In the blockchain.rs test file we did these computations before the above checks. Would we like to leave these as unreachable given sane values of VM_MAX_RAM or rearrange so that all checks can be tested?

mitch-fuel: I think for this type of situation we should perform these checks before the register comparisons to VM_MAX_RAM. The calculations are just rearrangements of the preceding checks and could be used directly  as `bx > VM_MAX_RAM` instead of `b > VM_MAX_RAM - Bytes64::LEN`

vlopes11: I agree there. We are checking some stuff that is logically guarantee to be sound. It is indeed overkill and we might want to reconsider that. Since we already agreed that safe arithmetic isn't applicable everywhere, then we might be able to be flexible here, too.

Regarding your question about `VM_MAX_RAM`; yes, you are correct. If this constant is set properly, then this panic is unreachable

src/interpreter/crypto.rs
```diff
@@ -1,37 +1,35 @@
...
-        let sig = &self.memory[b..bx];
+        // Safety: memory bounds are checked
+        let signature = unsafe { Signature::as_ref_unchecked(&self.memory[b..bx]) };
+        let message = unsafe { Message::as_ref_unchecked(&self.memory[c..cx]) };
 
-        match crypto::secp256k1_sign_compact_recover(sig, e) {
-            Ok(pk) => {
-                self.try_mem_write(a, pk.as_ref())?;
+        match signature.recover(message) {
+            Ok(public) => {
```
leviathanbeak, line 27: nit: `public` is not the clearest name, only from previous `pk` name was able to decipher what it is
maybe: `pk`, `pub_key`, `public_key` ?

mitch-fuel: absolutely, will change

mitch-fuel: fixed

leviathanbeak:
apart from a single nit that was fixed, nothing else on my side to add - deferring to others involved in the review

Dentosal:
LGTM, deferring to named reviewers

src/interpreter/crypto.rs
```diff
@@ -1,37 +1,35 @@
...
 impl<S> Interpreter<S> {
     pub(crate) fn ecrecover(&mut self, a: Word, b: Word, c: Word) -> Result<(), RuntimeError> {
-        if a > VM_MAX_RAM - Bytes64::LEN as Word
-            || b > VM_MAX_RAM - Bytes64::LEN as Word
-            || c > VM_MAX_RAM - Bytes32::LEN as Word
-        {
+        let bx = checked_add_word(b, Bytes64::LEN as Word)?;
+        let cx = checked_add_word(c, Bytes32::LEN as Word)?;
+
+        if a > VM_MAX_RAM - Bytes64::LEN as Word || bx > MIN_VM_MAX_RAM_USIZE_MAX || cx > MIN_VM_MAX_RAM_USIZE_MAX {
```
vlopes11, line 16: Seems we might need a `pub const fn sub_usize(a: usize, b: usize) -> usize` in `arith` too so we can enable clippy unsafe arith check

mitch-fuel: we already had a checked_sub_usize and I _just_ added it to this file

src/interpreter/crypto.rs
```diff
@@ -1,37 +1,38 @@
...
 
+use crate::arith::{checked_add_word, checked_sub_word};
 use fuel_asm::PanicReason;
-use fuel_crypto::Hasher;
+use fuel_asm::PanicReason::ArithmeticOverflow;
+use fuel_crypto::{Hasher, Message, PublicKey, Signature};
 use fuel_types::{Bytes32, Bytes64, Word};
 
 impl<S> Interpreter<S> {
     pub(crate) fn ecrecover(&mut self, a: Word, b: Word, c: Word) -> Result<(), RuntimeError> {
```
xgreenx, line 12: I think better to have more descriptive naming.
- 'a' -> `out_ptr`
- 'b' -> `sig_ptr`
- 'c' -> `m_ptr`

Also, instead of `Word`, it may be better to introduce the type as `Ptr` and `PtrMut`. In this case we don't need to add `ptr` suffix=)

src/interpreter/crypto.rs
```diff
@@ -1,37 +1,38 @@
...
-        if a > VM_MAX_RAM - Bytes64::LEN as Word
-            || b > VM_MAX_RAM - Bytes64::LEN as Word
-            || c > VM_MAX_RAM - Bytes32::LEN as Word
+        let bx = checked_add_word(b, Bytes64::LEN as Word)?;
```
xgreenx, line 13: We have duplicated this logic in several places.

We can add to `Interpreter` two new methods, `read_as` and `write`.

`read_as` will be generic and accept types that implement the `as_ref_unchecked` method(we need to move this method to the trait).

You will be able to write code like:

```rust
let signature: Signature = self.read_as(b)?;
let message: Message = self.read_as(c)?;
```

src/interpreter/crypto.rs
```diff
@@ -42,12 +43,16 @@ impl<S> Interpreter<S> {
     pub(crate) fn keccak256(&mut self, a: Word, b: Word, c: Word) -> Result<(), RuntimeError> {
```
xgreenx, line 43: The same here, more descriptive argument names=)

src/interpreter/crypto.rs
mitch-fuel: If we do this (better names are always good), we should do this outside of the function parameters as these letters are mapped to the opcode argument names from the spec.

bvrooman (dismissed):

src/interpreter/crypto.rs
xgreenx: Maybe we also need to do it for opcodes, but I think we can start from `Interpreter`. Because the context of those arguments is important here=) How opcodes transfer them is not so important.

tests/crypto.rs
```diff
@@ -198,79 +113,35 @@ fn keccak256() {
...
+    let script_data = message.iter().copied().chain(hash.iter().copied()).collect();
+
+    let tx = TransactionBuilder::script(script, script_data)
+        .gas_price(gas_price)
+        .gas_limit(gas_limit)
+        .maturity(maturity)
+        .finalize_checked(height, &params);
+
+    let receipts = client.transact(tx);
+    let success = receipts.iter().any(|r| matches!(r, Receipt::Log{ ra, .. } if ra == &1));
```
bvrooman, line 144: nit: Instead of comparing address with `ra`, can we compare values, e.g.:

```suggestion
    let success = receipts.iter().any(|r| matches!(r, Receipt::Log{ ra, .. } if *ra == 1));
```

mitch-fuel: fixed all of these, agreed

src/crypto.rs
```diff
@@ -59,86 +14,54 @@ where
...
-#[cfg(all(test, feature = "random"))]
-mod tests {
-    use crate::crypto::{
-        ephemeral_merkle_root, secp256k1_sign_compact_recover, secp256k1_sign_compact_recoverable, Secp256k1,
-    };
-    use crate::prelude::*;
-
+#[test]
+#[cfg(feature = "random")]
+fn ephemeral_merkle_root_works() {
```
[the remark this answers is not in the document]
mitch-fuel: fixed, created an issue to do a broader sweep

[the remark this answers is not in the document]
mitch-fuel: https://github.com/FuelLabs/fuel-vm/issues/221

src/interpreter/crypto.rs
mitch-fuel: created an issue to fix this broadly
https://github.com/FuelLabs/fuel-vm/issues/220

src/interpreter/crypto.rs
mitch-fuel: created issue https://github.com/FuelLabs/fuel-vm/issues/220 to address this broadly

src/interpreter/crypto.rs
mitch-fuel: created issue #222

mitch-fuel (approved):

src/call.rs
```diff
@@ -7,6 +7,7 @@ use fuel_tx::Contract;
 use fuel_types::bytes::{self, SizedBytes};
 use fuel_types::{AssetId, ContractId, Word};
```
Voxelot, line 9:
```suggestion
```

Voxelot (approved):

PR #207, opened 2022-09-09
Tests for Safe Arithmetic Checks in crypto.rs

mitch-fuel:
Continues work of #170 

Must not be merged until after #203


src/util.rs
```diff
@@ -368,6 +368,59 @@ pub mod test_helpers {
             receipts[0].ra().expect("Balance expected")
         }
     }
+
+    pub fn check_expected_reason_for_opcodes(opcodes: Vec<Opcode>, expected_reason: PanicReason) {
```
vlopes11, line 372: nit: consider taking `opcodes: &[Opcode]` instead of owned vec so it doesn't have to be cloned.

If we need owned opcodes, then it might be preferable to take an iterator of opcodes because this way you can also pass vecs.

vlopes11 (approved):
looks good; optional nit

src/util.rs
mitch-fuel: created issue #226

PR #211, opened 2022-09-12
Safe Arithmetic Implementation for contract.rs

mitch-fuel:
Part of #170.

Adds checked arithmetic functions to `contract.rs`

### DIFF
--- a/src/interpreter/contract.rs
+++ b/src/interpreter/contract.rs
@@ -1,6 +1,6 @@
 use super::Interpreter;
 use crate::consts::*;
-use crate::error::{Bug, BugId, BugVariant, RuntimeError};
+use crate::error::RuntimeError;
 use crate::storage::InterpreterStorage;
 
 use fuel_asm::{PanicReason, RegisterId, Word};
--- a/src/interpreter/contract.rs
+++ b/src/interpreter/contract.rs
@@ -8,6 +8,7 @@ use fuel_tx::{Contract, Output, Receipt};
 use fuel_types::{Address, AssetId, ContractId};
 
 use std::borrow::Cow;
+use std::cmp;
 
 impl<S> Interpreter<S>
 where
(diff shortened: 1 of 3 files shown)


src/interpreter/contract.rs
```diff
@@ -23,15 +23,23 @@ where
     pub(crate) fn contract_balance(&mut self, ra: RegisterId, b: Word, c: Word) -> Result<(), RuntimeError> {
         Self::is_register_writable(ra)?;
 
-        if b > VM_MAX_RAM - AssetId::LEN as Word || c > VM_MAX_RAM - ContractId::LEN as Word {
+        let bx = b
+            .checked_add(AssetId::LEN as Word)
+            .ok_or_else(|| Bug::new(BugId::ID014, BugVariant::CheckedRegisterOverflow))?;
```
vlopes11, line 28: This one is not a bug but a malformed input. The bugs are reserved for things who are already checked or guaranteed by the specs to not happen. In this case, the user can just send a bad `b`

vlopes11 (changes requested):

src/interpreter/contract.rs
mitch-fuel: this has been modified

src/interpreter/contract.rs
```diff
@@ -23,15 +24,27 @@ where
...
+            .checked_add(AssetId::LEN as Word)
+            .ok_or_else(|| PanicReason::ArithmeticOverflow)?;
+
+        let cx = c
+            .checked_add(ContractId::LEN as Word)
+            .ok_or_else(|| PanicReason::ArithmeticOverflow)?;
+
+        //if bx or cx is above usize::MAX then it cannot be safely cast to usize,
+        // check the tighter bound
+        let min_comparison_value = cmp::min(VM_MAX_RAM, usize::MAX as Word);
```
vlopes11, line 37: consider declaring this a `const` because it will be compiler to a constant anyway but then we see that happening in the code.

Also, this check might be overkill because `usize` is `u32` minimum and our VM memory won't ever be close to that value. However, doesn't hurt to check!

vlopes11 (dismissed):
lgtm, few suggestions for improvement

vlopes11 (approved):

PR #217, opened 2022-09-15
Add `arith` module

### LINKED ISSUE #214: Add `arithmetic` module
Create a new `arithmetic` module that will contain inlined functions:

### Fallible

Will return `PanicReason::ArithmeticOverflow` if fails

`fn checked_add_word(a: Word, b: Word) -> Result<Word, PanicReason>`
`fn checked_add_usize(a: usize, b: usize) -> Result<usize, PanicReason>`
`fn checked_sub_word(a: Word, b: Word) -> Result<Word, PanicReason>`
`fn checked_sub_usize(a: usize, b: usize) -> Result<usize, PanicReason>`

These functions are meant to be used with user inputs that might be fallible per protocol.

### Unfallible

Will return `Ok(_)` if a new feature `optimized` is enabled. Otherwise, will return a bug identification

`fn add_word(a: Word, b: Word) -> Result<Word, RuntimeError>`
`fn add_usize(a: usize, b: usize) -> Result<usize, RuntimeError>`
`fn sub_word(a: Word, b: Word) -> Result<Word, RuntimeError>`
`fn sub_usize(a: usize, b: usize) -> Result<usize, RuntimeError>`

These are meant to be used with operations that should not fail per protocol (e.g. gas deduction on post-execution)

Considering the function is implemented as inlined, the compiler will trivially optimize `Ok(x)` into `x`, if `optimized` feature is enabled

vlopes11:
This module aims to centralize all arithmetic operations so they can be easily updated whenever required.

It is desirable to run all protocol checked operations as unchecked code, but then it will be feasible only after we have comprehensive test coverage.

Closes #214
Relates to #170

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,10 @@ quickcheck_macros = "1.0"
 serde_json = "1.0"
 
 [features]
+default = ["optimized"]
 arbitrary = ["fuel-asm/arbitrary"]
 debug = []
+optimized = []
 profile-gas = ["profile-any"]
 profile-coverage = ["profile-any"]
 profile-any = ["dyn-clone"] # All profiling features should depend on this
--- a/src/arith.rs
+++ b/src/arith.rs
@@ -1,0 +1,75 @@
+//! Arithmetic functions for the Fuel VM
+
+use fuel_asm::{PanicReason, Word};
+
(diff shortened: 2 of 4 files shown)


mitch-fuel (approved):
looks great.

PR #218, opened 2022-09-19
Add `InterpreterStorage::deploy_contract`

vlopes11:
Consumers of the VM might want to replicate the contract deployment also to emulate the behavior of storages.

Currently, the only way to do that is via a `Create` transaction, but then we might want to facilitate this operation without creating the whole transaction requirements in order to allow complex operations such as benchmarking.

This commit introduces the ability to add a contract into an interpreter storage using the same strategy that transaction execution would use. It is added as part of the storage because the responsibility for the implementation is a concrete aspect of the database backend.

### DIFF
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -100,18 +100,8 @@ where
                 }
 
                 storage
-                    .storage_contract_insert(&id, &contract)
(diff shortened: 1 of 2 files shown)


xgreenx (approved):

mitch-fuel (approved):
lgtm

PR #229, opened 2022-09-30
Fix bug in `RETD` which throws `MemoryOverflow` for buffers at the very top of RAM.

### LINKED ISSUE #228: `RETD` has off-by-one for overflow check.
If `RETD` tries to return a buffer wedged right at the top of RAM (i.e., where the heap grows down from) then it will erroneously throw a `MemoryOverflow`.

otrho:
Closes #228.

### DIFF
--- a/src/interpreter/flow.rs
+++ b/src/interpreter/flow.rs
@@ -86,7 +86,7 @@ impl<S> Interpreter<S> {
     }
 
     pub(crate) fn ret_data(&mut self, a: Word, b: Word) -> Result<Bytes32, RuntimeError> {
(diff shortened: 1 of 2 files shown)


src/interpreter/flow.rs
```diff
@@ -86,7 +86,7 @@ impl<S> Interpreter<S> {
     }
 
     pub(crate) fn ret_data(&mut self, a: Word, b: Word) -> Result<Bytes32, RuntimeError> {
-        if b > MEM_MAX_ACCESS_SIZE || a >= VM_MAX_RAM - b {
```
mohammadfawaz, line 89: Wow!

vlopes11 (approved):

PR #232, opened 2022-10-12
Read Set/Unset Status along with SRW and SRWQ

mitch-fuel:
Part of FuelLabs/fuel-specs#395

Signed-off-by: Mitch Martin <mitch.martin@fuel.sh>


Cargo.toml
```diff
@@ -109,3 +109,6 @@ required-features = ["random"]
 name = "test-predicate"
 path = "tests/predicate.rs"
 required-features = ["random"]
+
+[patch.crates-io]
+fuel-asm = { git = "https://github.com/FuelLabs/fuel-asm/" }
```
Dentosal, line 114: The patch section shouldn't be committed. We should release a new fuel-asm version with the required changes instead.

leviathanbeak: I think it makes sense while the review is in progress, so that everything builds and passes, it can be easily updated once review is done

Voxelot (approved):

Dentosal (approved):

PR #233, opened 2022-10-14
Pass the contract id value to Panic

Salka1988:
In the function `append_panic_receipt` I added the contract_id of the contract that is missing in the inputs (`PanicReason::ContractNotInInputs`)

```rust
          let vm_result = vm
                .transact(checked_tx)
                .map_err(|error| Error::VmExecution {
                    error,
                    transaction_id: tx_id,
                })?
                .into_owned();
```
Testing the last version of `fuel-vm` on the last version of `fuel-core`, I got an error because `.into_owned()` no longer worked.
That's why I implemented:
```rust
 impl<'a> Into<StateTransition> for StateTransitionRef<'a>
```


Voxelot:
@Salka1988 I've released a new version of fuel-tx with the changes referenced by the patch on this PR. Please update the fuel-tx version and remove the patch so that we can release this change next.

Salka1988:
> @Salka1988 I've released a new version of fuel-tx with the changes referenced by the patch on this PR. Please update the fuel-tx version and remove the patch so that we can release this change next.

Updated!

Voxelot (approved):

PR #234, opened 2022-10-15
Add gas initial costs

vlopes11:
The gas cost calculation is derived from empirical results of the fuel-core benchmarks.

The evaluated behavior is an exponential cost increase for memory operations that overlaps OS pages.

The common value for different OS, and the linux default, is set to `4096`, so we used this reference value here as well. It is important we have a global value, regardless of the operating-system, so we don't end having a platform dependent gas calculation - that is undesirable for distributed systems whose shouldn't violate deterministic calculations depending on the host.

Some tests are no longer applicable because they either tested out of scope aspects of their module, or tried to enforce linearity in the gas costs - which now, by definition, is no longer applicable.

The atomic costs set in `src/gas.rs` are a simple statistical extrapolation from the empirical data fetched from fuel-core benchmarks. These values are expected to be tweaked over time, when we have multiple use-cases.


Dentosal (approved):
The new values look reasonable to me

PR #235, opened 2022-10-18
Refactoring of getters/setters/offset and adapt `Interpreter` to work with generic transactions.

xgreenx:
It is an adaption of the refactoring from the [`fuel-tx`](https://github.com/FuelLabs/fuel-tx/pull/197).
The main idea is that `Interpreter` works with a generic transaction that implements the `ExecutableTransaction` trait instead of the enum `CheckedTransaction`. The trait is defined inside of the `fuel-vm` to minimize dependency from `fuel-tx`.

Also, methods `check_predicates`, `transact`, and `deploy` accept only `Checked<Tx>` type, but after `Interpreter` unwraps them and works with `Tx` itself. The `Interpreter` can call `mut` methods, so I moved all `mut` methods from `CheckedTransaction` into `fuel-vm`.

`ExecutableTransaction` is implemented only by `Transaction::Script` and `Transaction::Create` transactions.

`Interpreter` has three main methods that should be supported and that affect the `Tx` type:
- `check_predicates` can be run for `Create` and `Script` transactions. It is a static method of the `Interpreter` because inside, we create a new instance of `Interpreter` for each predicate and run it.
- `deploy` can be run only with the `Create` transaction. Because we don't need to run a script, we don't need to use the inner state of the `Interpreter`. This method can be run for any instance of `Interpreter` and only insert the `Contract` into DB.
- `transact` with current usage it can be run for `Create` and `Script` transactions. But we actually only run a script for the `Script` transaction and for `Create`, we only call the `deploy` method. I left it generic for now. In the follow-up PRs we can use `transact` only with `Script` transactions.

Most of the changes in the PR are applying bounds for the generic, renaming some methods, changing of type from `CheckTransation` to `Checked<Tx>` or `Tx`, and moving functionality from `fuel-tx` into `fuel-vm`.

Also, I removed several unused methods(I checked the code of `fuel-vm`, `fuel-core`, `sway`) from `fuel-tx` and `fuel-vm` affected by refactoring not to support them.

There is no new logic, only copy/paste from one place to another and adaptation to use generic. All tests are passed, so the chance of regression is low, but I hope reviewers will able to catch them, if any=)


src/interpreter.rs
```diff
@@ -125,26 +134,209 @@ impl<S> Interpreter<S> {
...
+    fn as_script(&self) -> Option<&Script>;
+
+    /// Casts the `Self` transaction into `&mut Script` if any.
+    fn as_script_mut(&mut self) -> Option<&mut Script>;
+
+    /// Casts the `Self` transaction into `&Create` if any.
+    fn as_create(&self) -> Option<&Create>;
+
+    /// Dumps the `Output` by the `idx` into the `buf` buffer.
+    fn output_to_mem(&mut self, idx: usize, buf: &mut [u8]) -> io::Result<usize> {
```
xgreenx, line 173: All methods below moved from `fuel-tx` https://github.com/FuelLabs/fuel-tx/pull/197#discussion_r995026127

src/interpreter/executors/main.rs
```diff
@@ -27,171 +45,40 @@ impl Interpreter<PredicateStorage> {
...
                 result && vm.clone()._check_predicate(predicate)
             })
     }
-
-    /// Initialize the VM with the provided transaction and check the input predicate indexed by
-    /// `idx`. If the input isn't of type [`Input::CoinPredicate`], the function will return
-    /// `false`.
-    ///
-    /// For additional information, check [`Self::check_predicates`]
-    pub fn check_predicate(&mut self, tx: CheckedTransaction, idx: usize) -> bool {
```
xgreenx, line 58: It is not used in the `fuel-vm` or `fuel-core`, so I removed it

src/interpreter/executors/main.rs
```diff
@@ -226,6 +113,153 @@ where
...
+        {
+            return Err(InterpreterError::Panic(PanicReason::ContractNotInInputs));
+        }
+
+        let storage = &mut self.storage;
+        storage
+            .deploy_contract_with_id(salt, storage_slots, &contract, &root, &id)
+            .map_err(InterpreterError::from_io)?;
+
+        // TODO: Maybe the code below is useless because we can't spend any gas and can't execute
```
xgreenx, line 143: I'm not sure do we really need this for the `Create` transaction or not. I tried to have the same code that was before the refactoring, but if you agree that we don't need it, I will remove it in this PR=)

src/interpreter/executors/main.rs
```diff
@@ -226,6 +113,153 @@ where
...
+    }
+}
+
+impl<S, Tx> Interpreter<S, Tx>
+where
+    S: InterpreterStorage,
+    Tx: ExecutableTransaction,
+{
+    pub(crate) fn run(&mut self) -> Result<ProgramState, InterpreterError> {
+        // TODO: Remove `Create` from here
```
xgreenx, line 164: Before merging the PR, I will create according issue with the description

The main idea is that the `transact` method in the future should work only with the `Script` transaction. For `Create` there exists a `deploy` method that should be used.

src/interpreter/initialization.rs
```diff
@@ -1,20 +1,28 @@
...
+{
     /// Initialize the VM with a given transaction
-    fn _init(&mut self, tx: CheckedTransaction) -> Result<(), InterpreterError> {
-        self.tx = tx;
+    fn _init(&mut self, tx: Tx, initial_balances: InitialBalances) -> Result<(), InterpreterError> {
+        // TODO: Remove cloning of the transaction and convert it into the `Transaction`.
+        //  VM can work directly with inner types. It is only added to make the
+        //  code as it was before the refactoring.
+        self.tx = tx.clone();
+        let mut transaction: Transaction = tx.into();
```
xgreenx, line 23: I tried to have the same business logic as before the refactoring.
But it seems we don't need to serialize exactly the `Transaction` into the buffer of the VM. We can serialize `Script` or `Create` without the first 8 bytes that indicate discriminant of the `Transaction` enum.

I will create an issue and follow-up PR if you are okay with this TODO.
Better to do it in a separate PR because it also affects tests and may affect others who try to work with the `memory` field.

src/interpreter/post_execution.rs
```diff
@@ -18,19 +20,23 @@ where
...
     /// checked transactions - hence, validated - this case should be unreachable.
-    pub(crate) fn finalize_outputs(&mut self, revert: bool) -> Result<(), RuntimeError> {
-        let outputs = self.tx.transaction().outputs().len();
-        let params = &self.params;
-        let tx = &mut self.tx;
-
-        let remaining_gas = self.registers[REG_GGAS];
-
-        tx.update_outputs(params, revert, remaining_gas, &self.balances)
+    pub(crate) fn finalize_outputs<Tx>(
```
xgreenx, line 23: If we don't need this method for `Create`(TODO: link on the comment above), then I will revert this change, and it will be used only for `Script` =)

src/interpreter/executors/main.rs
```diff
@@ -27,171 +45,40 @@ impl Interpreter<PredicateStorage> {
     ///
     /// This is not a valid entrypoint for debug calls. It will only return a `bool`, and not the
     /// VM state required to trace the execution steps.
-    pub fn check_predicates(tx: CheckedTransaction, params: ConsensusParameters) -> bool {
+    pub fn check_predicates<S, Tx>(checked: Checked<Tx, S>, params: ConsensusParameters) -> bool
+    where
+        S: Stage,
```
Voxelot, line 50: The VM doesn't really care about the stage, since any level of checked must be guaranteed to be usable by the VM. Is there a way we could avoid the viral spread of this generic across the VM which doesn't need to distinguish between txs with checked vs unchecked signatures?

Voxelot: Maybe the VM could have it's own simplified checked type and these methods could accept any `Tx: Into<VmCheckedTx>`?

src/interpreter/executors/main.rs
Voxelot: Eventually we want to be able to parallelize predicate verification which is why the option to do it on a per input basis was created, but we can add this back later.

src/interpreter/executors/main.rs
Voxelot: Probably a good call to just maintain the existing behavior, but in theory the gas limit should only apply to scripts.

xgreenx: I prefer to remove it now if we agree that it is only applicable for `Script`, but I also can create a follow-up PR for it=)

src/interpreter/executors/main.rs
xgreenx: The first question is, do we need `Stage` in the type of `Checked` at all? I introduced `Stage` because I thought that:
- VM works only with fully checked transactions.
- TxPool works with partially checked transactions.
- During block production, we transform all transactions from TxPool into fully checked and pass them into VM.
- In the future, we will have a separate stage for verified predicates.
- In the future, we can have a stage for checked not finalized transactions(during the production of the block, we fill several fields like placeholders for the contract's utxo, outputs, etc).

But after a discussion with you, VM can work with any type of checked transaction. TxPool wants to work with fully checked transactions(for now, it works with both types). We don't want to make the code more complex.

The idea was to have a state machine for transaction checking on rust level type to manage our code in the right direction. And allow only a limited set of methods for each stage.

So, we can avoid state machine and instead use an enum for `Checked` to only highlight different stages.

```rust
pub enum Checked<Tx: IntoChecked> {
    Stateless(Tx, Tx::Metadata),
    Signatures(Tx, Tx::Metadata),
    ..,
}
```

The problem with this approach is that some methods will do nothing, like the `check_signatures` for the `Signature` variant.

Also, we can use the old approach and have only one struct `Checked<Tx>` with booleans inside or bitmask to describe checking stages.

xgreenx: > The VM doesn't really care about the stage, since any level of checked must be guaranteed to be usable by the VM. Is there a way we could avoid the viral spread of this generic across the VM which doesn't need to distinguish between txs with checked vs unchecked signatures?

It affects only input methods like `deploy`, `transaction`, and `check_predicates`. So the spread is not huge=)

> Maybe the VM could have it's own simplified checked type and these methods could accept any Tx: Into<VmCheckedTx>?

We can create a `CheckedForVM` or `VmChecked` type, but it will spread usage of `into()` during the call of methods above=) And it seems worse the definition of 3 methods with generic

Voxelot: If we don't need stage at all then maybe we should just remove it to reduce complexity. It's probably not a big deal if check_signatures does nothing for a `Checked::Signatures` variant, as that could be optimized away by the compiler potentially.

src/interpreter/executors/main.rs
Voxelot: Well in the case a user sets a high gas limit, we should still refund it if no gas is consumed.

src/interpreter/executors/main.rs
xgreenx: From the performance side, of course, there is no problem, it will be optimized, and the call will be removed at all. It is more about code clarity=)

Then let's remove it for now and return back to this idea later

src/interpreter/executors/main.rs
xgreenx: Yeah, but for the `Create` transaction, the `Output` by default can be initially fully refunded because we always spent zero gas=)

src/interpreter/executors/main.rs
xgreenx: Done https://github.com/FuelLabs/fuel-tx/pull/197/commits/c9b86b8a6a0ce1bfa123afffcd6b07a3d866519f

src/interpreter/executors/main.rs
Voxelot: yeah makes sense, as long as the gas / refund behavior is the same we can refactor this code

xgreenx: Removed in this pr then: https://github.com/FuelLabs/fuel-vm/pull/235/commits/b94c69b8519977f879b8bddac909a88221a8c179

Cargo.toml
```diff
@@ -109,3 +109,6 @@ required-features = ["random"]
 name = "test-predicate"
 path = "tests/predicate.rs"
 required-features = ["random"]
+
+[patch.crates-io]
+fuel-tx = { git = "https://github.com/FuelLabs/fuel-tx/" }
```
Voxelot, line 114: we can remove this now right?

xgreenx: We need to create a new release with `fuel-tx`. Do we want to create a new release for refactoring, and after a new release for `Mint`, or one release for both?

Voxelot: I just approved the mint tx, so we can release both at once.

Voxelot (approved):

Cargo.toml
```diff
@@ -108,4 +108,4 @@ required-features = ["random"]
 [[test]]
 name = "test-predicate"
 path = "tests/predicate.rs"
-required-features = ["random"]
+required-features = ["random"]
```
bvrooman, line 111:
```suggestion
required-features = ["random"]

```

PR #240, opened 2022-10-24
Avoid rounding issues in MLOG instruction

### LINKED ISSUE #150: Logarithm instruction `mlog` needs discussion
I would expect that `assert(1000.logarithm(10) == 3);` would pass, but it does not.
Instead, logging `1000.logarithm(10)` returns `2`,

The testing I've done uses this implementation in sway:
```
impl Logarithm for u64 {
    fn logarithm(self, base: Self) -> Self {
        asm(r1: self, r2: base, r3) {
            mlog r3 r1 r2;
            r3: Self
        }
    }
}
```
Note that other cases are working as expected, i.e:
```
assert(4.logarithm(2) == 2);
assert(16.logarithm(2) == 4);
assert(9.logarithm(3) == 2);
assert(100.logarithm(10) == 2);
```

Dentosal:
Closes #150.

See https://github.com/FuelLabs/fuel-vm/issues/150#issuecomment-1288797787 for detailed explanation and benchmark.

This is a minor breaking change: it fixes a bug, but it's unlikely that anything is depending on the broken behavior.

This contains a performance regression, but correctness is more important here. ~~In any case, this has to be benchmarked and gas cost adjusted.~~ Initial benchmarking shows no significant impact. Either the benchmark needs to be changed to add the edge cases, or the pricing is good enough already.

### DIFF
--- a/src/interpreter/executors/instruction.rs
+++ b/src/interpreter/executors/instruction.rs
@@ -132,13 +132,7 @@ where
 
             OpcodeRepr::MLOG => {
                 self.gas_charge(GAS_MLOG)?;
-                self.alu_error(
-                    ra,
-                    |b, c| (b as f64).log(c as f64).trunc() as Word,
-                    b,
-                    c,
-                    b == 0 || c <= 1,
-                )?;
+                self.alu_error(ra, |b, c| unchecked_ilog(b, c) as Word, b, c, b == 0 || c <= 1)?;
             }
 
             OpcodeRepr::MOD => {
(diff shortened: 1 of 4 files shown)


src/interpreter/executors/instruction.rs
```diff
@@ -496,5 +490,39 @@ where
     }
 }
 
+/// Computes logarithm for given exponent and base.
+/// Diverges when exp == 0 or base <= 1.
+#[inline(always)] // Force copy of each invocation for optimization, see comments below
+const fn _unchecked_ilog_inner(exp: Word, base: Word) -> u32 {
```
Voxelot, line 496: Maybe include a reference to where this code was sourced from?

src/interpreter/executors/instruction.rs
```diff
@@ -496,5 +490,39 @@ where
...
+
+/// Logarithm for given exponent and an arbitrary base, rounded
+/// rounded down to nearest integer value.
+///
+/// Must not be called if the exponent == 0, or if the base <= 1,
+/// as that causes either division by zero, or an infinite loop.
+///
+/// TODO: when https://github.com/rust-lang/rust/issues/70887 is stabilized,
+/// consider using that instead.
+const fn unchecked_ilog(exp: Word, base: Word) -> u32 {
```
Voxelot, line 514: Why do we want this to be unchecked? Is there an efficient way to make this checked?

Voxelot: see: https://github.com/FuelLabs/fuel-vm/issues/170

Dentosal: This is unchecked due to the check already being done on callsite, where we use `alu_error` to check for the invariants. However, the compiler should be able to eliminate the `.expect(...)`, so I've changed this to `checked_` instead.

src/interpreter/executors/instruction.rs
Dentosal: Done

Voxelot: Not seeing any changes, was a push missed somewhere?

Dentosal: Yep, fixed now

Voxelot (approved):

Issue #236, opened 2022-10-19
`cant_write_to_reserved_registers` test fails after introduction of the operation cost


xgreenx:
During CI after the commit got a random error in the `cant_write_to_reserved_registers` test for the following opt codes: 
- `MEQ(15, 7, 20, 7)`
- `MEQ(2, 44, 41, 7)`
- `MEQ(10, 22, 23, 7)`

https://github.com/FuelLabs/fuel-vm/actions/runs/3277949519/jobs/5395801037

Maybe we need increase gas limit and don't worry, but better to investigate.

Voxelot:
This is likely due to the recent gas schedule changes in https://github.com/FuelLabs/fuel-vm/pull/234

Dentosal:
I'm also getting this error locally, quickcheck gives it approximately on 10% of runs. Quite unpleasant flakiness.

```
---- interpreter::executors::instruction::tests::reserved_registers::cant_write_to_reserved_registers stdout ----
thread 'interpreter::executors::instruction::tests::reserved_registers::cant_write_to_reserved_registers' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: (691587335)
Error: expected ReservedRegisterNotWritable error (MEQ(14, 12, 36, 7), Err(PanicInstruction(InstructionResult { reason: OutOfGas, instruction: Instruction { op: 41, ra: 14, rb: 12, rc: 36, rd: 7, imm06: 7, imm12: 2311, imm18: 51463, imm24: 3721479 } })))'
```

Voxelot:
This test will likely need to be updated to use a more structured approach to fuzzing the opcodes, since as @mitch-fuel pointed out in our sync we now have the ability to use a very large amount of gas with just a single instruction.

In the meantime we can either:
- Substantially increase the max_gas_per_tx ConsensusParameter (u64::MAX) in this test to hopefully minimize the issue
- Ignore the test until it has a smarter way of generating random opcodes.
- Update the proptest to just discard failures due to out of gas errors.

PR #241, opened 2022-10-24
fix test flakiness

Voxelot:
After the recent addition of a gas price schedule, one of our proptests would randomly run into out of gas errors.

fixes: #236


xgreenx (approved):
I'm okay with ignoring `OutOfGas` cases. But maybe it is still better to minimize values in the registers or maybe increase the gas limit to not miss some cases=)

For example, an infinity loop in some methods like `ilog` can be matched with out of gas=) error)

Voxelot:
The issue with setting a large gas limit is that the value will end up in a register, which could end up randomly being selected as the range of memory in `MEQ` and cause the limit to be exceeded. Ideally we could have unit tests for opcodes which could verify the criteria like "does not write to reserved register" instead of trying to cover all the cases here.

PR #246, opened 2022-10-27
State Range Ops on InterpreterStorage trait

Voxelot:
add breaking interface changes now so we can patch in range functionality later without breaking version changes. 

This is in support of https://github.com/FuelLabs/fuel-specs/pull/414

### DIFF
--- a/src/storage/interpreter.rs
+++ b/src/storage/interpreter.rs
@@ -138,7 +138,8 @@ pub trait InterpreterStorage:
     }
 
     /// Fetch a range of values from a key-value mapping in a contract storage.
-    /// Returns None if any of the slots in the range are unset.
+    /// Returns the full range requested using optional values in case
+    /// that a particular storage slot is unset.  
     fn merkle_contract_state_range(
         &self,
         id: &ContractId,
--- a/src/storage/interpreter.rs
+++ b/src/storage/interpreter.rs
@@ -8,7 +8,7 @@ use crate::storage::{ContractsAssets, ContractsInfo, ContractsRawCode, Contracts
 use std::borrow::Cow;
 use std::error::Error as StdError;
 use std::io;
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 /// When this trait is implemented, the underlying interpreter is guaranteed to
 /// have full functionality
@@ -144,10 +144,8 @@ pub trait InterpreterStorage:
         &self,
         id: &ContractId,
         start_key: &Bytes32,
-        #[allow(unused)] range: Word,
-    ) -> Result<Vec<Option<Cow<Bytes32>>>, Self::DataError> {
-        self.storage::<ContractsState>().get(&(id, start_key)).map(|v| vec![v])
-    }
+        range: Word,
+    ) -> Result<&[Option<Cow<Bytes32>>], Self::DataError>;
 
     /// Insert a range of key-value mappings into contract storage.
     /// Returns None if any of the keys in the range were previously unset.
@@ -155,13 +153,8 @@ pub trait InterpreterStorage:
         &mut self,
         contract: &ContractId,
         start_key: &Bytes32,
-        #[allow(unused)] range: Word,
-        value: &[Bytes32],
-    ) -> Result<Option<()>, Self::DataError> {
-        self.storage::<ContractsState>()
-            .insert(&(contract, start_key), &value[0])
-            .map(|v| v.map(|_| ()))
-    }
+        values: &[Bytes32],
+    ) -> Result<Option<()>, Self::DataError>;
 
     /// Remove a range of key-values from contract storage.
     /// Returns None if any of the keys in the range were already unset.
@@ -169,12 +162,8 @@ pub trait InterpreterStorage:
         &mut self,
         contract: &ContractId,
         start_key: &Bytes32,
-        #[allow(unused)] range: Word,
-    ) -> Result<Option<()>, Self::DataError> {
-        self.storage::<ContractsState>()
-            .remove(&(contract, start_key))
-            .map(|v| v.map(|_| ()))
-    }
+        range: Word,
+    ) -> Result<Option<()>, Self::DataError>;
 
     /// Fetch the balance of an asset ID in a contract storage.
     fn merkle_contract_asset_id_balance(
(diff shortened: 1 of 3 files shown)


src/storage/interpreter.rs
```diff
@@ -137,6 +137,34 @@ pub trait InterpreterStorage:
...
     }
 
+    /// Fetch a range of values from a key-value mapping in a contract storage.
+    /// Returns the full range requested using optional values in case
+    /// a requested slot is unset.  
+    fn merkle_contract_state_range(
+        &self,
+        id: &ContractId,
+        start_key: &Bytes32,
+        range: Word,
```
xgreenx, line 147: I can't get what is `range` here and why it is `Word`=)

The storage slots are not required to follow each other. They can be randomly initialized, and in this case, how should it work?

Is it like `count` and after `start_key` we want to fetch the next `count` available slots? If yes, then how can they be empty?

I see how we can return the merkle tree by some merkle root(In this case, we can iterate throw the whole tree, and the code will work with sub-trees). But seems we want something another=)

src/storage/interpreter.rs
```diff
@@ -137,6 +137,34 @@ pub trait InterpreterStorage:
         self.storage::<ContractsState>().remove(&(contract, key))
     }
 
+    /// Fetch a range of values from a key-value mapping in a contract storage.
+    /// Returns the full range requested using optional values in case
+    /// a requested slot is unset.  
+    fn merkle_contract_state_range(
```
xgreenx, line 143: Hmm, do we want to duplicate each method with "ranged" version? Maybe better to add `range` to `merkle_contract_state` and don't use it right now=)

src/storage/interpreter.rs
```diff
@@ -137,6 +137,34 @@ pub trait InterpreterStorage:
...
+        &self,
+        id: &ContractId,
+        start_key: &Bytes32,
+        range: Word,
+    ) -> Result<&[Option<Cow<Bytes32>>], Self::DataError>;
```
xgreenx, line 148: I think it will be hard to return `&[_]`, because, in most cases, we will create a temporary object in the implementation, and we will face the problem with ownership.

src/storage/interpreter.rs
Voxelot: It's similar but not quite exactly the same. It's basically a "fetch all keys between [start_key..start_key + range]" assuming you convert the keys to a u256 or something similar. If there are unset slots within that range, we need to zero out the corresponding memory array in the VM.

src/storage/interpreter.rs
Voxelot: Not all storage ops are going to be ranged, for example SRW only fetches a single word instead of an entire slot. So it doesn't take a range.

src/storage/interpreter.rs
Voxelot: changed to vec

src/storage/interpreter.rs
xgreenx: Does `merkle_contract_state_range` with `range = 0` returns the same result as `merkle_contract_state`?

src/storage/interpreter.rs
xgreenx: Hmm, then the usage of `u256` in the signature definitely would make it cleaner=D Or we can pass something like `end: Byte32` or `end: Option<Byte32>'.

What is our limitation for `range`?

From the linked issues, I get that we have two states of the storage slots: set/unset. If the value is zeroed or not zeroed - it is set. If the value is `None` - unset. What is the return data for this method in the next case?

Storage slots:
[Set(1)], Set(2), ..., Set(x - 1), Unset(x), Unset(x + 1), ..., Unset(n)]

What is output `merkle_contract_state_range(start = 1, range = n)`? Is the length of the output vector `n`? Or is it `x - 1`?

Voxelot: From the [specs](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#srwq-state-read-sequential-32-byte-slots), it's prescribed as `MEM[$rA, 32 * rD] = STATE[MEM[$rC, 32 * rD]];`.

So this would mean the length of the returned vector should be "range" or `n`, including the start_key.

Since we need to zero out 32bytes slots in memory if that slot is empty, it would be as you say:
> [Set(1)], Set(2), ..., Set(x - 1), Unset(x), Unset(x + 1), ..., Unset(n-1)]

src/storage/interpreter.rs
Voxelot: hmm according to the [specs](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#srwq-state-read-sequential-32-byte-slots), it would return and do nothing.

xgreenx: Oh, sorry, I meant with `range = 1`=)

src/storage/interpreter.rs
Voxelot: > Hmm, then the usage of u256 in the signature definitely would make it cleaner=D Or we can pass something like end: Byte32 or `end: Option'.

We currently don't have a u256 primitive usable in our fuel-types, and without those it would be harder to interact with an "end" style api from the VM perspective.

src/storage/interpreter.rs
Voxelot: Yeah it would be the same result as before in that case

src/storage/interpreter.rs
xgreenx: Hmm, okay. Maybe instead of returning `None` for each `Unset` element, we can not return them at all?

If the user requests `start...start+range` and get only some subset, it means he got only `Set` slots, all others are `Unset`. In this case, we can support big `range` values and user can ask for all available slots in the whole storage by `start = 0, range = u256::Max`.

src/storage/interpreter.rs
xgreenx: So, do we want to have only one method and use it with `range = 1`, or do we prefer two separate methods?

src/storage/interpreter.rs
Voxelot: The thing is we need to know the exact position of each slot that was unset, and clear the corresponding bytes in `MEM[$rA + 32 * i, 32]`

src/storage/interpreter.rs
Voxelot: I figured its more efficient to have separate methods, so that we don't have do deal with heap types more than we have to

src/storage/interpreter.rs
Voxelot: Since we will charge them the same fee regardless of whether the data was set or not, large collections of state like that likely wouldn't fix into a normal tx gas limit either.

xgreenx: > The thing is we need to know the exact position of each slot that was unset, and clear the corresponding bytes in MEM[$rA + 32 * i, 32] from the VM memory

I see, because we don't return the length of the return data`MEM[$rA, 32 * rD] = STATE[MEM[$rC, 32 * rD]]` and we expect the array of the same size, we need to zero `Unset` slots.

Suppose the command would support dynamic length return(return length <= `rD`) - the array's length with only `Set` slots. In that case, we could return only `Set` values, and we don't need to charge the user for the processing of `Unset` because we didn't spend any computation resources(In RocksDB all data are sorted by key, `iter_all` will process only available slots). 

But it is another kind of command that we can implement later.

src/storage/interpreter.rs
xgreenx: If `Storage` returns only `Set` slots, we can simplify `merkle_contract_state_range` return type into `Result<Vec<Cow<Bytes32>>, Self::DataError>;`, and the processing for `range = 1` will be the same as in `merkle_contract_state`.

But yea, we still will work with heap type - `Vec` so let's keep both methods=)

xgreenx (approved):

src/storage/interpreter.rs
```diff
@@ -137,6 +137,34 @@ pub trait InterpreterStorage:
...
+        values: &[Bytes32],
+    ) -> Result<Option<()>, Self::DataError>;
+
+    /// Remove a range of key-values from contract storage.
+    /// Returns None if any of the keys in the range were already unset.
+    fn merkle_contract_state_remove_range(
+        &mut self,
+        contract: &ContractId,
+        start_key: &Bytes32,
+        range: Word,
```
xgreenx, line 165: Maybe then `Word` is better to replace with a smaller type like `u16`?

src/storage/interpreter.rs
Voxelot: yeah rocksdb iterators were my first impression when I originally saw the spec proposal, but that is a different behavior.

PR #250, opened 2022-10-28
Bugfix: Update outputs of `Create` transaction after execution

xgreenx:
Update outputs of `Create` transaction after execution and added unit test to cover this case.

### DIFF
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -118,10 +118,15 @@ impl<S, Tx> Interpreter<S, Tx>
 where
     S: InterpreterStorage,
 {
-    fn _deploy(&mut self, mut create: Create, initial_balances: InitialBalances) -> Result<Create, InterpreterError> {
+    fn _deploy(
+        create: &mut Create,
+        storage: &mut S,
+        initial_balances: InitialBalances,
+        params: &ConsensusParameters,
+    ) -> Result<(), InterpreterError> {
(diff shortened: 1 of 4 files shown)


src/interpreter/executors/main.rs
```diff
@@ -135,23 +140,20 @@ where
...
+            create,
             false,
             remaining_gas,
             &initial_balances,
             &RuntimeBalances::from(initial_balances.clone()),
-            &self.params,
+            params,
         )?;
-        Ok(create)
+        Ok(())
```
Voxelot, line 156: [nit] consider consolidating in-memory updating of outputs here

```suggestion
        self.update_transaction_outputs()?;
        Ok(())
```

src/interpreter/executors/main.rs
```diff
@@ -251,9 +260,7 @@ where
                 &self.balances,
                 &self.params,
             )?;
-
-            let outputs = self.transaction().outputs().len();
-            (0..outputs).try_for_each(|o| self.update_memory_output(o))?;
+            self.update_transaction_outputs()?;
```
Voxelot, line 263: [nit] move this to the `finalize_outputs()` method, to reduce the risk of this not being called.

xgreenx: We don't need to call `self.update_transaction_outputs()?;` for `deploy` method

Voxelot (approved):

PR #255, opened 2022-11-01
Use correct register to get `contract_id` during `Call`

xgreenx:


Voxelot (approved):

Voxelot:
cc @Salka1988 @digorithm 

we noticed that rB was being used incorrectly instead of rA, which is actually suppose to point at the callframe

PR #261, opened 2022-11-03
Added a temporary field used by `Receipt::Panic` to fill `contract_id` if any

xgreenx:
Fix for https://github.com/FuelLabs/fuel-vm/pull/255 and https://github.com/FuelLabs/fuel-vm/pull/233

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -409,7 +409,13 @@ pub mod test_helpers {
     ) {
         let receipts = client.transact(checked_tx);
 
-        if let Receipt::Panic { id: _, reason, .. } = receipts.get(0).expect("No receipt") {
+        if let Receipt::Panic {
+            id: _,
+            reason,
+            contract_id,
+            ..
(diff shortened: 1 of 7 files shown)


src/util.rs
```diff
@@ -409,14 +409,24 @@ pub mod test_helpers {
...
+        {
             assert_eq!(
                 &expected_reason,
                 reason.reason(),
                 "Expected {}, found {}",
                 expected_reason,
                 reason.reason()
             );
+            match expected_reason {
+                PanicReason::ContractNotInInputs => assert!(contract_id.is_some()),
```
Voxelot, line 427: Maybe we should also have tests that the contract id is set to an non-zero value?

Salka1988 (commented):
LGTM!

src/util.rs
xgreenx: Checked `contract_id` for `LCD` opt code

Voxelot (approved):

PR #262, opened 2022-11-03
Fix MROO rounding issues

### LINKED ISSUE #260: MROO has float rounding issues
Similarly to the MLOG issues discussed in https://github.com/FuelLabs/fuel-vm/issues/150, MROO also has floating point rounding issues. This is the only location of the vm still using floats.

For instance, calling MROO with values `mroo(647214625, 3)`, i.e. cube root of 647214625, gives 864.9999999999997 in floating point, while 865**3 == 647214625.

I'll investigate the performance impact of handling this with integer-only operations.

Dentosal:
Closes #260. See the issue comments for implementation  details as well.


src/interpreter/executors/instruction.rs
```diff
@@ -499,6 +499,57 @@ where
...
+
+    // Use floating point operation to get an approximation for the starting point.
+    // This is at most off by one in either direction.
+    let guess = (target as f64).powf((nth_root as f64).recip()) as u64;
+
+    debug_assert!(guess != 0, "This should never occur for {{target, n}} > 1");
+
+    // Compute guess**n to check if the guess is too large.
+    // Note that if guess == 1, then g1 == 1 as well, meaning that we will not return here.
+    let Some(g1) = guess.checked_pow(nth_root) else {
```
Voxelot, line 532: I don't think we should jump to rust 1.65 features immediately, as there are several downstream deps that may not be ready for this change which will prevent a release.

Voxelot (approved):
This seems reasonable - and the behavior is definitely an improvement over the previous algorithm. We may want to get this more thoroughly reviewed before mainnet though.

PR #265, opened 2022-11-06
Code Clarity

ControlCplusControlV:
Since this repo doesn't have clippy enabled there was a buildup of some clippy errors which should be fixed. I fixed the ones revolving clarity around the code

### DIFF
--- a/src/interpreter/balances.rs
+++ b/src/interpreter/balances.rs
@@ -71,7 +71,7 @@ impl RuntimeBalances {
 
                 state
                     .entry(asset)
-                    .or_insert(Balance::new(0, offset))
+                    .or_insert_with(|| Balance::new(0, offset))
                     .checked_add(balance)
                     .ok_or(CheckError::ArithmeticOverflow)?;
 
@@ -108,7 +108,7 @@ impl RuntimeBalances {
             .get_mut(asset)
             .and_then(|b| b.checked_add(value))
             .map(|balance| Self::_set_memory_balance(balance, memory))
-            .or((value == 0).then(|| 0))
+            .or_else(|| (value == 0).then_some(0))
     }
 
     /// Attempt to subtract the balance of an asset, updating the VM memory in the appropriate
@@ -118,7 +118,7 @@ impl RuntimeBalances {
             .get_mut(asset)
             .and_then(|b| b.checked_sub(value))
             .map(|balance| Self::_set_memory_balance(balance, memory))
-            .or((value == 0).then(|| 0))
+            .or_else(|| (value == 0).then_some(0))
     }
 
     /// Write all assets into the VM memory.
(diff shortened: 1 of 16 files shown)


src/storage/memory.rs
```diff
@@ -110,7 +110,8 @@ impl StorageInspect<ContractsRawCode> for MemoryStorage {
 
 impl StorageMutate<ContractsRawCode> for MemoryStorage {
     fn insert(&mut self, key: &ContractId, value: &[u8]) -> Result<Option<Contract>, Infallible> {
-        Ok(self.memory.contracts.insert(*key, value.clone().into()))
+        // Or Ok(self.memory.contracts.insert(*key, &(*value).clone().into())) if reference should be cloned
```
ControlCplusControlV, line 113: I did have a question here, clippy was warning this is cloning a clone and not the underlying value, what would be desirable behavior here?

src/interpreter/balances.rs
```diff
@@ -136,8 +136,8 @@ impl RuntimeBalances {
             let value = balance.value();
             let ofs = balance.offset();
 
-            (&mut vm.memory[ofs..ofs + AssetId::LEN]).copy_from_slice(asset.as_ref());
-            (&mut vm.memory[ofs + AssetId::LEN..ofs + AssetId::LEN + WORD_SIZE]).copy_from_slice(&value.to_be_bytes());
+            (vm.memory[ofs..ofs + AssetId::LEN]).copy_from_slice(asset.as_ref());
+            (vm.memory[ofs + AssetId::LEN..ofs + AssetId::LEN + WORD_SIZE]).copy_from_slice(&value.to_be_bytes());
```
xgreenx, line 140:
```suggestion
            vm.memory[ofs..ofs + AssetId::LEN].copy_from_slice(asset.as_ref());
            vm.memory[ofs + AssetId::LEN..ofs + AssetId::LEN + WORD_SIZE].copy_from_slice(&value.to_be_bytes());
```

src/interpreter/internal.rs
```diff
@@ -68,7 +68,7 @@ where
 
             // Transaction memory space length is already checked on initialization so its
             // guaranteed to fit
-            (&mut self.memory[offset..offset + Bytes32::LEN]).copy_from_slice(&root[..]);
+            (self.memory[offset..offset + Bytes32::LEN]).copy_from_slice(&root[..]);
```
xgreenx, line 71:
```suggestion
            self.memory[offset..offset + Bytes32::LEN].copy_from_slice(&root[..]);
```

src/interpreter/memory.rs
```diff
@@ -231,9 +232,9 @@ where
 
         self.has_ownership_range(&range)
             .then(|| {
-                (&mut self.memory[addr..]).iter_mut().take(len).for_each(|m| *m = 0);
+                (self.memory[addr..]).iter_mut().take(len).for_each(|m| *m = 0);
```
xgreenx, line 235:
```suggestion
                self.memory[addr..].iter_mut().take(len).for_each(|m| *m = 0);
```

src/storage/memory.rs
xgreenx: `Ok(self.memory.contracts.insert(*key, value.into()))` is right

xgreenx (approved):

Dentosal (dismissed):
LGTM. Good steps towards making clippy usable with this repo as well.

src/storage/memory.rs
ControlCplusControlV: So I haven't gotten around to fixing this one (will tomorrow), but I was wondering why it passes when I changed it clone the underlying value and not the clone of a clone? Shouldn't this break something or do our tests just not cover it?

ControlCplusControlV:
Sorry, removed comment to be in line with what greenx had stated so no confusing /misleading comments made it into the code

xgreenx (approved):

PR #271, opened 2022-11-28
Fix bug in load byte

freesig:
fixes https://github.com/FuelLabs/sway/issues/3436
Load byte was using the `RegisterId` instead of the actual registers value as the memory offset.


otrho (approved):
Ooh, nice find.

Voxelot (approved):

PR #272, opened 2022-11-28
Avoid cloning and deserializing contract code

freesig:
This PR allows the contract code to be copied directly to the vm memory which also avoid needless serialization. This is a substantial saving for the `Call` op. This PR will need to be merged and released together with an accompanying PR on fuel-core so I am marking this as a draft even though it's ready for review.

PR #279, opened 2022-12-05
Allow vm to take gas costs as input

freesig:
- Adds the `GasCosts` type.
- Sets the default gas costs.
- Adds `bench-helpers` feature to allow resetting the receipts buffer for benchmarks.
- Adds diffing ability that allows testing any op.


fuel-vm/src/interpreter/diff.rs
```diff
@@ -0,0 +1,418 @@
...
+    Tx: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.registers == other.registers
+            && self.memory == other.memory
+            && self.frames == other.frames
+            && self.receipts == other.receipts
+            && self.tx == other.tx
+            && self.initial_balances == other.initial_balances
+            // && self.storage == other.storage
```
ControlCplusControlV, line 375: Why are these comparisons commented out?

fuel-vm/src/gas.rs
```diff
@@ -124,3 +136,361 @@ impl GasUnit {
...
+    pub sw: Word,
+    pub sww: Word,
+    pub swwq: Word,
+    pub time: Word,
+    pub tr: Word,
+    pub tro: Word,
+    pub xor: Word,
+    pub xori: Word,
+
+    // Dependant
```
ControlCplusControlV, line 245: I know changing this would require a lot of changes, so just a comment, but we may want to consider the conventions of American vs Australian vs British English for comments and naming, would be good to add to the rfcs

fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -0,0 +1,122 @@
+use super::*;
+
+pub const GIT: &str = "1a660f59b892e7816da01c288c86696c9d7d7a36";
```
ControlCplusControlV, line 3: A comment would be nice to document exactly what this is referencing

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -75,62 +74,62 @@ where
 
         match op {
             OpcodeRepr::ADD => {
-                self.gas_charge(GAS_ADD)?;
+                self.gas_charge(self.gas_costs.add)?;
```
ControlCplusControlV, line 77: Not familiar, but what are the performance costs of changing this from a constant which is inlined into an instance variable? Gas cost counting I know is significant on performance otherwise I wouldn't bring it up

fuel-vm/src/interpreter/flow.rs
```diff
@@ -161,6 +161,13 @@ where
         let asset_id =
             AssetId::try_from(&self.memory[c as usize..cx as usize]).expect("Unreachable! Checked memory range");
 
+        let mut frame = self.call_frame(call, asset_id)?;
+
+        // FIXME: This is checking the cost after the db call has happened.
```
ControlCplusControlV, line 166: Would be nice to reference this line in the issue just for better clarity

fuel-vm/src/gas.rs
```diff
@@ -124,3 +136,361 @@ impl GasUnit {
         Self::Accumulated(self.cost() + other.cost())
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Gas costings for every op.
+/// The inner values are wrapped in an [`Arc`]
```
ControlCplusControlV, line 142: What are the performance implications of checking gas costs though through an `Arc`, since while it will be cheap to clone I'd imagine this is quite a bit more expensive than referencing constant values for Gas costs

Voxelot: The Arc is probably insignificant, but we could add a todo here to optimize this using lifetimes or a regular `Rc` since there's no concurrency in the VM.

fuel-vm/src/gas.rs
Voxelot: It seems like @adlerjohn has preferred using American English at least in the specs.

fuel-vm/src/gas.rs
ControlCplusControlV: Can we make a perf issue to then move these to `lazy_static` once things stabilize then? I understand why it is like this for now, but as we push more towards the efficiency frontier it may be helpful to remember this

fuel-vm/src/gas.rs
```diff
@@ -124,3 +136,361 @@ impl GasUnit {
...
+    pub sll: Word,
+    pub slli: Word,
+    pub srl: Word,
+    pub srli: Word,
+    pub srw: Word,
+    pub sub: Word,
+    pub subi: Word,
+    pub sw: Word,
+    pub sww: Word,
+    pub swwq: Word,
```
Voxelot, line 238: now that we have range ops, wouldn't all quad-word state ops have a dependent cost like `srwq`?

fuel-vm/src/gas.rs
Voxelot: lazy_static alone may be insufficient, would require some investigation to find the exact solution.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -403,7 +408,7 @@ where
         let output = Output::from_bytes(&self.memory[offset..])?;
 
         // amount isn't checked because we are allowed to send zero balances with a message
-        if !matches!(output, Output::Message { recipient, .. } if recipient == Address::zeroed()) {
+        if !matches!(output, Output::Message { recipient: r, .. } if r == Address::zeroed()) {
```
Voxelot, line 411: [nit] did this need to be changed?

fuel-vm/src/gas.rs
ControlCplusControlV: #302

freesig: `Deref` of an `Arc` is just de-refencing a pointer so it won't be any quicker too do a static or `Rc`

fuel-vm/src/gas.rs
freesig: I can just do a grep and changes this. I didn't realize it was spelt differently in the US.

fuel-vm/src/gas/default_gas_costs.rs
freesig: This is generated code so I would need to add this in the fuel-core PR

fuel-vm/src/interpreter/executors/instruction.rs
freesig: We have no choice if we want to be able to pass in the gas costs at runtime

fuel-vm/src/interpreter/flow.rs
freesig: https://github.com/FuelLabs/fuel-vm/pull/272#issuecomment-1376614156

fuel-vm/src/gas.rs
freesig: Just FYI there's no way to try and get const performance when we are passing in values at runtime.

Voxelot: > Just FYI there's no way to try and get const performance when we are passing in values at runtime.

*without unsafe

fuel-vm/src/gas/default_gas_costs.rs
Voxelot: Maybe the codegen could inject some comments?

E.g. 

```
format!("//Generated by fuel-core: {}:{}", file!(), line!())
```

fuel-vm/src/interpreter/diff/storage.rs
```diff
@@ -0,0 +1,474 @@
...
+use super::ExecutableTransaction;
+use super::Interpreter;
+use super::*;
+
+#[derive(Debug)]
+/// The set of state changes that are recorded.
+pub(super) enum StorageDelta {
+    State(MappableDelta<(ContractId, Bytes32), Bytes32>),
+    Assets(MappableDelta<(ContractId, AssetId), u64>),
+    Info(MappableDelta<ContractId, (fuel_types::Salt, Bytes32)>),
```
Voxelot, line 19: [nit] might be more readable as `ContractInfo`

```suggestion
    ContractInfo(MappableDelta<ContractId, (fuel_types::Salt, Bytes32)>),
```

ControlCplusControlV: If this change is made I would also `ContractState` and `ContractState` for consistency

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -172,207 +171,206 @@ where
...
 
             OpcodeRepr::MCP => {
-                self.gas_charge_monad(GAS_MCP, c)?;
+                self.dependant_gas_charge(self.gas_costs.mcp, c)?;
                 self.memcopy(a, b, c)?;
             }
 
             OpcodeRepr::MCPI => {
-                self.gas_charge_monad(GAS_MCPI, imm)?;
+                self.gas_charge(self.gas_costs.mcpi)?;
```
Voxelot, line 334: why are other opcodes like MCLI dependent, but MCPI isnt?

fuel-vm/src/interpreter/flow.rs
```diff
@@ -161,6 +161,13 @@ where
...
+        // when https://github.com/FuelLabs/fuel-vm/pull/272 lands this check
+        // should happen on the pinned slice before reading it.
+        self.dependant_gas_charge(self.gas_costs.call, frame.code().len() as u64)?;
```
Voxelot, line 169:
```suggestion
        self.dependent_gas_charge(self.gas_costs.call, frame.code().len() as u64)?;
```

fuel-vm/src/interpreter/gas.rs
```diff
@@ -2,174 +2,22 @@ use super::Interpreter;
...
-            MCL | MCLI => Arithmetic(1).join(MemoryOwnership(1)),
-
-            MCP | MCPI => Arithmetic(2).join(MemoryOwnership(1)),
-
-            MEQ => Arithmetic(3).join(Branching(2)).join(RegisterWrite(1)),
-
-            CCP => Branching(2).join(StorageReadTree(1)),
-
-            _ => panic!("Opcode is not variable gas"),
+    pub(crate) fn dependant_gas_charge(&mut self, gas_cost: DependentCost, arg: Word) -> Result<(), RuntimeError> {
```
Voxelot, line 15: [nit]

```suggestion
    pub(crate) fn dependent_gas_charge(&mut self, gas_cost: DependentCost, arg: Word) -> Result<(), RuntimeError> {
```

fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -0,0 +1,122 @@
...
+        mroo: 2,
+        mul: 1,
+        muli: 1,
+        noop: 1,
+        not: 1,
+        or: 1,
+        ori: 1,
+        ret: 1,
+        rvrt: 1,
+        s256: 5,
```
Voxelot, line 55: since sha256 takes a slice of memory with arbitrary length, wouldn't it also have a dependent cost?

fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -0,0 +1,122 @@
...
+            base: 148,
+            dep_per_unit: 16,
+        },
+        ldc: DependentCost {
+            base: 156,
+            dep_per_unit: 14,
+        },
+        logd: DependentCost {
+            base: 47,
+            dep_per_unit: 21,
```
Voxelot, line 91: This seems fairly expensive to me 🤔

fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -0,0 +1,122 @@
...
+        },
+        mcl: DependentCost {
+            base: 1,
+            dep_per_unit: 2539,
```
Voxelot, line 95: while only ~3% difference from mcli, seems strange that these aren't closer in cost

fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -0,0 +1,122 @@
...
+            base: 1,
+            dep_per_unit: 1377,
+        },
+        meq: DependentCost {
+            base: 1,
+            dep_per_unit: 2802,
+        },
+        retd: DependentCost {
+            base: 64,
+            dep_per_unit: 21,
```
Voxelot, line 111: at least this is consistent with logd, even if it seems expensive

Voxelot (approved):
The default benchmark results have some unexpected variances (e.g. state opcodes having very different dependent cost behavior). But I think this is a solid start that we can continue to refine and improve on.

fuel-vm/src/gas.rs
freesig: I think you are confusing static and const. Const has to be known at compile time as it's in-lined. You can make a static global for the gas costs but it will still require a pointer de-reference so it will not be any faster then an Arc.

fuel-vm/src/interpreter/diff/storage.rs
freesig: Yeh and then they all end up with contract infront

fuel-vm/src/gas/default_gas_costs.rs
freesig: Ok made an issue for this

freesig:
@Voxelot You have raised a lot of concerns about costs not seeming correct. Would you like me to dig into them before merging this PR or merge as is and create an issue to look into them later?

fuel-vm/src/gas.rs
Voxelot: ah your right, i was thinking of static mut

Voxelot:
@freesig let's merge this as is and get the benchmarks landed in fuel core. We can readjust once more before the release.

Voxelot (approved):

PR #274, opened 2022-11-29
Summon the MonoRepo

ControlCplusControlV:

PR #280, opened 2022-12-05
Include the `fuel-crypto` repo in the great "Summon the MonoRepo" merge

mitchmindtree:
This PR is based on @ControlCplusControlV's WIP merge in #274.

This PR merges the `fuel-crypto` repository into this repo (in a history-preserving manner) to avoid introducing a dependency cycle between the `fuel-vm` and `fuel-crypto` repositories as highlighted [here](https://github.com/FuelLabs/fuel-vm/pull/274#issuecomment-1336388924).

**NOTE:** If/when we merge this into @ControlCplusControlV's branch, we should *not* use the "Squash" option and instead should add a new merge commit. Otherwise, I think we risk squashing the preserved history commits included within this PR into a single commit, which is probably not what we want.

cc @adlerjohn @sezna I think one of you might need to re-enable the option to merge without squashing temporarily until both this PR and #274 have landed?

### DIFF
--- a/fuel-crypto/README.md
+++ b/fuel-crypto/README.md
@@ -1,6 +1,6 @@
 # Fuel Crypto
 
-[![build](https://github.com/FuelLabs/fuel-crypto/actions/workflows/ci.yml/badge.svg)](https://github.com/FuelLabs/fuel-crypto/actions/workflows/ci.yml)
+[![build](https://github.com/FuelLabs/fuel-vm/actions/workflows/ci.yml/badge.svg)](https://github.com/FuelLabs/fuel-vm/actions/workflows/ci.yml)
 [![crates.io](https://img.shields.io/crates/v/fuel-crypto?label=latest)](https://crates.io/crates/fuel-crypto)
 [![docs](https://docs.rs/fuel-crypto/badge.svg)](https://docs.rs/fuel-crypto/)
 [![discord](https://img.shields.io/badge/chat%20on-discord-orange?&logo=discord&logoColor=ffffff&color=7389D8&labelColor=6A7EC2)](https://discord.gg/xfpK4Pe)
--- a/fuel-crypto/README.md
+++ b/fuel-crypto/README.md
@@ -1,0 +1,14 @@
+# Fuel Crypto
+
+[![build](https://github.com/FuelLabs/fuel-crypto/actions/workflows/ci.yml/badge.svg)](https://github.com/FuelLabs/fuel-crypto/actions/workflows/ci.yml)
+[![crates.io](https://img.shields.io/crates/v/fuel-crypto?label=latest)](https://crates.io/crates/fuel-crypto)
+[![docs](https://docs.rs/fuel-crypto/badge.svg)](https://docs.rs/fuel-crypto/)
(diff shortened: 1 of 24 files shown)


adlerjohn:
Ping me when ready to merge and I can disable requirements on squash merging temporarily.

ControlCplusControlV:
Do we need `fuel-crypto` in the PR? I had briefly looked through it and didn't seem like there would be much there in it necessitating it being added. Are you thinking just for fixing dep issues?

ControlCplusControlV:
@mitchmindtree did you merge using a strategy similar to [sub-tree](http://choly.ca/post/git-merge-to-monorepo/)? I used it for the rest of the modules as James Prestwich recommended and the Optimism team also suggested it and used it for their monorepo

mitchmindtree:
> Are you thinking just for fixing dep issues?

Exactly, this is solely because without merging `fuel-crypto` into this repo, we would have a dependency cycle between the repos that would make publishing new versions while keeping CI green in `fuel-vm` a bit frustrating.

Specifically, the repo cycle would come from:

- `fuel-crypto` depends on `fuel-types`.
- `fuel-tx` and `fuel-vm` depend on `fuel-crypto`.

This means when we want to publish a new version of each of the crates, we would have to first publish a new version of `fuel-types` at this repo at which point our `[patch]` would stop working and CI would go red, *then* publish a new version of `fuel-crypto` at the fuel-crypto repo, *then* publish a new version of `fuel-tx` and `fuel-vm` under this repo and get CI green again.

The reason we'd need the `fuel-types` `[patch]` in our workspace here is to patch `fuel-crypto`'s `fuel-types` dependency, which would otherwise be a different version than the local one used by `fuel-tx` and `fuel-vm`.

Having `fuel-crypto` in this repo doesn't really affect how downtream uses it (they still use it via crates.io), but it makes our maintenance and publishing process quite a bit simpler.

All that said, we should hold off on merging this one until we get a nod from @Voxelot!

> @mitchmindtree did you merge using a strategy similar to [sub-tree](http://choly.ca/post/git-merge-to-monorepo/)?

I used the approach described here: https://stackoverflow.com/a/10548919/1711917

It seems to have included all history (including `git blame`) for all files other than those that I had to resolve conflicts in (`ci.yml`, `verify_tags.sh`, `Cargo.toml`). Is there anything else I should look out for in particular?

Voxelot:
I hadn't considered the circular dep issue from fuel-crypto, good investigating! 

Originally we were preferring to keep fuel-crypto separate, but it should generally be harmless to include it here even if there's a bunch of minor releases since it will likely be backwards compatible with older wallets/tooling anyways.

Voxelot (approved):

fuel-crypto/README.md
```diff
@@ -0,0 +1,14 @@
+# Fuel Crypto
+
+[![build](https://github.com/FuelLabs/fuel-crypto/actions/workflows/ci.yml/badge.svg)](https://github.com/FuelLabs/fuel-crypto/actions/workflows/ci.yml)
```
Voxelot, line 3: these URLs are outdated

mitchmindtree: Good spot! I've updated these and squashed the work into the previous "Fix-up" commit, to try and keep the post-merge fixup all under one commit (seeing as we don't want to squash this PR).

PR #283, opened 2022-12-07
`fuel-asm` - refactor for improved type-safety around instructions, opcodes

mitchmindtree:
*This PR is based on the monorepo branch, #274.*

---

This PR includes a pretty significant refactor of the `fuel-asm` crate and updates the `fuel-tx` and `fuel-vm` crates accordingly. The goal is to improve type-safety around the `Instruction` and `Opcode` types, and to derive all items related to the `Instruction` type from a single source of truth: the `impl_instructions!` table.

## Name Changes

The following renames have been performed as a part of this PR to more closely align with [the spec](https://fuellabs.github.io/fuel-specs/master/vm/index.html#semantics):

- `Opcode` -> `Instruction`
- `OpcodeRepr` -> `Opcode`.

The old `Instruction` type has been removed, as it tended to encourage some anti-patterns. Namely, converting to the old `Instruction` type meant we *always* unpacked every possible register ID and immediate value, even if they weren't relevant to the instruction variant specified by the opcode. Further, it allowed access to register IDs and immediates that weren't valid at all to the instruction specified by the opcode.

Instead, the new `Instruction` type requires matching the specific instruction type specified by the opcode before allowing to unpack the valid register IDs and immediate. This also helps to avoid some of the unnecessary unpacking that would occur on every instruction within the `Interpreter::_instruction` step.

Another notable difference is that the new `Instruction` type is *packed* (represented in 4 bytes) while the previous equivalent `Opcode` type was *unpacked* (closer to ~40 bytes). The idea here is that the `Instruction` type should remain in a packed form until the moment it is interpreted, at which point only the relevant data is unpacked based on its opcode variant.

## Handwriting ASM

The process of handwriting ASM has also changed slightly. Previously, handwriting ASM took advantage of the unpacked `Opcode` variant, e.g. `Opcode::ADDI(ra, rb, imm)`. Seeing as the new equivalent `Instruction` enum is packed, we instead provide an `op` module with a convenience function for constructing each type of instruction, e.g. `op::addi(ra, rb, imm)`.

By using a function for constructing handwritten instructions, we can also validate the register ID and immediate arguments to ensure the values specified are within the 6/12/18/24-bit expected range and produce an error otherwise. As these functions are solely for handwriting ASM, which is generally only useful for writing tests or benchmarking, we *panic* in the case that a given register ID or immediate value is out of range. On the other hand, during bytecode decoding, we do *not* check the value range for register IDs or immediates as we know that the values cannot possibly lie outside of the range of valid values when packed within the 32-bit raw instruction layout.

## `impl_instructions!`

The new `impl_instructions!` macro allows us to specify a table of all instructions as a single source of truth, from which the declarations and implementations for `Opcode` and `Instruction` (and the nested instruction variant types) are derived. The idea here is to reduce the need to update multiple areas of the crate each time an instruction is added-to or removed-from the spec, removing some scope for mistakes.

That said, the `macro_rules!` implementation is of course not as easy to read as plain copy-paste branches and types. In order to make it a little more accessible, I've added some detailed docs to the `macros` module which aims to explain how each of the types and implementations are derived. I've also attempted to separate and comment each of the different sets of pattern matching branches within the `macro_rules!` definition to help guide future contributions.

## No Reserved or Undefined Opcodes

The old RESERVED* and Undefined variants have been ommitted from the `Instruction` and `Opcode` types. Instead, we consider encountering these during bytecode decoding as error cases. By making it impossible to construct an `Instruction` or `Opcode` from a reserved or undefined value, we can guarantee that `Instruction` or `Opcode` instances within our code must be valid for a lot more cases.

## Typed `RegId`, `Imm*`

When packing or unpacking instructions from/to their respective register IDs and immediates, we now use unit structs rather than type aliased primitive types. This allows us to guarantee correct masking by construction, and avoids the need to double-check potentially incorrect masks when inspecting the unpacked instruction data.

As a convenience, `Index<RegId>` is implemented for `[T]` where `[T]: Index<usize>`. This makes indexing with the typed `RegId`s into the interpreter's `registers` array more ergonomic and avoids the need to cast to `usize` everywhere.

These types are currently only used for instruction packing/unpacking, though in future PRs we might benefit from further type safety by progressively replacing our use of the old `RegisterId` and `Immediate*` type aliases in favour of the new types.

## Instruction Debug impl

A custom Debug implementation is provided for the new `Instruction` type to make it easier to grok an instruction's contents, e.g.

```rust
println!("{:?}", op::addi(0x10, 0x11, 10));
println!("{:?}", op::sub(0x10, 0x11, 0x12));
```

prints as

```console
ADDI { ra: 16, rb: 17, imm: 10 }
SUB { ra: 16, rb: 17, rc: 18 }
```

which is a bit easier to read than the default implementation which would only show the packed data and look more like this:

```console
ADDI(ADDI([56, 142, 241]))
SUB(SUB([56, 212, 128]))
```

## Removal of `io::{Read/Write}` Implementations

These implementations seemed a little out of place in `fuel-asm`, and noticing that we don't appear to use these downstream, I decided to omit these. Instead, we can always read/write `u32`s or `u8`s and use the provided `Instruction` conversions from there. 

## Breaking !!

As there's quite a bit of breakage throughout these changes, I plan to leave this in draft at least until I've got a follow-up PR ready at the `fuel-core` repo as I imagine there are probably loads of tests, etc there too that will also need updating. I plan to help shepherd these changes up through the SDK and Sway repos too.

## Follow-up

There's still a lot more that can be done to improve type-safety throughout `fuel-vm`. This PR aims to limit the scope of changes to those related to the `Instruction` and `Opcode` types as there's already a lot here. I'll aim to open some follow-up issues I've come across during this refactor and link them back here tomorrow.


fuel-asm/src/lib.rs
```diff
@@ -1,26 +1,578 @@
...
+pub use gm_args::{GMArgs, GTFArgs};
+pub use instruction_result::InstructionResult;
+pub use panic_reason::PanicReason;
 
-#[cfg(feature = "serde")]
-#[doc(no_inline)]
-pub use serde;
+/// Represents a 6-bit register ID, guaranteed to be masked by construction.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct RegId(u8);
```
mitchmindtree, line 26: These `RegId` and `Imm*` types are probably better off in the `fuel-types` crate, potentially replacing the old `RegisterId` and `Immediate*` types as a type-safe alternative? Doing so would require a lot more breakage throughout fuel-vm and downstream, so I thought I'd leave that for follow-up work as this PR's already quite big.

fuel-vm/src/util.rs
```diff
@@ -19,7 +20,7 @@
 /// let call = Call::new(contract_id, 0, 0).to_bytes();
 /// let asset_id = [0x00; 32];
 /// let transfer_amount: Word = 100;
-/// let gas_to_forward = 1_000_000;
+/// let gas_to_forward = 100_000;
```
mitchmindtree, line 23: The original value specified here was out-of-range of the possible 18-bit values for the immediate (max 262,144), caught by the new instruction constructor.

Dentosal:
I'm working on getting this merged now

fuel-asm/src/macros.rs
```diff
@@ -1,42 +1,678 @@
...
+    // Generate a shorthand free function named after the $op for constructing an `Instruction`.
+    (impl_op_constructor $doc:literal $Op:ident $op:ident [RegId]) => {
+        #[doc = $doc]
+        pub fn $op(ra: u8) -> Instruction {
+            $Op::new(check_reg_id(ra)).into()
+        }
+    };
+    (impl_op_constructor $doc:literal $Op:ident $op:ident [RegId RegId]) => {
+        #[doc = $doc]
+        pub fn $op(ra: u8, rb: u8) -> Instruction {
```
Dentosal, line 393: I think we should be taking the typed `RegId` here instead of `u8`

mitchmindtree: The idea behind this constructor was to be a more "ergonomic" alternative, solely for handwriting ASM (i.e. writing tests - see [this section of the doc comment](https://github.com/FuelLabs/fuel-vm/blob/cf52dc7eedd2e7f55ab5467e92e33bc27396cd00/fuel-asm/src/macros.rs#L168-L190)). You can still use the `$Op::new` constructor directly which checks `RegId` at compile-time rather than runtime (e.g. as the compiler will do), but it is a little verbose for tests.

I'm open to making this change if you still think it's worth taking `RegId` here, but it will mean updating loads of tests :sweat_smile:

mitchmindtree:
Thanks for the review and merging master @Dentosal!

mitchmindtree:
Switching to "Ready for review", but keep in mind this is a breaking change and will likely require updating loads of tests in `fuel-core` too, so it could be worth preparing a PR for `fuel-core` first? I can take a stab at this soon, but if someone wants to beat me to it feel free!

Dentosal:
@mitchmindtree I'm working on the fuel-core integration, I should have a draft PR up tomorrow.

fuel-asm/src/macros.rs
Dentosal: Right. The idea of updating this came to me from doing lots of test changes, which require changing lines like `Opcode::RET(REG_ONE)` to `op::ret(REG_ONE.into())` or `op::ret(1)`. The latter seems quite misleading; changing it to `op::ret(2)` would return the value of `$of` instead of `2`.  The nice thing with typed registers is that we could instead have `op::ret(RegId::ONE)` instead.

mitchmindtree: @Dentosal ahh I can see your point. I have no major preference either way, will leave the call up to you!

Dentosal: After wasting a couple of hours on trying to figure how to update the tests without spending days on rewriting them, I've given up on this. The current way of things is good enough.

Dentosal: I made the shorthands take `CheckRegId` trait instead of `u8` for `RegId` and impl'd that for both `u8` and `RegId`, so that at least we don't have to do `REG_ZERO.into()` at every callsite, but `RegId` can be passed directly.

Dentosal:
`fuel-core` follow up PR: https://github.com/FuelLabs/fuel-core/pull/973

fuel-vm/src/interpreter/flow.rs
```diff
@@ -238,34 +238,34 @@ where
...
+        rb: RegId,
+        rc: RegId,
+        rd: RegId,
     ) -> Result<(), RuntimeError> {
         const M: &str = "the provided id is not a valid register";
 
         let a = self
             .registers
-            .get(ra)
+            .get(ra.to_u8() as usize)
```
mitchmindtree, line 244: These could be
```suggestion
            .get(ra.into())
```

mitchmindtree (commented):
All lgtm :+1:

fuel-vm/src/interpreter/flow.rs
mitchmindtree: Ahh just realised this causes type inference issues - nevermind

mitchmindtree:
Just added a tiny commit that addresses some docs typos I made.

fuel-asm/src/macros.rs
```diff
@@ -1,42 +1,683 @@
...
-                unsafe { core::mem::transmute::<u8, Self>(b) }
+/// This macro is intentionaly private. See the module-level documentation for a thorough
+/// explanation of how this macro works.
+macro_rules! impl_instructions {
+    // Recursively declares a unique struct for each opcode.
+    (decl_op_struct $doc:literal $ix:literal $Op:ident $op:ident [$($field:ident)*] $($rest:tt)*) => {
+        #[doc = $doc]
+        #[derive(Clone, Copy, Eq, Hash, PartialEq)]
+        #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+        pub struct $Op(pub (super) [u8; 3]);
```
freesig, line 200: One of the biggest pain points with working with ops is that instead of them actually indicating which args they take it's all register so you have to constantly refer back to the docs.

I would love to go from `op::call($rA, $rB, $rC, $rD,)` to `op::call(contract_id, //etc)` .
With `OpCode` we had the chance to name the fields by turning the variants into struct variants. It seems this removes that possibility. Is there some way would could add that do this macro? It doesn't have to be within this PR but I'd like to see it be possible before approving this.

freesig (changes requested):

freesig (commented):
Overall this looks really good, I like how clean it is. My only concern would be the above. Nice work!!

fuel-asm/src/macros.rs
mitchmindtree: I think we can always wrap the constructors with functions that take whatever args in whatever more human-readable layout we like - I think [these `GM` and `GTF` alternate constructors](https://github.com/FuelLabs/fuel-vm/blob/c08823e7b1ee84405779ce880bbd0d99554d26e2/fuel-asm/src/op.rs#L11-L33) might be a good example of the kind of direction we can take.

Similarly, we can also add alternatives to the `unpack` functions that "unpack" to some more human-readable type (rather than just a tuple with `RegId`s and `Imm*`s).

That's a sweet idea actually, I like the idea of adding types like that as a part of `fuel-asm` in the future like you mention (rather than doing all the register and immediate interpretation in the fuel-vm for example) as other downstream crates might find them useful / safer too.

freesig: Yeh I really like the idea of unpacking to types. Do you think the following is possible?:
```rust
impl_instructions! {
    "Adds two registers."
    0x10 ADDI addi [Sum: RegId, Lhs: RegId, Rhs: U12]
```
Which generates something like the following:
```rust
pub fn new(sum: RegId<u64>, lhs: RegId<u64>, rhs: Imm12<U12>) -> Self

pub fn sum(&self) -> RegId<u64>
pub fn lhs(&self) -> RegId<u64>
pub fn rhs(&self) -> Imm12<U12>
```
and a debug like:
```
ADDI { sum: RegId(16), lhs: RegId(17), rhs: U12(10) }
```

mitchmindtree: Yeah that looks sick - totally doable!

freesig (dismissed):
Approving but probably want at least two approvals and maybe @Voxelot before merging

Voxelot:
The rename of `fuel-asm/src/opcode/args.rs → fuel-asm/src/gm_args.rs` isn't entirely accurate, as it contains both gm and gtf args

fuel-asm/src/lib.rs
```diff
@@ -1,26 +1,627 @@
...
+    [T]: core::ops::IndexMut<usize, Output = T>,
+{
+    fn index_mut(&mut self, ix: RegId) -> &mut Self::Output {
+        &mut self[usize::from(ix)]
+    }
+}
+
+// Collect instructions into bytes or halfwords
+
+#[cfg(feature = "std")]
```
Voxelot, line 517: since we're using `core::iter` does this need to be `std` only?

fuel-asm/src/op.rs
```diff
@@ -0,0 +1,33 @@
...
+impl GTF {
+    /// Construct a `GTF` instruction from its arguments.
+    pub fn from_args(ra: RegId, rb: RegId, args: GTFArgs) -> Self {
+        Self::new(ra, rb, Imm12::new(args as _))
+    }
+}
+
+/// Construct a `GM` instruction from its arguments.
+pub fn gm_args(ra: u8, args: GMArgs) -> Instruction {
+    Instruction::GM(GM::from_args(RegId::from(ra), args))
```
Voxelot, line 27: should we be validating that the RegId is within a valid range (0..64)?

fuel-asm/tests/encoding.rs
```diff
@@ -11,261 +10,188 @@ fn opcode() {
...
-        Opcode::gtf(r, r, GTFArgs::OutputContractBalanceRoot),
-        Opcode::gtf(r, r, GTFArgs::OutputContractStateRoot),
-        Opcode::gtf(r, r, GTFArgs::OutputMessageRecipient),
-        Opcode::gtf(r, r, GTFArgs::OutputMessageAmount),
-        Opcode::gtf(r, r, GTFArgs::OutputContractCreatedContractId),
-        Opcode::gtf(r, r, GTFArgs::OutputContractCreatedStateRoot),
-        Opcode::gtf(r, r, GTFArgs::WitnessDataLength),
-        Opcode::gtf(r, r, GTFArgs::WitnessData),
-        Opcode::Undefined,
+    let mut instructions = vec![
```
Voxelot, line 13: would there be a way to autogenerate this based on the macro to ensure no cases are missed?

fuel-asm/tests/encoding.rs
```diff
@@ -11,261 +10,188 @@ fn opcode() {
...
-        }
-
-        // Assert no panic with checked function
-        while !op_bytes.is_empty() {
-            op_bytes.pop();
-
-            assert!(Opcode::from_bytes(op_bytes.as_slice()).is_err());
-        }
-
-        #[cfg(feature = "serde")]
```
Voxelot, line 254: Is this removed serde coverage tested elsewhere now?

fuel-vm/src/consts.rs
```diff
@@ -10,56 +11,56 @@ use std::mem;
 pub const VM_REGISTER_COUNT: usize = 64;
 
 /// Contains zero (0), for convenience.
-pub const REG_ZERO: usize = 0x00;
```
Voxelot, line 13: Do we still need these VM constants if they are basically just aliasing the ones from fuel-asm?

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -50,446 +50,539 @@ where
...
-        if self.is_predicate() && !op.is_predicate_allowed() {
+        if self.is_predicate() && !instruction.opcode().is_predicate_allowed() {
             return Err(PanicReason::TransactionValidity.into());
         }
 
-        match op {
-            OpcodeRepr::ADD => {
+        // Short-hand for retrieving the value from the register with the given ID.
+        // We use a macro to "close over" `self.registers` without taking ownership of it.
+        macro_rules! r {
```
Voxelot, line 71: is a macro really necessary for doing this basic fetch? Couldn't we achieve similar results with an inlined helper function that takes `&self` as the first arg?

fuel-vm/src/interpreter/flow.rs
```diff
@@ -20,7 +20,7 @@ where
     Tx: ExecutableTransaction,
 {
     pub(crate) fn jump(&mut self, j: Word) -> Result<(), RuntimeError> {
-        let j = self.registers[REG_IS].saturating_add(j.saturating_mul(Instruction::LEN as Word));
+        let j = self.registers[REG_IS].saturating_add(j.saturating_mul(core::mem::size_of::<RawInstruction>() as Word));
```
Voxelot, line 23: [nit] put `core::mem::size_of::<RawInstruction>() as Word` into a constant

fuel-vm/src/interpreter/internal.rs
```diff
@@ -108,7 +108,7 @@ impl<S, Tx> Interpreter<S, Tx> {
 
     pub(crate) fn inc_pc(&mut self) -> Result<(), RuntimeError> {
         self.registers[REG_PC]
-            .checked_add(Instruction::LEN as Word)
+            .checked_add(core::mem::size_of::<RawInstruction>() as Word)
```
Voxelot, line 111: [nit] put `core::mem::size_of::<RawInstruction>() as Word` into a constant

fuel-vm/src/interpreter/internal.rs
```diff
@@ -229,7 +231,7 @@ mod tests {
         let maturity = 0;
         let height = 0;
 
-        let script = Opcode::RET(0x01).to_bytes().to_vec();
+        let script = <[u8; 4]>::from(op::ret(0x01)).to_vec();
```
Voxelot, line 234: maybe we should make a helper fn for this common pattern?

e.g.
```rust
fn serialize_ops<I: IntoIterator<Item = Instruction>>(ops: I) -> Vec<u8> {
...
}
```

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -41,7 +41,7 @@ where
     }
 
     /// Execute a provided instruction
-    pub fn instruction(&mut self, instruction: Instruction) -> Result<ExecuteState, InterpreterError> {
+    pub fn instruction(&mut self, raw: RawInstruction) -> Result<ExecuteState, InterpreterError> {
```
Voxelot, line 44: Consider changing this to cleanup a lot of the .into() changes elsewhere:

```suggestion
    pub fn instruction<R: Into<RawInstruction>>(&mut self, raw: R) -> Result<ExecuteState, InterpreterError> {
        let raw = raw.into();
```

fuel-vm/src/state/debug.rs
```diff
@@ -25,7 +25,7 @@ impl Breakpoint {
     /// size. Also, the op count is always relative to `$is` so it should
     /// consider only the bytecode of the contract.
     pub const fn new(contract: ContractId, pc: Word) -> Self {
-        let pc = pc * (Opcode::LEN as Word);
+        let pc = pc * (core::mem::size_of::<RawInstruction>() as Word);
```
Voxelot, line 28: [nit] put core::mem::size_of::<RawInstruction>() as Word into a constant

fuel-vm/src/util.rs
```diff
@@ -280,7 +283,7 @@ pub mod test_helpers {
             let salt: Salt = self.rng.gen();
             let program: Witness = contract
                 .into_iter()
-                .flat_map(Opcode::to_bytes)
+                .flat_map(<[u8; 4]>::from)
```
Voxelot, line 286: This seems less ergonomic to me than `to_bytes()`, as it forces me to think about how many bytes each instruction should take everytime I want to serialize some instructions. Maybe we should keep the `to_bytes()` as a trait fn for all instructions?

fuel-vm/tests/cgas.rs
```diff
@@ -138,20 +139,20 @@ fn cgas_uses_min_available_gas() {
 
     let program = vec![
         // load amount of tokens
-        Opcode::ADDI(reg_call_a, REG_FP, CallFrame::a_offset() as Immediate12),
-        Opcode::LW(reg_call_a, reg_call_a, 0),
+        op::addi(reg_call_a, REG_FP, CallFrame::a_offset() as Immediate12),
```
Voxelot, line 142: Doesn't seem like we need these casts anymore?

```suggestion
        op::addi(reg_call_a, REG_FP, CallFrame::a_offset()),
```

Voxelot (approved):
Some minor nits, but this is a really nice simplification overall. This macro approach will make it much easier to maintain our asm.

fuel-vm/tests/cgas.rs
Dentosal: We do still need those. `CallFrame::a_offset` returns and `usize` which cannot be passed to `op::addi` which expects an `Imm12` (really `u16`) for this.

fuel-vm/src/state/debug.rs
Dentosal: Done

fuel-vm/src/interpreter/internal.rs
Dentosal: Personally I like the explicitness of `to_bytes().to_vec()`, which both autocompletes from types and is still composable. Added the `to_bytes` back to help with this.

fuel-vm/src/interpreter/internal.rs
Dentosal: Done

fuel-vm/src/interpreter/executors/instruction.rs
Dentosal: It's repeated so much in the rest of the function that I feel usage of a simple macro here more readable. We would have to replace `r!(a).into()` with `self.reg(a).into()` which isn't any more readable for me.

fuel-vm/src/consts.rs
Dentosal: Not really, but these are used in many places. I'll try to remove them.

Dentosal: Done

fuel-vm/src/interpreter/executors/instruction.rs
Dentosal: Done in cb7bed0

fuel-asm/tests/encoding.rs
Dentosal: @mitchmindtree ^

fuel-asm/src/op.rs
Dentosal: Done in c8b6ad1

fuel-asm/src/lib.rs
Dentosal: We don't currently provide `alloc` (alloc-but-not-std) feature. Should we?

fuel-asm/tests/encoding.rs
Dentosal: Implemented here https://github.com/FuelLabs/fuel-vm/pull/340, a separate review for that would be nice

fuel-asm/src/lib.rs
Voxelot: Might be useful in certain contexts, e.g. wasm integrations with the VM.

fuel-asm/tests/encoding.rs
mitchmindtree: Ahh no it's not! I think I got rid of the rest of this loop as it's not really relevant now we've the added type safety, but must have accidentally overlooked this bincode sanity check. I've pushed a commit that re-adds it.

fuel-asm/src/lib.rs
mitchmindtree: WASM supports std, though only a subset of it (I believe `Vec` is included in that).

I'm skeptical any lower-level usage would want to use `alloc`, e.g. In my experience if you can't use `std`, generally you're in such a constrained environment that `alloc`'s `Vec` doesn't make sense either, and you're probably better off with `arrayvec` or something along those lines.

Maybe we're better off waiting until we get requests for features for external collections implementations (like `alloc`) and add them as needed?

mitchmindtree:
> The rename of fuel-asm/src/opcode/args.rs → fuel-asm/src/gm_args.rs isn't entirely accurate, as it contains both gm and gtf args

Good catch! I've renamed that private mod to `args` for now.

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -26,7 +26,7 @@ where
             .ok_or(InterpreterError::Panic(PanicReason::MemoryOverflow))?;
 
         // Store the expected `$pc` after executing `hi`
-        let pc = self.registers[REG_PC] + core::mem::size_of::<RawInstruction>() as Word;
+        let pc = self.registers[REG_PC] + Instruction::SIZE as Word;
```
mitchmindtree: We might want to be careful here - `RawInstruction` is a type alias for `u32`, and doesn't necessarily *need* to match the `Instruction` size.

It just so happens that I *have* made sure that `Instruction` itself *is* the same size as the `RawInstruction` (as it doesn't require any more space for the same information) with an included test in `fuel-asm/src/lib.rs`, so this is technically safe.

That said, we might be better off naming this const `RAW_SIZE` or having a separate `fuel_asm::RAW_INSTRUCTION_SIZE` or whatever to clarify that we're referring to the `u32` repr? If not, we might want to ensure that `Instruction`'s size always matches `RawInstruction`'s size. We do have a test for this, but it might be worth adding a comment to explain why it's important that the sizes match.

mitchmindtree: I've added a comment to the relevant test explaining the importance of this, that should be enough https://github.com/FuelLabs/fuel-vm/pull/283/commits/dd41273fc05c0effc80e6959089681e4125f6072.

Voxelot (approved):

PR #284, opened 2022-12-09
Skip empty `Script` transaction

xgreenx:
Fixes https://github.com/FuelLabs/fuel-core/issues/794

It is fast but not an optimal solution to unblock users. With the right solution, we need to skip execution totally without allocating any memory and `Interpreter`. But it requires some refactoring to support transactions which not affect VM during execution(like empty `Script` or `Create` transactions).

### DIFF
--- a/src/interpreter/executors/main.rs
+++ b/src/interpreter/executors/main.rs
@@ -172,7 +172,7 @@ where
         let state = if let Some(create) = self.tx.as_create_mut() {
             Self::_deploy(create, &mut self.storage, self.initial_balances.clone(), &self.params)?;
             self.update_transaction_outputs()?;
-            ProgramState::Skipped
+            ProgramState::Return(1)
         } else {
             if self.transaction().inputs().iter().any(|input| {
                 if let Input::Contract { contract_id, .. } = input {
@@ -204,7 +204,10 @@ where
             {
                 self.run_program()
             } else {
-                Ok(ProgramState::Skipped)
+                // Return `1` as successful execution.
+                let return_val = 1;
+                self.ret(return_val)?;
+                Ok(ProgramState::Return(return_val))
             };
 
             let gas_used = self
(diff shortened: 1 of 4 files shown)


Voxelot:
fixes: https://github.com/FuelLabs/fuel-core/issues/794

src/interpreter/executors/main.rs
```diff
@@ -193,7 +193,20 @@ where
...
+            let program = if !self
+                .transaction()
+                .as_script()
+                .expect("It should be `Script` transaction")
+                .script()
+                .is_empty()
+            {
+                self.run_program()
+            } else {
+                Ok(ProgramState::Skipped)
```
Voxelot, line 207: While this is a better variant, I think we should just do `ProgramState::Return(1)` like create txs did before to avoid extra breaking changes downstream.

xgreenx:
Why do we need to filter them?

src/interpreter/executors/main.rs
xgreenx: Yeah, let's introduce `ProgramState::Skipped` together with an optimal solution.

Voxelot (approved):

Voxelot:
> From a client side as well we should probably filter these out from the mempool as well right? (separate pr and issue ofc)

we still want to process the change output for empty script txs. We need to skip the vm's `run_program` since it expects a real script that ends in either ret or rvrt. However, empty scripts is a use-case we want to support for basic transfer txs.

ControlCplusControlV:
> Why do we need to filter them?

Is there any reason the node should process empty tx's? Seems like bloat from a larger perspective, and just adds history without doing anything

xgreenx:
> Is there any reason the node should process empty tx's? Seems like bloat from a larger perspective, and just adds history without doing anything

It can be useful for transfer like in Bitcoin, you can consume inputs and produce outputs

Voxelot:
The utxo updates can still make a lot of meaningful change to the blockchain state without any scripts or contracts.

Voxelot:
It's still needs to be a valid script tx type, but only the script field is empty.

PR #289, opened 2022-12-20
Return `None` for zeroed `ContractId` in `Receipt`

xgreenx:
# Problem

We have optional methods for fetching contract id's from receipts, however the VM returns all zeros instead of None. This caused unexpected failures since downstream we expect these methods to only return Some if there is a real contract id.

# Solution

If contract id is zeroed, convert to an Option::None when fetching receipt fields.


ControlCplusControlV (approved):
LGTM

Voxelot (approved):

PR #299, opened 2023-01-09
Show and fix bug in try_mem_write

freesig:
make try_mem_write and try_zeroize testable to show bug in current impl
Now also include the fix

### DIFF
--- a/fuel-vm/src/interpreter/memory.rs
+++ b/fuel-vm/src/interpreter/memory.rs
@@ -194,6 +194,11 @@ impl<S, Tx> Interpreter<S, Tx>
 where
     Tx: ExecutableTransaction,
 {
+    /// Return the registers used to determine ownership.
+    pub(crate) fn ownership_registers(&self) -> OwnershipRegisters {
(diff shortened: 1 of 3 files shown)


fuel-vm/src/interpreter/memory.rs
```diff
@@ -202,36 +202,13 @@ where
...
-            .then(|| {
-                let src = data.as_ptr();
-                let dst = &mut self.memory[addr] as *mut u8;
-
-                unsafe {
-                    ptr::copy_nonoverlapping(src, dst, data.len());
-                }
-            })
-            .ok_or_else(|| PanicReason::MemoryOwnership.into())
+        let registers = OwnershipRegisters::new(self);
```
xgreenx, line 205: [nit] Hmm, maybe we can add one helper method like `fn registers(&self)` and use it `try_mem_write(addr, data, self.registers(), &mut self.memory)`

xgreenx (approved):

Voxelot (approved):

PR #300, opened 2023-01-09
Fix bug in try_mem_write and try_zeroize

freesig:


fuel-vm/src/interpreter/memory.rs
```diff
@@ -451,7 +451,7 @@ fn try_mem_write(
     let ax = addr.checked_add(data.len()).ok_or(PanicReason::ArithmeticOverflow)?;
```
xgreenx, line 451: I guess you can remove this line, and the same for another function

fuel-vm/src/interpreter/memory.rs
```diff
@@ -451,7 +451,7 @@ fn try_mem_write(
...
 
     let range = (ax <= VM_MAX_RAM as usize)
```
xgreenx, line 453: `(ax <= VM_MAX_RAM as usize)` this check can be inside of the `MemoryRange::new`(the same for other methods)

ControlCplusControlV: This might require other changes to move the check into `MemoryRange::new`, as other operations like memclear also use

```
        let (ab, overflow) = a.overflowing_add(b);

        let range = MemoryRange::new(a, b);
        if overflow || ab > VM_MAX_RAM || b > MEM_MAX_ACCESS_SIZE || !self.has_ownership_range(&range) {
            Err(PanicReason::MemoryOverflow.into())
        } else {
 ```
 and similar blocks, so if some cases require additional checks then those may need to be moved into `MemoryRange::new` adding additional load for operations which don't use it. If we don't move those additional checks though we still end up having the checked / (unchecked in some cases) addition being performed twice, once to see if its OOB, and others to check MEM_MAX_ACCESS_SIZE. Maybe it might be best as a small inline function?

**Review by ControlCplusControlV — APPROVED** [review]
Changes LGTM

File: fuel-vm/src/interpreter/memory.rs (hunk 2 above)
Comment (ControlCplusControlV) [reply fuel-vm/src/interpreter/memory.rs:453]: This might require other changes to move the check into `MemoryRange::new`, as other operations like memclear also use

```
        let (ab, overflow) = a.overflowing_add(b);

        let range = MemoryRange::new(a, b);
        if overflow || ab > VM_MAX_RAM || b > MEM_MAX_ACCESS_SIZE || !self.has_ownership_range(&range) {
            Err(PanicReason::MemoryOverflow.into())
        } else {
 ```
 and similar blocks, so if some cases require additional checks then those may need to be moved into `MemoryRange::new` adding additional load for operations which don't use it. If we don't move those additional checks though we still end up having the checked / (unchecked in some cases) addition being performed twice, once to see if its OOB, and others to check MEM_MAX_ACCESS_SIZE. Maybe it might be best as a small inline function?

freesig: Yeh I agree with @ControlCplusControlV. It seems a bit odd to pull caller specific logic into the memory range.

fuel-vm/src/interpreter/memory.rs
freesig: I don't think it can be due to the comment below

fuel-vm/src/interpreter/memory.rs
Voxelot: I think it's also worth noting that the current approach makes it easier to audit compliance with the specs. If all the validation is buried behind generic types it may make code look cleaner but harder to understand from an audit perspective.

Voxelot (approved):

PR #301, opened 2023-01-09
Updated `InterpreterStorage` to be object safe

xgreenx:
Removed `Sized` from `InterpreterStorage`.
Used GATs for lifetimes in the `Mappable` type to remove `for<'a> StorageInspect<Mappable<'a>>`


fuel-vm/src/storage.rs
```diff
@@ -26,7 +25,7 @@ impl Mappable for ContractsRawCode {
 pub struct ContractsInfo;
 
 impl Mappable for ContractsInfo {
-    type Key = ContractId;
+    type Key<'a> = ContractId;
```
freesig, line 28:
```suggestion
    type Key<'static> = ContractId;
```

freesig (commented):
The code looks fine but I'm not sure why references for keys are needed? Every implementation seems to just copy the keys anyway. Is there anywhere they can actually remain a reference?

xgreenx:
Not every implementation, we use it for `ContractsAssets` and `ContractsState` 
image

freesig:
Yeh but I mean are those keys ever actually used as references? It seems like everywhere they are just copied?
Like this:
https://github.com/FuelLabs/fuel-core/blob/a5514420e53fbd09fa0eb4d8de8f65822c18f660/crates/fuel-core/src/database/balances.rs#L37
calls this:
https://github.com/FuelLabs/fuel-core/blob/a5514420e53fbd09fa0eb4d8de8f65822c18f660/crates/fuel-core/src/state.rs#L34

freesig:
I think the above example will end up with it being copied once it hits the database. If it was passed by move then you would only have a single copy at the start. Unless we plan to figure out a way to create multikeys that don't have to be owned (for rocksdb) in the future then I'd say this isn't worth the complexity. It was very hard to work around in the fuel-vm diffing

xgreenx:
We copying it only because of the multikey. Passing by reference or by value doesn't change this fact. The rocksdb works with references, so in the case of not multikey, it will perform optimally.

> If it was passed by move then you would only have a single copy at the start.

If I pass it by reference, I don't need to do any copy at the start=)

> It was very hard to work around in the fuel-vm diffing

I see. It is true that the current solution in the `fuel-vm/src/interpreter/diff/storage.rs` requires a workaround with `ToOwnedKey` and `ToOwned`. I simplified it, and it looks good now=)

BTW, I tried to replace `&(&, &)` with `&(,)` and it requires usage of `from_unchecked` instead of `as_ref_unchecked` in many(more than 4) places of the `fuel-vm`. It means decreasing the performance in production mode. This PR is not about it=)

freesig (approved):

PR #303, opened 2023-01-10
Add a top-level readme file

Dentosal:
I realized this repo is missing one, and the repo page looks rather weird without it.


ControlCplusControlV (commented):
We should probably have a paragraph explaining that this is a mono-repo, and briefly explaining what each sub-project is

Dentosal:
This is not really a monorepo in the usual sense, it's just a set of crates. We could add desriptions here, but I feel like the folder names are self-explanatory enough.

xgreenx:
I think we still need some kind of general description of the repository structure and the purpose of each folder=)

Voxelot (approved):

PR #307, opened 2023-01-13
Predicate gas metering

Dentosal:
Makes predicates respect gas limit. Currently the transcation gas limit is applied to the predicate as well. Also adds predicates flag to `Checked<Tx>`, and a way to perform that check using `CheckedTxExt` trait in fuel-vm.

The following concerns remain unadressed:

* Should predicates have a separate gas limit and/or gas price factor?
* Should we deduct the gas used by the predicate from the gas available to the transaction?


fuel-tx/src/checked_transaction.rs
```diff
@@ -39,12 +42,18 @@ impl core::fmt::Display for Checks {
...
 /// (via the `Into<(Tx, Tx ::Metadata)>` trait), modify them and check again.
 ///
-/// # Dev note: Avoid serde serialization of this type. Since checked tx would need to be
-/// re-validated on deserialization anyways, it's cleaner to redo the tx check.
+/// # Dev note: Avoid serde serialization of this type.
+///
+/// Since checked tx would need to be re-validated on deserialization anyways,
+/// it's cleaner to redo the tx check.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Checked<Tx: IntoChecked> {
```
xgreenx, line 58: Maybe it makes sense to move `Checked` into `fuel-vm`, because it is where we really start to use it=)  In this case you can pass `Interpreter` that would check predicates=)

Voxelot:
> Should predicates have a separate gas limit and/or gas price factor?

They should share the same gas limit/pricing as the txn.

> Should we deduct the gas used by the predicate from the gas available to the transaction?

Yes

fuel-vm/tests/predicate.rs
```diff
@@ -105,3 +105,93 @@ fn get_verifying_predicate() {
...
+        );
+
+        builder.add_input(input);
+    }
+
+    let tx = builder.finalize_checked_basic(0, &ConsensusParameters::default());
+    Interpreter::<PredicateStorage>::check_predicates(tx, Default::default(), Default::default())
+}
+
+#[test]
```
ControlCplusControlV, line 182: You could simplify a lot of these tests with `rtest`, since you've already built out helper functions to execute predicates, more tests could easily be added making each predicate an `rstest` testcase like so

```rust
use rstest::rstest;

#[rstest]
#[case(vec![Opcode::RET(REG_ONE)], true)] // This just succeeds
#[case(vec![Opcode::JI(0)], false)] // This runs out of gas
fn predicate_gas_metering(#[case] predicate: Vec<Opcode>, #[case] expected: bool) {
    assert_eq!(execute_gas_metered_predicate(predicate), expected);
}
```

fuel-vm/src/predicate.rs
```diff
@@ -123,3 +128,30 @@ fn from_tx_works() {
...
+
+impl<Tx: IntoChecked + ExecutableTransaction + CheckedMetadata> CheckedTxExt for Checked<Tx>
+where
+    Self: Clone,
+    <Tx as IntoChecked>::Metadata: CheckedMetadata,
+{
+    /// Check predicates, if not yet done.
+    fn check_predicates(mut self, params: ConsensusParameters, gas_costs: GasCosts) -> Result<Self, CheckError> {
+        if !self.checks().contains(Checks::Predicates) {
+            if Interpreter::<PredicateStorage>::check_predicates(self.clone(), params, gas_costs) {
```
ControlCplusControlV, line 149: It is kind of confusing that the boolean returned by `check_predicates` indicates a failure if `true`? Or am I misreading this check?

Dentosal: Changed the return type to a `Result`, so that it's more readable

fuel-tx/src/transaction/checkable/error.rs
```diff
@@ -105,6 +106,8 @@ pub enum CheckError {
     /// The user provided amounts for coins or gas prices that caused an arithmetic
     /// overflow.
     ArithmeticOverflow,
+    /// Predicate verification failed
+    PredicateVerification,
```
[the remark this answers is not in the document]
Dentosal: Renamed

fuel-vm/tests/predicate.rs
Dentosal: I really don't see the benefit here; the current code isn't any less readable in my opinion.

fuel-tx/src/checked_transaction.rs
Dentosal: Maybe that should be a separate PR after this one? I'd also like to find a way to force tx checking, maybe by encoding that into the type itself.

Dentosal:
This is now ready for actual reviews. There are still two things I'd like to achieve, but those can be addressed in follow-up PRs:

* Making sure that predicates are verified before executing the script
  * Ideally this should be done on the type level
* Moving `Checked<Tx>` to the fuel-vm crate would simplify it a bit
  * I don't think anything needs to use that type without fuel-vm anyway

fuel-vm/tests/predicate.rs
ControlCplusControlV: I think `rstest` is easier to identify what cases are being tested, and easier to add cases in the future. I know we've been trying to use it more across `fuel-core` too. Main benefit imo though is tests just become more maintainable since adding a test-case (lets say one where a bug is encountered) just becomes 1 line rather than an additional patch test. Technically FWIW each test case can be run in parallel too this way

mitch-fuel: this works fine for single opcode tests but becomes more difficult when constructing larger individual test cases

fuel-tx/src/checked_transaction.rs
Dentosal: We have to move lots of stuff to over, because transcation builder and `checked` mod of each transaction type depend on this as well.

Voxelot:
> Moving Checked<Tx> to the fuel-vm crate would simplify it a bit
> - I don't think anything needs to use that type without fuel-vm anyway

Agreed, since we can't check the gas validity of transactions without some upfront VM execution, there's limited utility to having CheckedTransaction type owned by fuel-tx. The alternative would be to create a new VMCheckedTransaction that is a wrapper around CheckedTransaction with some additional state, but it could also lead to confusion for business logic implementation in places like the txpool for instance, as they might not choose the kind of checked transaction type.

Dentosal:
TODO: this breaks trait method `IntoChecked::into_checked` which is currently documented to perform all checks. Either we adjust documentation or we perform the predicate check there as well.

fuel-tx/src/checked_transaction.rs
Dentosal: PR for this: https://github.com/FuelLabs/fuel-vm/pull/310

Dentosal:
We should merge https://github.com/FuelLabs/fuel-vm/pull/310 to this one and then fix the into_checked mentioned above.

Dentosal:
Discovered more work to be done: we should either move `Checkable` trait in `fuel-tx` to `fuel-vm`, or document that it doesn't do predicate verfication checks. Or maybe it could be eliminated entirely?

Voxelot:
It may make more sense to use a trait extension for `Checkable` in fuel-vm. The `Checkable` is heavily used in fuel-tx tests for doing signature verification.

Voxelot:
Decided Checkable was not a good place to do predicate verification and that just renaming `Checkable` to something more fuel-tx specific like `FormatValidityChecks` was a better route.

@mitch-fuel @Dentosal

fuel-vm/src/checked_transaction.rs
```diff
@@ -0,0 +1,870 @@
...
+impl<Tx: IntoChecked + ExecutableTransaction + fuel_tx::field::GasLimit> Checked<Tx>
+where
+    Self: Clone,
+    <Tx as IntoChecked>::Metadata: crate::interpreter::CheckedMetadata,
+{
+    /// Check predicates, if not yet done.
+    pub fn check_predicates(mut self, params: ConsensusParameters, gas_costs: GasCosts) -> Result<Self, CheckError> {
+        if !self.checks().contains(Checks::Predicates) {
+            let checked = Interpreter::<PredicateStorage>::check_predicates(self.clone(), params, gas_costs)?;
+            self.checks_bitmask.insert(Checks::Signatures);
```
Voxelot, line 122: Should be `Checks::Predicates`

Fixed this in my PR here:
https://github.com/FuelLabs/fuel-vm/pull/313

fuel-vm/src/checked_transaction.rs
```diff
@@ -0,0 +1,972 @@
...
+    }
+}
+
+impl<Tx: IntoChecked> AsRef<Tx> for Checked<Tx> {
+    fn as_ref(&self) -> &Tx {
+        &self.transaction
+    }
+}
+
+// #[cfg(feature = "internals")]
```
ControlCplusControlV, line 141: Uncaught commented out feature flag

ControlCplusControlV, line 141:
```suggestion

```

fuel-vm/src/checked_transaction.rs
```diff
@@ -0,0 +1,972 @@
...
+    pub fn check_signatures(mut self) -> Result<Self, CheckError> {
+        if !self.checks_bitmask.contains(Checks::Signatures) {
+            self.transaction.check_signatures()?;
+            self.checks_bitmask.insert(Checks::Signatures);
+        }
+        Ok(self)
+    }
+}
+
+// #[cfg(feature = "internals")]
```
ControlCplusControlV, line 113:
```suggestion
```

ControlCplusControlV (commented):
Should these commented out feature flags be deleted?

fuel-vm/src/checked_transaction.rs
Voxelot: These methods were commented out due to migrating the types from fuel-tx, but they shouldn't be made available as part of the normal public api. I put them under `test-helpers` since we don't want the ability for other crates like fuel-core to have mutable access to checked transactions which could allow for manually flipping validation flags without doing the real validation etc.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -43,33 +52,53 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+                return Err(PredicateVerificationFailed);
+            }
+
+            remaining_gas = vm.registers[REG_GGAS];
+        }
+
+        Ok(PredicatesChecked {
+            gas_used: tx_gas_limit
+                .checked_sub(remaining_gas)
+                .expect("Bug! Execution increased available gas"),
```
Voxelot, line 100: We should use the `Bug` error type here instead of expect: https://github.com/FuelLabs/fuel-vm/blob/master/fuel-vm/src/error.rs#L292

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -43,33 +52,53 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+
+        for predicate in predicates {
+            // VM is cloned because the state should be reset for every predicate verification
+            let mut vm = vm.clone();
+
+            vm.context = Context::Predicate { program: predicate };
+            vm.registers[REG_GGAS] = remaining_gas;
+            vm.registers[REG_CGAS] = remaining_gas;
+
+            if !matches!(vm.verify_predicate(), Ok(ProgramState::Return(0x01))) {
```
Voxelot, line 90: Should we capture OutOfGas error here?

fuel-vm/src/interpreter/initialization.rs
```diff
@@ -38,8 +43,13 @@ where
 
         let tx_size = self.transaction().serialized_size() as Word;
 
-        self.registers[REG_GGAS] = self.transaction().limit();
-        self.registers[REG_CGAS] = self.transaction().limit();
+        let gas = self
+            .transaction()
+            .limit()
+            .checked_sub(gas_used_by_predicates)
+            .expect("Bug! Predicates have used more gas than allowed");
```
Voxelot, line 50: Use bug error type

Voxelot: Also since the VM doesn't check whether the predicates flag is checked, I don't think this is necessarily a bug/panic. It could just be a regular error.

Voxelot:
Converted PredicateVerificationFailed into an error enum with more descriptive failure reasons. 
Also replaced panics with `Bug` error type.

Voxelot (dismissed):
I approve but someone else should review the changes I recently pushed @mitch-fuel @Dentosal

xgreenx (changes requested):
TODO

fuel-vm/src/checked_transaction.rs
```diff
@@ -128,25 +151,90 @@ impl<Tx: IntoChecked> Borrow<Tx> for Checked<Tx> {
...
 
+/// Performs predicate verification for a transaction
+pub trait CheckPredicates {
+    /// Define predicate verification logic (if any)
+    fn check_predicates(self, params: &ConsensusParameters, gas_costs: GasCosts) -> Result<Self, CheckError>
+    where
+        Self: Sized;
+}
+
+impl<Tx: IntoChecked + ExecutableTransaction + fuel_tx::field::GasLimit> CheckPredicates for Checked<Tx>
```
xgreenx, line 186: Why do you need the `fuel_tx::field::GasLimit` constrain here?

xgreenx: `ExecutableTransaction` contains all constrains(including `IntoChecked`)

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -43,33 +43,53 @@ impl<T> Interpreter<PredicateStorage, T> {
...
     /// VM state required to trace the execution steps.
-    pub fn check_predicates<Tx>(checked: Checked<Tx>, params: ConsensusParameters, gas_costs: GasCosts) -> bool
+    pub fn check_predicates<Tx>(
+        checked: Checked<Tx>,
+        params: ConsensusParameters,
+        gas_costs: GasCosts,
+    ) -> Result<PredicatesChecked, PredicateVerificationFailed>
     where
-        Tx: ExecutableTransaction,
+        Tx: ExecutableTransaction + fuel_tx::field::GasLimit,
```
xgreenx, line 52: Hmm, how that worked before without `fuel_tx::field::GasLimit`? I guess you need functionality from `Chargeable` trait.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -43,33 +43,53 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+        let mut remaining_gas = tx_gas_limit;
+
+        vm.init_predicate(checked);
+
+        for predicate in predicates {
+            // VM is cloned because the state should be reset for every predicate verification
+            let mut vm = vm.clone();
+
+            vm.context = Context::Predicate { program: predicate };
+            vm.registers[REG_GGAS] = remaining_gas;
```
xgreenx, line 78: Could we add `set_remaining_gas` and `get_remaining_gas` methods and reuse them here(and in the `init` function too)? In this case, we can hide work with registers(and we will not expose unique information that we need to set 2 registers).

fuel-vm/tests/metadata.rs
```diff
@@ -196,14 +198,16 @@ fn get_transaction_fields() {
...
-        .finalize_checked(height, &params);
+        .finalize_checked(height, &params, client.gas_costs());
 
     client.deploy(tx);
 
     let mut predicate = vec![0u8; 128];
     let mut predicate_data = vec![0u8; 512];
 
     rng.fill(predicate.as_mut_slice());
+    // make predicate valid by replacing first bytes with RET(1)
```
xgreenx, line 209: Hmm, it looks weird because we filled the predicate with random data, but after we replaced the first command with the `Opcode::RET(0x01)` -> all random data is useless. Maybe we don't need to sue random bytes at all?

Voxelot: This test is mainly about testing offsets, and not about checking predicate validity. Now that predicate verification is part of IntoChecked, we can't test using invalid predicates anymore.

fuel-vm/src/interpreter.rs
```diff
@@ -132,6 +134,11 @@ impl<S, Tx> Interpreter<S, Tx> {
         &self.params
     }
 
+    /// Gas costs for opcodes
+    pub fn gas_costs(&self) -> GasCosts {
```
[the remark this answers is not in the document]
Voxelot: It is Arc'd internally.

fuel-vm/src/checked_transaction.rs
```diff
@@ -128,25 +151,90 @@ impl<Tx: IntoChecked> Borrow<Tx> for Checked<Tx> {
     }
 }
 
-pub trait IntoChecked: Checkable + Sized {
+/// Performs checks for a transaction
+pub trait IntoChecked: FormatValidityChecks + Sized {
```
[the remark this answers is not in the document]
Voxelot: Yeah that might be a more flexible approach

fuel-vm/tests/metadata.rs
xgreenx: I get it, but we can replace 100 random bytes with one opcode - `Opcode::RET(0x01)`. Because we already introduced a valid predicate

fuel-vm/src/checked_transaction.rs
```diff
@@ -128,25 +151,90 @@ impl<Tx: IntoChecked> Borrow<Tx> for Checked<Tx> {
...
     type Metadata: Sized;
 
     /// Returns transaction that passed all `Checks`.
-    fn into_checked(self, block_height: Word, params: &ConsensusParameters) -> Result<Checked<Self>, CheckError> {
-        self.into_checked_basic(block_height, params)?.check_signatures()
+    fn into_checked(
+        self,
+        block_height: Word,
+        params: &ConsensusParameters,
+        gas_costs: GasCosts,
```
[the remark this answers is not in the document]
Voxelot: At this point we know the VM need to own it, so I'm not sure if we really gain much by adding a reference here.

fuel-vm/src/interpreter.rs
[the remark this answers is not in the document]
xgreenx: Oh, yeah, I forgot about that=D So I'm okay with cloning. 

But it still better to return a reference from an API perspective

fuel-vm/src/checked_transaction.rs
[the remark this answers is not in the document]
xgreenx: Using reference here allow us to optimize it in the future easily. This comment is related to the comment about `gas_costs`

fuel-vm/src/checked_transaction.rs
```diff
@@ -271,73 +388,11 @@ impl IntoChecked for Transaction {
     }
 }
 
-pub(crate) fn initial_free_balances<T>(
```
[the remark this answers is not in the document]
Dentosal: Moved to a new file in the checked_transaction module

fuel-vm/src/checked_transaction.rs
```diff
@@ -43,28 +54,31 @@ impl core::fmt::Display for Checks {
...
+/// it's cleaner to redo the tx check.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Checked<Tx: IntoChecked> {
     transaction: Tx,
     metadata: Tx::Metadata,
     checks_bitmask: Checks,
+    /// If predicates have been checked, this is how much gas checking them used.
+    /// This must be zero if the predicates have not been checked yet.
+    /// Note that the checkedness of predicates is marked in checks_bitmask.
+    gas_used_by_predicates: Word,
```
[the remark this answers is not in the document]
Dentosal: Done

fuel-vm/src/checked_transaction.rs
```diff
@@ -680,6 +739,53 @@ mod tests {
...
+
+        let tx = predicate_tx(&mut rng, 1, 1000000, 1000000);
+        let checked = tx
+            // Sets Checks::Basic
+            .into_checked_basic(block_height, &params)
+            .unwrap()
+            // Sets Checks::Predicates
+            .check_predicates(&params, gas_costs)
+            .unwrap();
+        assert_eq!(*checked.checks(), Checks::Basic | Checks::Predicates)
```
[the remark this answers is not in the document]
Dentosal: Done

fuel-vm/src/checked_transaction/types.rs
```diff
@@ -0,0 +1,113 @@
+#![allow(missing_docs)]
```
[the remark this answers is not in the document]
Dentosal: Also for the `mod`s. Done.

fuel-vm/src/error.rs
```diff
@@ -223,6 +223,39 @@ impl From<Infallible> for io::Error {
...
+    #[error("Predicate failed to evaluate")]
+    False,
+    /// An unexpected error occurred.
+    #[error(transparent)]
+    Io(#[from] io::Error),
+}
+
+impl From<PredicateVerificationFailed> for CheckError {
+    fn from(_: PredicateVerificationFailed) -> Self {
+        CheckError::PredicateVerificationFailed
```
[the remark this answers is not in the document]
Dentosal: The error is an empty newtype. We don't currently collect any error information.

fuel-vm/src/interpreter/executors/main.rs
Dentosal: We would need two getters, since remaining gas could mean either GGAS or CGAS. I don't really see any point in abstracting this away.

fuel-vm/src/error.rs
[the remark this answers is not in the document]
xgreenx: I meant to add the `PredicateVerificationFailed` into `CheckError::PredicateVerificationFailed`, because they all leave in the `fuel-vm` now, we can do that

fuel-vm/src/interpreter/executors/main.rs
xgreenx: `set_remaining_gas` should be useful=)

Regarding `remaining_gas`, seems we already have this kind of the function 
image

fuel-vm/src/checked_transaction.rs
Dentosal: `Interpreter::check_predicates` requires that, since it needs the gas limit of the tx now.

xgreenx: `ExecutableTransaction` has superfruit `Chargeable` that contains a `limit` function.

fuel-vm/src/checked_transaction/builder.rs
```diff
@@ -0,0 +1,59 @@
+//! Extension trait for [`fuel_tx::TransactionBuilder`]
+
+use super::{Checked, IntoChecked};
+use crate::prelude::*;
+use fuel_tx::ConsensusParameters;
+
+/// Extension trait for [`fuel_tx::TransactionBuilder`] adding finalization methods
+pub trait TransactionBuilderExt<Tx>
```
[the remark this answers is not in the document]
Dentosal: Done. I ended up duplicating the `finalize` instead of moving it to avoid having to import `Finalizable` when not calling it from a generic context.

fuel-vm/src/checked_transaction.rs
Dentosal: It's only named `limit`? Weird name choice. Anyway, I changed to the code to use that instead.

fuel-vm/src/interpreter/executors/main.rs
Dentosal: I disagree about usefulness, since the resulting function is quite error prone; it's rather rare to set both CGAS and GGAS. Added the function anyways: 0669c3358c4722b55b00d30d2780424d62b0788a

fuel-vm/src/error.rs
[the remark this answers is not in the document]
Dentosal: I still don't understand what you mean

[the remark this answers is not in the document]
xgreenx: Never mind, `CheckError` lives in the `fuel-tx`. We can't use it in the `PredicateVerificationFailed` from `fuel-vm` there

xgreenx (dismissed):

ControlCplusControlV (approved):

Voxelot (approved):

PR #310, opened 2023-01-16
Move Checked<Tx> type to the fuel-vm crate

Dentosal:


fuel-tx/src/checked_transaction.rs
```diff
@@ -1,856 +1 @@
-//! A checked transaction is type-wrapper for transactions which have been checked.
```
Voxelot, line 1: Too bad git wasn't smart enough to detect this file was moved instead of created in fuel-vm, that way it could preserve some of the history.

fuel-vm/src/checked_transaction.rs
```diff
@@ -0,0 +1,870 @@
...
+    }
+}
+
+/// Performs checks for a transaction
+pub trait IntoChecked: Checkable + Sized {
+    /// Metadata produced during the check.
+    type Metadata: Sized;
+
+    /// Returns transaction that passed all `Checks`.
+    /// TODO: Predicate checks
```
Voxelot, line 173: Are predicate checks going to be added once this is merged into the target branch?

Dentosal: Yes

fuel-tx/src/checked_transaction.rs
Dentosal: Yep, too many modifications.

mitch-fuel (approved):
looks good

PR #312, opened 2023-01-18
rename check to check_basic_and_signature

Voxelot:
Make the intent of checkable more clear that it doesn't cover all potential checks.

I originally set out to extend the checkable trait in the VM to replace the existing full check method with one that also includes predicate execution. However, after some more digging it didn't really make sense to enable predicate verification outside of CheckedTransactions, since predicate verification depends on extra verification not performed by the Checkable trait (such as computing free balances and tracking remaining gas).

Removing Checkable isn't a very clean solution either, since it's used within the impl's of IntoChecked for basic verification.

Ultimately Checkable just means that a transaction has some simple validation procedures that should be performed before attempting to convert into a fully checked transaction type. This is limited to checks that don't compute any sort of dynamic outputs based on input (i.e. remaining gas or free balances) that would be cached as metadata for further execution later on. So this PR scaled back the original ambitions [here](https://github.com/FuelLabs/fuel-vm/pull/307#issuecomment-1386107745) to just adding clarification about what this trait is useful for.


fuel-tx/src/transaction/checkable.rs
```diff
@@ -157,12 +157,17 @@ impl Output {
     }
 }
 
-/// Means that the transaction can be validated.
+/// Contains logic for stateless validations that don't result in any reusable metadata such
+/// as spendable input balances or remaining gas.
 pub trait Checkable {
```
xgreenx, line 162: Maybe better to rename the trait itself into `BasicallyCheckable` or `SupportBasicChecks`? In this case, we don't need to rename `check` and only update the comment that it runs all basic checks=)

Voxelot: Good idea, updated the PR. What do you think of `FormatValidityChecks`, as it seems a little less ambiguous than basic.

xgreenx: Yep, good naming

Voxelot: Maybe we should even consider renaming usages of `basic` in checked transactions to `formatted`?

xgreenx (approved):

fuel-tx/src/transaction/checkable.rs
xgreenx: Then we need to do it for all "Basic"=D
image

PR #317, opened 2023-01-20
Update default gas costs

Voxelot:
After fixing some issues in the benchmarking framework, I reran the suite to correct the cost of `srwq`. The srwq base cost is greatly reduced now.

We will need to update these prices again once SMT merklization is implemented on fuel-core.


fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -1,24 +1,24 @@
...
         burn: 35,
         cb: 2,
         cfei: 1,
         cfsi: 1,
-        croo: 163,
+        croo: 28,
         div: 1,
         divi: 1,
-        ecr: 1601,
+        ecr: 1703,
```
ControlCplusControlV, line 21: Interesting to see `ecr` go up by so much while most others went down

Voxelot: It varies by about 100 each time I run the test, we likely need larger sample sizes but this good enough approximation for now.

xgreenx (approved):

PR #319, opened 2023-01-23
Remove lifetime from the `Mappable::Key`

xgreenx:


fuel-vm/src/storage.rs
```diff
@@ -49,9 +52,109 @@ impl Mappable for ContractsAssets {
...
+/// Both types are represented by one big array. It is done from the performance perspective
+/// to minimize the number of copies. The current code doesn't use consumed values and uses
+/// it in most cases as on big key(except tests, which require access to sub-keys).
+/// But in the future, we may change the layout of the fields based on
+/// the performance measurements/new business logic.
+macro_rules! double_key {
+    ($i:ident, $first:ident, $first_getter:ident, $second:ident, $second_getter:ident) => {
+        #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+        /// The FuelVM storage double key.
+        pub struct $i([u8; { $first::LEN + $second::LEN }]);
```
xgreenx, line 75: On the `fuel-core` side, we create a composite key(it is called `MultiKey`) to work with the storage. With this double storage key, I want to provide this functionality(of multikey). 

The `fuel-core` is responsible for types and tuples serialization. But the layout of the inner field may help with the serialization performance.

We can have two fields and represent(via `mem::transmute`) the structure as a bytes array for `AsRef<[u8]>`, or we can have one big array(that contains two fields) and then the implementation of the `AsRef<u8>` is very native.

I prefer the approach with one big array=)

Voxelot (approved):

PR #331, opened 2023-01-26
Added additional methods for storage for the cases when the developer need to work with `StorageAsRef` and `StorageAsMut`

xgreenx:


fuel-storage/src/lib.rs
```diff
@@ -123,6 +123,14 @@ pub struct StorageRef<'a, T: 'a + ?Sized, Type: Mappable>(&'a T, core::marker::P
 pub trait StorageAsRef {
     #[inline(always)]
     fn storage<Type>(&self) -> StorageRef<Self, Type>
+    where
+        Type: Mappable,
```
ControlCplusControlV, line 127: What is the main motivation behind this PR? Is it blocking you somewhere else? If so do we still have time to sneak it in before next release for everything?

fuel-storage/src/impls.rs
```diff
@@ -53,6 +53,16 @@ impl<'a, T: StorageInspect<Type>, Type: Mappable> StorageRef<'a, T, Type> {
     }
 }
 
+impl<'a, T, Type: Mappable> StorageRef<'a, T, Type> {
+    #[inline(always)]
+    pub fn root<Key>(self, key: &Key) -> Result<MerkleRoot, T::Error>
```
freesig, line 58: Is this a new rust syntax? Where is the `for` in this `impl`

freesig (approved):
Makes sense

ControlCplusControlV (approved):

fuel-storage/src/lib.rs
xgreenx: It it to improve some part of the code in the future to avoid `(&mut database)` or `(&**database)`

PR #336, opened 2023-02-02
Remove `Clone` constraint from `PathIterator`

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-core/issues/527

Removes `Clone` constraint from `PathIterator` by using `take` instead of `clone`. Requiring `Clone` is too cumbersome because it requires that the `ParentNode::Error` (e.g., `StorageError` in the case of `fuel-core`) also implement `Clone`, which cannot be guaranteed.

### DIFF
--- a/fuel-merkle/src/common/path_iterator.rs
+++ b/fuel-merkle/src/common/path_iterator.rs
@@ -145,14 +145,13 @@ impl<T> Iterator for PathIter<T>
 where
     T: ParentNode + Clone,
     T::Key: Path + Clone,
-    T::Error: Clone,
 {
     type Item = (ChildResult<T>, ChildResult<T>);
 
     fn next(&mut self) -> Option<Self::Item> {
-        let value = self.current.clone();
+        let value = self.current.take();
 
-        if let Some((ref path_node, _)) = self.current {
+        if let Some((ref path_node, _)) = value {
             match path_node {
                 Ok(path_node) if path_node.is_node() => {
                     let path = self.leaf.leaf_key();
--- a/fuel-merkle/src/common/path_iterator.rs
+++ b/fuel-merkle/src/common/path_iterator.rs
@@ -143,8 +143,8 @@ where
 
 impl<T> Iterator for PathIter<T>
 where
-    T: ParentNode + Clone,
-    T::Key: Path + Clone,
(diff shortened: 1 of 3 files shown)


fuel-merkle/src/common/path_iterator.rs
```diff
@@ -145,14 +145,13 @@ impl<T> Iterator for PathIter<T>
 where
     T: ParentNode + Clone,
     T::Key: Path + Clone,
```
xgreenx, line 147: Hmm, maybe we also can remove those `Clone`?=)

xgreenx: Yes, we can=)

xgreenx (approved):
I also checked the code around, and it seems that we can simplify `type Item = (ChildResult<T>, ChildResult<T>)` into `type Item = (T, T)`, because `left_child` and `right_child` can't return an error based on the current implementation.


Also, you can accept `Self` instead of `&Self` here:

```rust
pub trait AsPathIterator<T: ParentNode> {
    fn as_path_iter(&self, leaf: &Self) -> PathIter<T>;
}
```
It will remove `Clone` from here and from `new` function too=)

fuel-merkle/src/common/path_iterator.rs
bvrooman: Good catch!

bvrooman:
> I also checked the code around, and it seems that we can simplify `type Item = (ChildResult<T>, ChildResult<T>)` into `type Item = (T, T)`, because `left_child` and `right_child` can't return an error based on the current implementation.

The Sparse Node `left_child` and `right_child` fns return `ChildResult<Self>` because they can return errors, such as key not found or database error: https://github.com/FuelLabs/fuel-vm/blob/master/fuel-merkle/src/sparse/node.rs#L290.

bvrooman:
> Also, you can accept `Self` instead of `&Self` here:
> 
> ```rust
> pub trait AsPathIterator<T: ParentNode> {
>     fn as_path_iter(&self, leaf: &Self) -> PathIter<T>;
> }
> ```
> 
> It will remove `Clone` from here and from `new` function too=)

Sounds good, I can try that too and put up a new PR

PR #339, opened 2023-02-03
always dry run publish

Voxelot:
*Test PR*

Checking to see if the publish action will support dry running on every ci workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,6 @@ jobs:
 
   publish-check:
     runs-on: ubuntu-latest
-    if: github.event_name == 'release'
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -108,12 +124,14 @@ jobs:
       - cargo
       - cargo-toml-fmt-check
       - cargo-unused-deps-check
+      - publish-always-check
     runs-on: ubuntu-latest
     steps:
       - run: true
 
-  publish-check:
+  publish-final-check:
```
Voxelot, line 132: We should split this job into two parts, tag verification vs actual crate publishing. This name seems misleading as we are doing more than checking things here, this job will have side-effects by doing actual publishing

PR #340, opened 2023-02-03
Iterate through instruction encondings using codegen

Dentosal:
Pulls in `strum` dependency to generate enum variant iterators automatically


Voxelot (approved):

fuel-asm/src/macros.rs
```diff
@@ -286,6 +286,73 @@ macro_rules! impl_instructions {
...
+        impl_instructions!(impl_opcode_test_construct $($rest)*);
+    };
+    (impl_opcode_test_construct) => {};
+
+    (impl_opcode_test_construct_fn [RegId]) => {
+        #[doc(hidden)]
+        /// Construct the instruction from all possible raw fields, ignoring inapplicable ones.
+        pub fn test_construct(ra: RegId, _rb: RegId, _rc: RegId, _rd: RegId, _imm: u32) -> Self {
+            Self(pack::bytes_from_ra(ra))
+        }
```
mitchmindtree, line 303: If we're only using these for testing, maybe we should conservatively add them behind `#[cfg(test)]`?

fuel-asm/src/macros.rs
```diff
@@ -606,6 +674,17 @@ macro_rules! impl_instructions {
...
+        impl Opcode {
+            /// Construct the instruction from all possible raw fields, ignoring inapplicable ones.
+            pub fn test_construct(self, ra: RegId, rb: RegId, rc: RegId, rd: RegId, imm: u32) -> Instruction {
+                match self {
+                    $(
+                        Self::$Op => Instruction::$Op(crate::_op::$Op::test_construct(ra, rb, rc, rd, imm)),
+                    )*
+                }
+            }
+        }
```
mitchmindtree, line 687: Same for this

mitchmindtree (commented):
Nice lgtm, just one minor suggestion

fuel-asm/src/macros.rs
Dentosal: We would have to change the encoding test to be an unit test, as integration tests cannot use `#[cfg(test)]` features

mitchmindtree: Ahh in that case, maybe we would be better off moving these tests into unit tests? I'm a bit weary of exposing this constructor in our public API is all - ideally we'd only expose stuff we expect downstream to actually use.

Voxelot: doesn't really seem like the encoding tests need to be integ tests anyways

mitchmindtree (approved):
:ok_hand:

PR #342, opened 2023-02-07
Read contract directly into memory

freesig:
This adds a few traits:
- `StorageSize` Allows checking the size of a value without copying or de-serializing into memory
- `StorageRead` Like `io::Read` except takes a key. Not this deals with raw bytes and avoids any de-serialization.
- `StorageWrite` Like `io::Write` except takes a key. Not this deals with raw bytes and avoids any serialization.

This allows cheap copying of contract data directly from storage into the vm memory.
This PR replaces #272


fuel-vm/src/call.rs
```diff
@@ -124,43 +123,53 @@ impl CallFrame {
...
+    pub const fn contract_id_offset() -> usize {
+        0
     }
 
-    /// Contract code memory offset.
-    pub const fn code_offset() -> usize {
-        Self::code_size_offset() + WORD_SIZE
+    /// Start of the asset id offset from the beginning of the call frame.
+    pub const fn asset_id_offset() -> usize {
+        ContractId::LEN
```
bvrooman, line 147: Perhaps something like this would make the layout relationship a bit more clear and automatic: 
```suggestion
        Self::contract_id_offset() + ContractId::LEN
```

bvrooman (dismissed):
LGTM, one minor suggestion

fuel-storage/src/impls.rs
```diff
@@ -35,6 +38,38 @@ impl<'a, T: StorageMutate<Type> + ?Sized, Type: Mappable> StorageMutate<Type> fo
     }
 }
 
+impl<'a, T: StorageSize<Type> + ?Sized, Type: Mappable> StorageSize<Type> for &'a T {
+    fn size_of_value(&self, key: &<Type as Mappable>::Key) -> Result<Option<usize>, Self::Error> {
+        <T as StorageSize<Type>>::size_of_value(self, key)
+    }
+}
+
+impl<'a, T: StorageSize<Type> + ?Sized, Type: Mappable> StorageSize<Type> for &'a mut T {
```
Voxelot, line 47:
```suggestion
impl<'a, T: StorageSize<Type> + ?Sized, Type: Mappable> StorageSize<Type> for &'a T {
```

this doesn't seem like it should require a `mut`

fuel-storage/src/lib.rs
```diff
@@ -70,6 +73,66 @@ pub trait StorageMutate<Type: Mappable>: StorageInspect<Type> {
...
+}
+
+/// Base storage trait for Fuel infrastructure.
+///
+/// Allows writing the raw bytes of the value stored to a given key
+/// from a provided buffer.
+///
+/// This trait should skip any serialization and simply copy the raw bytes
+/// to the storage.
+pub trait StorageWrite<Type: Mappable>: StorageMutate<Type> {
```
xgreenx, line 114:
```suggestion
pub trait StorageWrite<Type: Mappable>: StorageMutate<Type> + StorageRead<Type> {
```

fuel-storage/src/lib.rs
```diff
@@ -70,6 +73,66 @@ pub trait StorageMutate<Type: Mappable>: StorageInspect<Type> {
...
+
+    /// Write the value to the given key from the provided buffer and
+    /// return the previous value if it existed.
+    ///
+    /// Does not perform any serialization.
+    ///
+    /// Returns the number of bytes written and the previous value if it existed.
+    fn replace(&self, key: &Type::Key, buf: Vec<u8>) -> Result<(usize, Option<Vec<u8>>), Self::Error>
+    where
+        Self: StorageSize<Type>;
```
xgreenx, line 130:
```suggestion
    fn replace(&mut self, key: &Type::Key, buf: Vec<u8>) -> Result<(usize, Option<Vec<u8>>), Self::Error>;
```

fuel-storage/src/lib.rs
```diff
@@ -70,6 +73,66 @@ pub trait StorageMutate<Type: Mappable>: StorageInspect<Type> {
...
+///
+/// This trait should skip any serialization and simply copy the raw bytes
+/// to the storage.
+pub trait StorageWrite<Type: Mappable>: StorageMutate<Type> {
+    /// Write the value to the given key from the provided buffer.
+    ///
+    /// Does not perform any serialization.
+    ///
+    /// Returns the number of bytes written.
+    fn write(&self, key: &Type::Key, buf: Vec<u8>) -> Result<usize, Self::Error>;
```
xgreenx, line 120:
```suggestion
    fn write(&mut self, key: &Type::Key, buf: Vec<u8>) -> Result<usize, Self::Error>;
```

fuel-storage/src/lib.rs
```diff
@@ -70,6 +73,66 @@ pub trait StorageMutate<Type: Mappable>: StorageInspect<Type> {
...
+    ///
+    /// Does not perform any serialization.
+    ///
+    /// Returns the number of bytes written and the previous value if it existed.
+    fn replace(&self, key: &Type::Key, buf: Vec<u8>) -> Result<(usize, Option<Vec<u8>>), Self::Error>
+    where
+        Self: StorageSize<Type>;
+
+    /// Removes a value from the storage and returning it without deserializing it.
+    fn take(&self, key: &Type::Key) -> Result<Option<Vec<u8>>, Self::Error>;
```
xgreenx, line 133:
```suggestion
    fn take(&mut self, key: &Type::Key) -> Result<Option<Vec<u8>>, Self::Error>;
```

fuel-vm/src/call.rs
freesig: I'm not sure I follow? The asset id offset is one `ContractId::LEN` from the start. Or do you mean `Self::contract_id_offset()` returns `0`?

fuel-storage/src/lib.rs
freesig: This is already inferred through `StorageMutate`

fuel-vm/src/call.rs
bvrooman: Yeah, exactly, it looks like contract id is at 0, and asset id comes right after contract id. So it makes sense that it would be one `ContractId::LEN`.  but if we ever change `contract_id_offset`, then including it in the `asset_id_offset` calculation will ensure that it is still in the right position.

fuel-storage/src/lib.rs
xgreenx: When you don't need `Self: StorageSize<Type>` below?

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -108,9 +108,11 @@ where
 
         // update frame pointer, if we have a stack frame (e.g. fp > 0)
         if fp > 0 {
-            let fpx = add_usize(fp, CallFrame::code_size_offset());
+            let fp_code_size = add_usize(fp, CallFrame::code_size_offset());
+            let fp_code_size_end = add_usize(fp_code_size, WORD_SIZE);
 
-            self.memory[fp..fpx].copy_from_slice(&length.to_be_bytes());
+            let length = length as Word;
```
Voxelot, line 114: > Increment $fp->codesize, $ssp, and $sp by $rC padded to word alignment.
https://github.com/FuelLabs/fuel-specs/blob/master/src/vm/instruction_set.md#ldc-load-code-from-an-external-contract

Shouldn't this be added to the existing code size instead of replacing it with the length of code loaded?

Voxelot, line 114: > https://github.com/FuelLabs/fuel-specs/blob/master/src/vm/instruction_set.md#ldc-load-code-from-an-external-contract
Shouldn't the length of LDC be added to the existing size of the contract instead of replacing it?

pseudocode suggestion:
```suggestion
            let length = Word::from_be_bytes(self.memory[fp_code_size..fp_code_size_end]).checked_add(length as word).ok_or_else(|| PanicReason::MemoryOverflow.into())?;
```

fuel-storage/src/lib.rs
freesig: Because of `pub trait StorageMutate<Type: Mappable>: StorageInspect<Type> `

Voxelot (approved):

Voxelot:
we should also consider doing this for contract state 🤔

fuel-storage/src/lib.rs
xgreenx: You misunderstood me=) I mentioned `StorageRead`, not `StorageInspect`.

PR #344, opened 2023-02-08
Fix panics in fuel-vm for exponent or bitshift larger than u32::MAX

Dentosal:
Discovered when writing spec-compliance test suite. This PR doesn't contain test cases for these bugs. Those will be done in a follow-up PR, containing the whole test suite.


fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -235,8 +247,17 @@ where
             Instruction::SLL(sll) => {
                 self.gas_charge(self.gas_costs.sll)?;
                 let (a, b, c) = sll.unpack();
-                let rhs = u32::try_from(r!(c)).expect("value out of range");
-                self.alu_set(a.into(), r!(b).checked_shl(rhs).unwrap_or_default())?;
+
+                self.alu_error(
+                    a.into(),
+                    |l, r| {
+                        l.checked_shl(u32::try_from(r).expect("value out of range"))
```
Voxelot, line 254: wouldn't this still panic if unsafe math is disabled?

freesig (approved):
Nice

PR #345, opened 2023-02-09
Spec compliance testing, round 1

Dentosal:
Steps towards verifying some fundamental spec compliance features, namely:

* Opcodes increment $pc by one
* Immediate operands are zero-extended correctly
* ALU operations set $err and $of registers correctly
* Reserved registers cannot be written to

Builds on top of #340 and #344


ControlCplusControlV (approved):
Good work, thanks for using rstest to cover specific cases

PR #346, opened 2023-02-10
Refactor of `fuel-crypto` for clarity around safety

mitchmindtree:
This started with the aim of carrying over https://github.com/FuelLabs/fuel-crypto/pull/36 which aimed to address https://github.com/FuelLabs/fuel-crypto/issues/35, though after noticing a few other instances of the same issue decided to take a closer look throughout the crate.

This PR removes many of the unsafe constructors exposed in the `fuel-crypto` API, some due to the implementations not actually being `unsafe`, and some due to a revised implementation that safely encapsulates the unsafe behaviour.

The `Message` and `Signature` `from_bytes` and `from_bytes_ref` constructors have been documented to clarify expectations that they expect signature or cryptographically hashed data respectively.

The `PublicKey` and `SecretKey` `from_bytes_unchecked` constructors have been made private in order to ensure holding an instance of one of these types guarantees their inner byte arrays at least satisfies the curve.

The original, custom curve-checking functions have been removed in favour of using the secp256k1 library constructors directly which share the same implementation but with an added range check that is likely to get optimised out due to the use of const-sized arrays anyway.

Adds some missing `#[repr(transparent)]` decorators to `fuel-types`.

The `FuelMnemonic` type provided unnecessary indirection between the user and the free function, and has been removed in favour of exposing the function directly from the crate root.


fuel-crypto/src/public.rs
```diff
@@ -236,4 +165,13 @@ mod use_std {
...
         }
     }
+
+    /// Check if the public key byte representation is in the curve.
+    fn public_key_bytes_valid(bytes: &[u8; PublicKey::LEN]) -> Result<(), Error> {
+        let mut public_with_flag = [0u8; UNCOMPRESSED_PUBLIC_KEY_SIZE];
+        public_with_flag[1..].copy_from_slice(bytes);
+        secp256k1::PublicKey::from_slice(&public_with_flag)
+            .map(|_| ())
+            .map_err(|_| Error::InvalidPublicKey)
```
mitchmindtree, line 175: TODO: The original `is_slice_in_curve_unchecked` function would also leave the first byte zeroed before invoking the `secp256k1` pubkey parser like this, so I've carried over the behaviour in this function. However, it looks like we should possibly be setting the first byte to `SECP_UNCOMPRESSED_FLAG`? Should investigate or open an issue before landing.

fuel-crypto/src/signature.rs
```diff
@@ -12,49 +12,28 @@ use core::{fmt, str};
...
+    /// Construct a `Signature` directly from its bytes.
     ///
-    /// # Safety
-    ///
-    /// There is no guarantee the provided bytes will be a valid signature. Internally, some FFI
-    /// calls to `secp256k1` are performed and we might have undefined behavior in case the bytes
-    /// are not canonically encoded to a valid `secp256k1` signature.
-    pub unsafe fn from_bytes_unchecked(bytes: [u8; Self::LEN]) -> Self {
+    /// This constructor expects the given bytes to be a valid signature. No signing is performed.
+    pub fn from_bytes(bytes: [u8; Self::LEN]) -> Self {
```
freesig, line 21: What happens if this is not a valid signature? I know we use this function in fuel core with bytes from the network so maybe we need to add `from_bytes_checked -> Option<Self>`

fuel-crypto/src/message.rs
```diff
@@ -12,54 +10,39 @@ use core::ops::Deref;
...
-    /// `Self::LEN`. Instead, it will cause undefined behavior and read random
-    /// disowned bytes.
+    /// Construct a `Message` reference directly from a reference to its bytes.
     ///
-    /// This function extends the unsafety of [`Self::from_bytes_unchecked`].
-    pub unsafe fn from_slice_unchecked(bytes: &[u8]) -> Self {
-        Self(Bytes32::from_slice_unchecked(bytes))
+    /// This constructor expects the given bytes to be a valid,
+    /// cryptographically hashed message. No hashing is performed.
+    pub fn from_bytes_ref(bytes: &[u8; Self::LEN]) -> &Self {
```
freesig, line 37: Does this actually uphold the lifetime. If the bytes array is moved after this call and `&Self` is still being used, do we get a compile error? If not this should still be an unsafe function right?

fuel-vm/src/interpreter/crypto.rs
```diff
@@ -22,11 +22,13 @@ where
...
 
+        // TODO: These casts may overflow/truncate on 32-bit?
         let (a, b, bx, c, cx) = (a as usize, b as usize, bx as usize, c as usize, cx as usize);
 
-        // Safety: memory bounds are checked
-        let signature = unsafe { Signature::as_ref_unchecked(&self.memory[b..bx]) };
-        let message = unsafe { Message::as_ref_unchecked(&self.memory[c..cx]) };
+        let sig_bytes = <&_>::try_from(&self.memory[b..bx]).expect("memory bounds checked");
+        let msg_bytes = <&_>::try_from(&self.memory[c..cx]).expect("memory bounds checked");
+        let signature = Signature::from_bytes_ref(sig_bytes);
```
freesig, line 30: Should these be results instead of panics?

mitchmindtree: If these fail, I believe it's a bug in our implementation logic, i.e. these should always be correct because `bx` and `cx` are exactly `Bytes*::LEN` greater than `b` or `x` respectively.

fuel-crypto/src/signature.rs
mitchmindtree: Afaik there's no way to know if the signature is valid until we actually try to validate the transaction, as we don't know what data has been signed at this point.

It should be fine to use this with bytes read from the network as if the signature isn't valid, the transaction validation will fail because the signature won't be valid for some spend or whatever.

fuel-crypto/src/message.rs
mitchmindtree: > Does this actually uphold the lifetime.

Yes, in this case the lifetime of the reference in `&Self` is bound by the only input lifetime, i.e. the lifetime of `bytes: &[u8; Self::LEN]`. In this case, Rust elides the lifetime because there is only one possible valid lifetime for the output. Here is what the un-ellided signature would look like:

```rust
    pub fn from_bytes_ref<'a>(bytes: &'a [u8; Self::LEN]) -> &'a Self {
```

Voxelot:
This would also be a good opportunity to upgrade from `secp256k1:0.24` to `secp256k1:0.26`

freesig (approved):

mitchmindtree:
Feel free to merge this when ready (keeping in mind there is a little breakage of the pub API).

Issue #350, opened 2023-02-13
Make all op functions unit testable


freesig:
Our Op functions are currently not unit testable because they all take `self`. This creates difficulties in verifying their correct functioning. We should to make all OP functions unit testable by removing `self` and only passing in exactly what is needed.

## Before
```rust
pub(crate) fn gas_charge(
    &mut self,
    gas: Word,
) -> Result<(), RuntimeError>
{
    if gas > self.registers[RegId::CGAS] {
        self.registers[RegId::GGAS] = arith::sub_word(self.registers[RegId::GGAS], self.registers[RegId::CGAS])?;
        self.registers[RegId::CGAS] = 0;

        Err(PanicReason::OutOfGas.into())
    } else {
        self.registers[RegId::CGAS] = arith::sub_word(self.registers[RegId::CGAS], gas)?;
        self.registers[RegId::GGAS] = arith::sub_word(self.registers[RegId::GGAS], gas)?;

        Ok(())
}
```
## After
```rust
pub(crate) fn gas_charge(
    mut cgas: RegMut<CGAS>,
    mut ggas: RegMut<GGAS>,
    gas: Word,
) -> Result<(), RuntimeError>
{
    if gas > *cgas {
        *ggas = arith::sub_word(*ggas, *cgas)?;
        *cgas = 0;

        Err(PanicReason::OutOfGas.into())
    } else {
        *cgas = arith::sub_word(*cgas, gas)?;
        *ggas = arith::sub_word(*ggas, gas)?;

        Ok(())
    }
}
```

PR #353, opened 2023-02-15
Make prepare_call unit testable

freesig:
This PR introduces lots of utilities (especially around multiple mutable borrows of arrays, registers and memory) so that the scope of what is passed into a function is only related to that function and then can be unit tested. 
Refs #350


fuel-vm/src/interpreter/gas/tests.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+#[derive(Debug, PartialEq, Eq)]
+struct GasChargeOutput {
+    cgas: u64,
+    ggas: u64,
+}
+#[test_case(GasChargeInput{cgas: 0, ggas: 0, gas: 0} => Ok(GasChargeOutput{ cgas: 0, ggas: 0}); "zero")]
+#[test_case(GasChargeInput{cgas: 0, ggas: 0, gas: 1} => Err(RuntimeError::Recoverable(PanicReason::OutOfGas)); "no gas")]
+// Currently panics
+// #[test_case(GasChargeInput{cgas: 2, ggas: 0, gas: 1} => Err(RuntimeError::Recoverable(PanicReason::OutOfGas)); "no global gas")]
```
Dentosal, line 17: Why does this panic? Should we have a `TODO` comment here?

freesig: It panics when the `optimized` feature flag is set. Which I don't know how to disable just this test case when that flag is set

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -0,0 +1,342 @@
...
+    GGAS, ggas
+    CGAS, cgas
+    BAL, bal
+    IS, is
+    RET, ret
+    RETL, retl
+    FLAG, flag
+}
+
+pub(crate) struct ReadRegisters<'a> {
```
Voxelot, line 114: I think the Read vs Write distinction is somewhat confusing, since we need to modify these registers internally within the VM. Perhaps this should be renamed to something like `InternalRegisters` or `ProtectedRegisters`.

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -0,0 +1,342 @@
...
+        *read_registers.hp,
+        *read_registers.err,
+        *read_registers.ggas,
+        *read_registers.cgas,
+        *read_registers.bal,
+        *read_registers.is,
+        *read_registers.ret,
+        *read_registers.retl,
+        *read_registers.flag,
+        write_registers.0[0],
```
xgreenx, line 198: You can copy slices or use [concat-arrays](https://github.com/canndrew/concat-arrays). The compiler should be clever enough to optimize it

fuel-vm/src/interpreter/blockchain/test.rs
```diff
@@ -6,10 +6,10 @@ mod scwq;
 mod srwq;
 mod swwq;
 
-fn mem(chains: &[&[u8]]) -> Vec<u8> {
+fn mem(chains: &[&[u8]]) -> Box<[u8; VM_MEMORY_SIZE]> {
```
xgreenx, line 9: Hmm, we have `Box<[u8; VM_MEMORY_SIZE]>` in many places. Maybe we want to add an alias like `VMMemory` or `Memory` or `Buffer` or `VMBuffer`

fuel-vm/src/interpreter/flow.rs
```diff
@@ -149,138 +161,385 @@ where
...
+    sp: RegMut<'a, SP>,
+    ssp: RegMut<'a, SSP>,
+    fp: RegMut<'a, FP>,
+    pc: RegMut<'a, PC>,
+    is: RegMut<'a, IS>,
+    bal: RegMut<'a, BAL>,
+    cgas: RegMut<'a, CGAS>,
+    ggas: RegMut<'a, GGAS>,
+}
+struct PrepareCallRegisters<'a> {
```
xgreenx, line 262:
```suggestion
}

struct PrepareCallRegisters<'a> {
```

fuel-vm/src/interpreter/flow.rs
```diff
@@ -149,138 +161,385 @@ where
...
+
+struct PrepareCallInput<'vm, S> {
+    params: PrepareCallParams,
+    registers: PrepareCallRegisters<'vm>,
+    memory: PrepareCallMemory<'vm>,
+    context: &'vm mut Context,
+    gas_cost: DependentCost,
+    runtime_balances: &'vm mut RuntimeBalances,
+    storage: &'vm mut S,
+    input_contracts: Vec<fuel_types::ContractId>,
```
xgreenx, line 298: It seems we can replace a vector with an iterator. But we can do that in the follow-up PR

fuel-vm/src/interpreter/flow.rs
```diff
@@ -149,138 +161,385 @@ where
...
+    let mut code_range = code_mem_range.clone();
+    code_range.grow_start(CallFrame::serialized_size());
+    code_range.shrink_end(frame.code_size_padding() as usize);
+    let bytes_read = storage
+        .storage::<ContractsRawCode>()
+        .read(frame.to(), code_range.write(memory))
+        .map_err(RuntimeError::from_io)?
+        .ok_or(PanicReason::ContractNotFound)?;
+    if bytes_read as Word != frame.code_size() {
+        return Err(PanicReason::ContractNotFound.into());
```
xgreenx, line 442: It would be nice to have another error here

fuel-vm/src/interpreter/flow.rs
```diff
@@ -149,138 +161,385 @@ where
...
+            retl: registers.unused_registers.retl,
+            flag: registers.unused_registers.flag,
+        }
     }
+}
 
-    pub(crate) fn call(&mut self, a: Word, b: Word, c: Word, d: Word) -> Result<ProgramState, RuntimeError> {
-        self._prepare_call(a, b, c, d)?;
-        self.run_call()
+impl<'reg> From<&'reg mut [Word; VM_REGISTER_COUNT]> for PrepareCallRegisters<'reg> {
```
xgreenx, line 495: [nit] Maybe we want to have some alias for `[Word; VM_REGISTER_COUNT]`

fuel-vm/src/interpreter/gas.rs
```diff
@@ -21,37 +27,102 @@ impl<S, Tx> Interpreter<S, Tx> {
...
+            is: is.as_ref(),
+            current_contract,
+            profiler: &mut self.profiler,
+        };
+        dependent_gas_charge(cgas, ggas, profiler, gas_cost, arg)
     }
 
     pub(crate) fn gas_charge(&mut self, gas: Word) -> Result<(), RuntimeError> {
+        let current_contract = self.contract_id();
+        let ReadRegisters { pc, ggas, cgas, is, .. } = split_registers(&mut self.registers).0;
```
xgreenx, line 43: Hmm, do we really need to split it each time?

fuel-vm/src/interpreter/gas.rs
```diff
@@ -21,37 +27,102 @@ impl<S, Tx> Interpreter<S, Tx> {
...
+pub(crate) fn gas_charge(
+    cgas: RegMut<CGAS>,
+    ggas: RegMut<GGAS>,
+    mut profiler: ProfileGas<'_>,
+    gas: Word,
+) -> Result<(), RuntimeError> {
+    profiler.profile(cgas.as_ref(), gas);
+    gas_charge_inner(cgas, ggas, gas)
+}
+fn gas_charge_inner(mut cgas: RegMut<CGAS>, mut ggas: RegMut<GGAS>, gas: Word) -> Result<(), RuntimeError> {
```
xgreenx, line 90:
```suggestion
}

fn gas_charge_inner(mut cgas: RegMut<CGAS>, mut ggas: RegMut<GGAS>, gas: Word) -> Result<(), RuntimeError> {
```

fuel-vm/src/interpreter/gas.rs
```diff
@@ -21,37 +27,102 @@ impl<S, Tx> Interpreter<S, Tx> {
...
+    ggas: RegMut<GGAS>,
+    mut profiler: ProfileGas<'_>,
+    gas_cost: DependentCost,
+    arg: Word,
+) -> Result<(), RuntimeError> {
+    if gas_cost.dep_per_unit == 0 {
+        gas_charge(cgas, ggas, profiler, gas_cost.base)
+    } else {
+        let cost = dependent_gas_charge_inner(cgas.as_mut(), ggas, gas_cost, arg)?;
+        profiler.profile(cgas.as_ref(), cost);
```
xgreenx, line 66: Do we need to put `profiler` under the `#[cfg(feature = "profile-gas")]` flag?

fuel-vm/src/interpreter.rs
```diff
@@ -159,6 +152,15 @@ impl<S, Tx> Interpreter<S, Tx> {
     }
 }
 
+#[cfg(feature = "profile-gas")]
+fn current_location(
+    current_contract: Option<ContractId>,
+    pc: crate::constraints::reg_key::Reg<{ crate::constraints::reg_key::PC }>,
```
Voxelot, line 158: Maybe these should be aliased somehow? This seems quite verbose / not very ergonomic.

fuel-vm/src/interpreter/internal.rs
```diff
@@ -192,9 +187,8 @@ impl<S, Tx> Interpreter<S, Tx> {
     }
 
     pub(crate) fn set_frame_pointer(&mut self, fp: Word) {
-        self.context.update_from_frame_pointer(fp);
-
-        self.registers[RegId::FP] = fp;
+        let ReadRegisters { fp: register, .. } = split_registers(&mut self.registers).0;
```
xgreenx, line 190: Hmm, we want to split it again. Maybe the compiler will see that we need access only to one register here and will optimize it, but I'm not sure=)

Maybe it is better to store registers like `ReadRegisters` and `Others`.

fuel-vm/src/storage/memory.rs
```diff
@@ -243,7 +271,7 @@ impl MerkleRootStorage<ContractId, ContractsState> for MemoryStorage {
         Ok(crypto::ephemeral_merkle_root(root).into())
     }
 }
-
+impl ContractsAssetsStorage for MemoryStorage {}
```
xgreenx, line 274:
```suggestion

impl ContractsAssetsStorage for MemoryStorage {}

```

fuel-vm/src/interpreter/balances.rs
```diff
@@ -103,22 +105,32 @@ impl RuntimeBalances {
     /// Note: This will not append a new asset into the set since all the assets must be created
     /// during VM initialization and any additional asset would imply reordering the memory
     /// representation of the balances since they must always be ordered, as in the protocol.
-    pub fn checked_balance_add(&mut self, memory: &mut [u8], asset: &AssetId, value: Word) -> Option<Word> {
+    pub fn checked_balance_add(
```
Voxelot, line 108: I don't think there's a usecase for adding to the initial balances. We discussed removing this method, was that done in a separate PR?

fuel-vm/src/interpreter/constructors.rs
```diff
@@ -24,7 +26,7 @@ where
     pub fn with_storage(storage: S, params: ConsensusParameters, gas_costs: GasCosts) -> Self {
         Self {
             registers: [0; VM_REGISTER_COUNT],
-            memory: vec![0; VM_MAX_RAM as usize],
+            memory: vec![0; VM_MEMORY_SIZE].try_into().expect("Falied to allocate memory"),
```
Voxelot, line 29:
```suggestion
            memory: vec![0; VM_MEMORY_SIZE].try_into().expect("Failed to allocate memory"),
```

fuel-vm/src/constraints.rs
```diff
@@ -0,0 +1,118 @@
...
+pub mod reg_key;
+
+#[derive(Clone)]
+/// A range of memory that has been checked that it fits into the VM memory.
+pub struct CheckedMemRange(core::ops::Range<usize>);
+
+#[derive(Clone)]
+/// A range of memory that has been checked that it fits into the VM memory.
+/// This range can be used to read a value of type `T` from memory.
+pub struct CheckedMemValue<T>(CheckedMemRange, core::marker::PhantomData<T>);
```
Voxelot, line 19: It'd be nice if we had a `SerializedSize` trait in fuel-types we could use to constrain `T` with here. Then we could ensure that the correct length is always used.

fuel-vm/src/interpreter.rs
freesig: This is because of feature flags...
There's no nice way to do this but you normally don't need to do it like this when it's not behind a feature flag.

fuel-vm/src/interpreter/gas.rs
freesig: Yes because we need multiple borrows to the registers where at least one borrow is mutable. Also I've benched this and it takes under 1ns vs 3ns if you do it with `split_at_mut`

fuel-vm/src/constraints/reg_key.rs
freesig: That would require initializing a blank array to begin with.

fuel-vm/src/interpreter/blockchain/test.rs
freesig: We can although I do think it's good seeing the size explicitly. It makes it clear how much memory you have when indexing etc.

fuel-vm/src/interpreter/flow.rs
freesig: I don't want to alias this as its very clear what it is from the type.

fuel-vm/src/interpreter/gas.rs
freesig: No this is a blank profiler if the flag is not set

fuel-vm/src/interpreter/internal.rs
freesig: I have benchmarked and it does optimize

fuel-vm/src/interpreter/balances.rs
freesig: Yeh I think it's in the next pr

fuel-vm/src/constraints.rs
freesig: That trait can't be const though

Voxelot: You can have associated constants on a trait though, made an example here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1519603c242a9a86820791fd669b1357

fuel-vm/src/constraints/reg_key.rs
Voxelot: The compiler can usually optimize the blank initialization though if it knows it will later be fully allocated. @xgreenx and I verified this behavior when he implemented the `double_key!` type.

fuel-vm/src/storage/interpreter.rs
```diff
@@ -17,8 +17,8 @@ pub trait InterpreterStorage:
     + StorageSize<ContractsRawCode, Error = Self::DataError>
     + StorageRead<ContractsRawCode, Error = Self::DataError>
     + StorageMutate<ContractsInfo, Error = Self::DataError>
-    + MerkleRootStorage<ContractId, ContractsAssets, Error = Self::DataError>
     + MerkleRootStorage<ContractId, ContractsState, Error = Self::DataError>
+    + ContractsAssetsStorage<Error = Self::DataError>
```
Voxelot, line 21: Should we also do the same thing for state, and have a new `ContractsStateStorage` trait?

fuel-vm/src/interpreter/memory.rs
```diff
@@ -707,8 +712,9 @@ mod tests {
         => (false, [0u8; 100]); "Internal too large for heap"
     )]
     fn test_mem_write(addr: usize, data: &[u8], registers: OwnershipRegisters) -> (bool, [u8; 100]) {
-        let mut memory = [0u8; 100];
-        let r = try_mem_write(addr, data, registers, &mut memory[..]).is_ok();
+        let mut memory: Box<[u8; VM_MEMORY_SIZE]> = vec![0u8; VM_MEMORY_SIZE].try_into().unwrap();
```
Voxelot, line 715: [nit] wouldn't it be cleaner to just use `Box::new` here instead of `try_into().unwrap()`?

```suggestion
        let mut memory: Box<[u8; VM_MEMORY_SIZE]> = Box::new(vec![0u8; VM_MEMORY_SIZE]);
```

fuel-vm/src/constraints/reg_key.rs
Dentosal: Also, hardcoding the register count like this seems error-prone, and it would be nice to avoid that

fuel-vm/src/interpreter/gas/tests.rs
Dentosal: I don't think the optimized-feature should change the behavior, maybe this is a bug?

Anyway, you could do

```rust
#[cfg_attr(
    not(feature = "optimized"),
    test_case(GasChargeInput{cgas: 2, ggas: 0, gas: 1} => Err(RuntimeError::Recoverable(PanicReason::OutOfGas)); "no global gas")
)]
```

fuel-vm/src/interpreter/flow.rs
```diff
@@ -149,138 +161,385 @@ where
...
+        );
 
         // subtract gas
-        self.registers[RegId::CGAS] = arith::sub_word(self.registers[RegId::CGAS], forward_gas_amount)?;
+        *self.registers.read_registers.cgas = arith::sub_word(*self.registers.read_registers.cgas, forward_gas_amount)?;
 
-        *frame.set_context_gas() = self.registers[RegId::CGAS];
-        *frame.set_global_gas() = self.registers[RegId::GGAS];
+        *frame.set_context_gas() = *self.registers.read_registers.cgas;
+        *frame.set_global_gas() = *self.registers.read_registers.ggas;
```
Voxelot, line 368: [nit] this doesn't look like a very conventional way to name rust apis, usually setters would take the value as input, while getting mutable access to some field is done with a `*_mut` method. 

```suggestion
        *frame.context_gas_mut() = *self.registers.read_registers.cgas;
        *frame.global_gas_mut() = *self.registers.read_registers.ggas;
```

fuel-vm/src/interpreter/flow.rs
```diff
@@ -149,138 +161,385 @@ where
...
-                .copy_from_slice(&[0; 32][..pad_len as usize]);
-        }
+        *self.registers.read_registers.sp += len;
+        *self.registers.read_registers.ssp = *self.registers.read_registers.sp;
 
-        self.registers[RegId::BAL] = b;
-        self.registers[RegId::PC] = fpx;
-        self.registers[RegId::IS] = self.registers[RegId::PC];
-        self.registers[RegId::CGAS] = forward_gas_amount;
+        let code_mem_range = CheckedMemRange::new(*self.registers.read_registers.fp, len as usize)?;
```
Voxelot, line 390: [nit] this range includes more than just code

```suggestion
        let frame_mem_range = CheckedMemRange::new(*self.registers.read_registers.fp, len as usize)?;
```

fuel-vm/src/interpreter/gas.rs
```diff
@@ -21,37 +27,102 @@ impl<S, Tx> Interpreter<S, Tx> {
...
 
     pub(crate) fn dependent_gas_charge(&mut self, gas_cost: DependentCost, arg: Word) -> Result<(), RuntimeError> {
-        if gas_cost.dep_per_unit == 0 {
-            self.gas_charge(gas_cost.base)
-        } else {
-            self.gas_charge(gas_cost.base.saturating_add(arg.saturating_div(gas_cost.dep_per_unit)))
-        }
+        let current_contract = self.contract_id();
+        let ReadRegisters { pc, ggas, cgas, is, .. } = split_registers(&mut self.registers).0;
+        let profiler = ProfileGas {
```
Voxelot, line 32: does this add overhead when the profiler is disabled?

fuel-vm/src/interpreter/gas/tests.rs
```diff
@@ -0,0 +1,108 @@
...
+)]
+#[test_case(
+    DepGasChargeInput{
+        input: GasChargeInput{cgas: 1, ggas: 1, gas: 1},
+        gas_cost: DependentCost{base: 1, dep_per_unit: 2}
+    } => Ok(GasChargeOutput{ cgas: 0, ggas: 0}); "just base with gas"
+)]
+#[test_case(
+    DepGasChargeInput{
+        input: GasChargeInput{cgas: 3, ggas: 3, gas: 8},
```
Voxelot, line 73: [nit] Seems strange to call the dependent input `gas`, when the gas usage is derived from this value.

```suggestion
        input: GasChargeInput{cgas: 3, ggas: 3, dependent: 8},
```

fuel-vm/src/interpreter/blockchain/test.rs
xgreenx: You are right=) I didn't know this

image

In most cases, this information still is not very actual, so I prefer an alias. But if it gives some benefits to you during development then it's up to you =)

fuel-vm/src/interpreter/gas/tests.rs
freesig: It causes panics when they are errors without the flag

fuel-vm/src/constraints.rs
freesig: I just made an issue for this because I have a trait that basically does this in my PR that removes unsafe and this should probably be handled in that pr or a follow up to it.

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+}
+
+impl<const INDEX: u8> Deref for Reg<'_, INDEX> {
+    type Target = Word;
+    fn deref(&self) -> &Self::Target {
+        self.0
+    }
+}
+impl<const INDEX: u8> Deref for RegMut<'_, INDEX> {
```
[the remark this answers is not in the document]
freesig: Can we make rustfmt do this??

fuel-vm/Cargo.toml
```diff
@@ -44,7 +44,7 @@ serde_json = "1.0"
 test-case = "2.2"
 
 [features]
-default = ["optimized"]
+default = []
```
freesig, line 47: Just flagging this as it's a significant change

fuel-vm/src/interpreter/blockchain/test.rs
freesig: What about:
```rust
pub type Memory<const SIZE: usize> = Box<[u8; SIZE]>;

fn foo() -> Memory<VM_MEMORY_SIZE> {}
```
The we still see the size in the type but hide the other details?

xgreenx: Works for me=) [nit] `VM_MEMORY_SIZE` seems soooooo long for me=D

fuel-vm/src/constraints/reg_key.rs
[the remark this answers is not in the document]
xgreenx: Nope=( https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#blank_lines_lower_bound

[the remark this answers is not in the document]
xgreenx: But `reorder_impl_items` adds new lines for `impl` sections=)

fuel-vm/src/consts.rs
```diff
@@ -7,6 +7,12 @@ use std::mem;
 /// Register count for checking constraints
 pub const VM_REGISTER_COUNT: usize = 64;
 
+/// The number of readable registers.
+pub const VM_REGISTER_READ_COUNT: usize = 16;
```
Voxelot, line 11:
```suggestion
pub const VM_REGISTER_SYSTEM_COUNT: usize = 16;
```

fuel-vm/src/consts.rs
```diff
@@ -7,6 +7,12 @@ use std::mem;
...
+
+/// The number of writable registers.
+pub const VM_REGISTER_WRITE_COUNT: usize = VM_REGISTER_COUNT - VM_REGISTER_READ_COUNT;
```
Voxelot, line 14:
```suggestion
pub const VM_REGISTER_PROGRAM_COUNT: usize = VM_REGISTER_COUNT - VM_REGISTER_SYSTEM_COUNT;
```

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -0,0 +1,312 @@
...
+
+impl<'a> From<ProgramRegisters<'a>> for ProgramRegistersRef<'a> {
+    fn from(value: ProgramRegisters<'a>) -> Self {
+        Self(value.0)
+    }
+}
+
+impl<'a> From<&SystemRegistersRef<'a>> for [Word; VM_REGISTER_READ_COUNT] {
+    fn from(value: &SystemRegistersRef<'a>) -> Self {
+        [
```
xgreenx, line 256: Could you use `let SystemRegistersRef { and your variables here } = value;` please? It will help not to miss new registers=)

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -0,0 +1,312 @@
...
+fn can_split() {
+    let mut reg: [Word; VM_REGISTER_COUNT] = std::iter::successors(Some(0), |x| Some(x + 1))
+        .take(VM_REGISTER_COUNT)
+        .collect::<Vec<_>>()
+        .try_into()
+        .unwrap();
+    let expect = reg;
+
+    let (r, w) = split_registers(&mut reg);
+
```
xgreenx, line 287: Could you use `let SystemRegistersRef { and your variables here } = r;` please? It will help not to miss new registers=)

fuel-vm/src/interpreter/balances.rs
xgreenx: Do we need to create an issue to track it?

fuel-vm/src/interpreter/flow.rs
```diff
@@ -147,138 +159,392 @@ where
...
 
-        let call = Call::try_from(&self.memory[a as usize..])?;
-        let asset_id =
-            AssetId::try_from(&self.memory[c as usize..cx as usize]).expect("Unreachable! Checked memory range");
+    /// Prepare a call instruction for execution
+    pub fn prepare_call(&mut self, ra: RegId, rb: RegId, rc: RegId, rd: RegId) -> Result<(), RuntimeError> {
+        const M: &str = "the provided id is not a valid register";
 
-        let mut frame = self.call_frame(call, asset_id)?;
+        let a = self
```
xgreenx, line 204: Maybe we can put it into inner closure or a method not to duplicate the same code?

fuel-vm/src/interpreter/flow.rs
xgreenx: I agree, it would be more in "rust wat"

fuel-vm/src/interpreter/balances.rs
freesig: It's already fixed in the next PR

fuel-vm/src/interpreter/blockchain/test.rs
freesig: `MEM_SIZE`?

fuel-vm/src/interpreter/flow.rs
freesig: What kind of error are you thinking?

fuel-vm/src/interpreter/flow.rs
freesig: This isn't from this PR

fuel-vm/src/interpreter/flow.rs
freesig: Yeh but it's a pain and it's not really that exspensive to copy `ContractId`

fuel-vm/src/interpreter/flow.rs
xgreenx: Like `ContractSizeMismatch`

fuel-vm/src/interpreter/gas.rs
freesig: What should I do instead though?

fuel-vm/src/storage/interpreter.rs
freesig: Probably but this call doesn't use that so it's maybe good for a follow up?

fuel-vm/src/constraints/reg_key.rs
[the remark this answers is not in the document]
freesig: Dam

fuel-vm/src/interpreter/gas.rs
xgreenx: `ProfileGas` can work with generic instead of `dyn`, and in this case, the compiler will optimize it. 

Also, you don't need at all to have a generic and a trait. You can have an alias like `type ProfileCapture = Profile;` when it is enabled and `type ProfileCapture = ProfileWithEmptyMethods;` when not, and use this alias in all places. And it seems you already do that, why not use an alias then without `dyn ProfileCapture`?

xgreenx: Done in [this commit](https://github.com/FuelLabs/fuel-vm/pull/353/commits/bb4d87f04d3230362780ec57edce3facfc7311f6)

fuel-vm/src/storage/interpreter.rs
xgreenx: Yeah, let's do that in follow up, because we can apply the same ideas for for "contracts" methods

fuel-vm/src/interpreter/gas/tests.rs
xgreenx: Global gas can't be less than context gas. I added the handling of this case with the corresponding error.

xgreenx (approved):
Looks good to me. Waiting for approve from @Voxelot

PR #354, opened 2023-02-16
Make all op functions unit testable

freesig:
closes #353 
This PR removes self from all op functions so they can be unit tested. The actual unit tests are in a follow up PR.


fuel-vm/tests/backtrace.rs
```diff
@@ -50,7 +50,7 @@ fn backtrace() {
 
     #[rustfmt::skip]
     let mut function_call = vec![
-        op::movi(0x10,  (contract_undefined.as_ref().len() + WORD_SIZE * 2) as Immediate18),
+        op::movi(0x10,  (contract_undefined.as_ref().len() + WORD_SIZE * 2 + 1) as Immediate18),
```
mitchmindtree, line 53: Are these changes the result of those byte offsets you mentioned in our chat?

mitchmindtree (commented):
Niccce. I haven't looked too closely at each of the individual operations assuming the behaviour is generally the same, but the general direction looks great :+1: Makes the inputs/outputs of each operation very clear! I've left a couple suggestions, some of which we already covered in our sync but thought I'd leave them in anyways.

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -80,35 +115,42 @@ macro_rules! impl_keys {
...
-    ERR, err
-    GGAS, ggas
-    CGAS, cgas
-    BAL, bal
-    IS, is
-    RET, ret
-    RETL, retl
-    FLAG, flag
+    ZERO, zero, zero_mut
+    ONE, one, one_mut
```
Voxelot, line 139: These registers should be immutable, should we even allow a `zero_mut` and `one_mut` to exist?

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -19,6 +25,35 @@ pub struct RegMut<'r, const INDEX: u8>(&'r mut Word);
...
+    /// of the writable registers.
+    pub fn new(k: impl Into<usize>) -> Result<Self, RuntimeError> {
+        let k = k.into();
+        is_register_writable(&k)?;
+        Ok(Self(k))
+    }
+
+    fn translate(self) -> usize {
+        self.0 - VM_REGISTER_READ_COUNT
+    }
```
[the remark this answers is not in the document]
Voxelot: It's because while there are 64 registers in total, the read registers are tracked in a separate array. It seems like this translate method converts from a global register index to a local register index within the set of writable registers.

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -246,6 +288,28 @@ pub(crate) fn copy_registers(
     ]
 }
 
+impl<'r> WriteRegisters<'r> {
+    pub fn split(&mut self, a: WriteRegKey, b: WriteRegKey) -> Option<(&mut Word, &mut Word)> {
```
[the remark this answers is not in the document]
Voxelot: Agree that split is a bit vague at first glance. Maybe something like `split_2_mut` or `double_mut` would be more clear?

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -304,39 +368,22 @@ impl<'a> From<WriteRegisters<'a>> for WriteRegistersRef<'a> {
...
-    assert_eq!(r.is, RegMut::<IS>(&mut (IS as u64)));
-    assert_eq!(r.ret, RegMut::<RET>(&mut (RET as u64)));
-    assert_eq!(r.retl, RegMut::<RETL>(&mut (RETL as u64)));
-    assert_eq!(r.flag, RegMut::<FLAG>(&mut (FLAG as u64)));
-
-    for i in 0..VM_REGISTER_WRITE_COUNT {
-        assert_eq!(w.0[i], i as u64 + 16);
+impl TryFrom<RegisterId> for WriteRegKey {
+    type Error = RuntimeError;
+    fn try_from(ra: RegisterId) -> Result<Self, Self::Error> {
```
Voxelot, line 373: [nit] does ra mean anything special here? Usually it refers to the first register in an opcode but that doesn't seem relevant in this case.

fuel-vm/src/interpreter.rs
```diff
@@ -152,6 +145,14 @@ impl<S, Tx> Interpreter<S, Tx> {
     }
 }
 
+pub(crate) fn is_wrapping(flag: Reg<FLAG>) -> bool {
+    *flag & 0x02 == 0x02
```
Voxelot, line 149: maybe we should use constants for these flag values? Or bring back the bitfield library?

fuel-vm/src/interpreter/alu.rs
```diff
@@ -13,79 +13,161 @@ where
...
-        // set the OF register to high bits of the u128 result
-        self.registers[RegId::OF] = (result >> 64) as u64;
-        self.registers[RegId::ERR] = 0;
-
-        // set the return value to the low bits of the u128 result
-        self.registers[ra] = (result & Word::MAX as u128) as u64;
-
-        self.inc_pc()
+        let (ReadRegisters { flag, of, err, pc, .. }, mut w) = split_registers(&mut self.registers);
+        let dest = &mut w[ra.try_into()?];
```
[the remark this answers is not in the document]
Voxelot: It'll be difficult to land that while these big PRs are in flight, unless we want to make a PR that targets this one directly.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -23,21 +33,247 @@ where
...
+        };
+        input.code_copy(a, b, c, d)
+    }
+
+    pub(crate) fn block_hash(&mut self, a: Word, b: Word) -> Result<(), RuntimeError> {
+        let owner = self.ownership_registers();
+        block_hash(&self.storage, &mut self.memory, owner, self.registers.pc_mut(), a, b)
+    }
+
+    pub(crate) fn set_block_height(&mut self, ra: RegisterId) -> Result<(), RuntimeError> {
```
Voxelot, line 93:
```suggestion
    pub(crate) fn get_block_height(&mut self, ra: RegisterId) -> Result<(), RuntimeError> {
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -173,79 +446,117 @@ where
...
 
-        self.inc_pc()
-    }
+    inc_pc(pc)
+}
 
-    pub(crate) fn block_proposer(&mut self, a: Word) -> Result<(), RuntimeError> {
-        self.coinbase()
-            .and_then(|data| self.try_mem_write(a as usize, data.as_ref()))?;
+pub(crate) fn block_proposer<S: InterpreterStorage>(
```
Voxelot, line 491:
```suggestion
pub(crate) fn coinbase<S: InterpreterStorage>(
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -257,161 +568,166 @@ where
...
 
-        self.block_height()
-            .and_then(|c| (b <= c as Word).then_some(()).ok_or(PanicReason::TransactionValidity))?;
+    inc_pc(pc)
+}
 
-        let b = u32::try_from(b).map_err(|_| PanicReason::ArithmeticOverflow)?;
+pub(crate) fn timestamp(
+    storage: &impl InterpreterStorage,
+    block_height: u32,
```
Voxelot, line 670: While passing block_height instead of fetching it from storage here does make unit testing of b > block_height easier, it seems somewhat strange to me to have several layers of functions plucking one thing out of storage at a time before calling the next function. 

Personally I'd prefer to keep the primary domain logic contained within in these selfless functions as much as possible, and limit the logic in the &self methods to the bare minimum needed to pass references to registers, memory & storage from the Interpreter instance. The more fragmented these functions become, the harder it is to analyze and also potentially leave more room for error in future refactors if anything can be passed in for block height.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -257,161 +568,166 @@ where
...
-        let ax = checked_add_word(a, Address::LEN as Word)?;
-        let bx = checked_add_word(ax, b)?;
+    inc_pc(pc)
+}
 
-        //TODO check on b vs MEM_MAX_ACCESS_SIZE is looser than msg length check
-        if b > self.params.max_message_data_length || b > MEM_MAX_ACCESS_SIZE || bx > VM_MAX_RAM {
-            return Err(PanicReason::MemoryOverflow.into());
-        }
+struct MessageOutputInput<'vm, Tx> {
```
Voxelot, line 686: I find it somewhat counter-intuitive to have "Input" objects that also do the real execution. As I wouldn't expect an input struct to perform complicated tasks. Maybe something like `Operation` instead of `Input` would be easier to follow?

```suggestion
struct MessageOutputOperation<'vm, Tx> {
```

fuel-vm/src/interpreter.rs
xgreenx: I guess https://github.com/FuelLabs/fuel-vm/pull/363 will cause a merge conflict that should be resolved in favor of @Dentosal change=)

fuel-vm/tests/backtrace.rs
freesig: It's because of the hp not being set correctly.

fuel-vm/src/constraints/reg_key.rs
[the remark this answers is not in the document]
mitchmindtree: Yeah I'd probably go with `get_2_mut` or `get_twice_mut`

fuel-vm/src/constraints/reg_key.rs
```diff
@@ -213,12 +215,12 @@ pub(crate) fn split_registers(registers: &mut [Word; VM_REGISTER_COUNT]) -> (Rea
...
     };
-    (r, WriteRegisters(rest))
+    (r, ProgramRegisters(rest))
 }
 
 pub(crate) fn copy_registers(
-    read_registers: &ReadRegistersRef<'_>,
-    write_registers: &WriteRegistersRef<'_>,
+    read_registers: &SystemRegistersRef<'_>,
+    write_registers: &ProgramRegistersRef<'_>,
```
mitchmindtree, line 229: Nice, I love these name changes! Much clearer.

fuel-vm/src/interpreter/blockchain.rs
mitchmindtree: Another common pattern is to use `Context`/`Ctx` when providing the context for some operation, e.g.
```suggestion
struct MessageOutputCtx<'vm, Tx> {
```

fuel-vm/src/constraints/reg_key.rs
mitchmindtree: Might also want to update the binding names too i.e. something like
```suggestion
    sys_regs: &SystemRegistersRef<'_>,
    prog_regs: &ProgramRegistersRef<'_>,
```

fuel-vm/src/interpreter/alu.rs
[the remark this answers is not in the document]
freesig: s is actually checking that the key is above the system registers and below the end of all registers. I don't think `RegKey` does that?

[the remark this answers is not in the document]
Voxelot: This is a really nice way to prevent invalid register ids from being written to! We had this bug before and it was discovered via proptesting, but having a way to guarantee the correctness via the type system is much better.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -257,161 +569,166 @@ where
...
-        let contract_id = *self.internal_contract()?;
-        let input = StateReadQWord::new(a, c, d, self.ownership_registers())?;
-        let Self {
-            ref storage,
-            ref mut memory,
-            ref mut registers,
-            ..
-        } = self;
+pub(crate) fn state_write_word<S: InterpreterStorage>(
+    StateWordCtx {
```
Voxelot, line 632: didn't realize it was possible to destructure inside fn args, this is nice way to reduce boilerplate

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -257,161 +569,166 @@ where
...
+pub(crate) fn timestamp(
+    storage: &impl InterpreterStorage,
+    block_height: u32,
+    pc: RegMut<PC>,
+    result: &mut Word,
+    b: Word,
+) -> Result<(), RuntimeError> {
+    (b <= block_height as Word)
+        .then_some(())
+        .ok_or(PanicReason::TransactionValidity)?;
```
Voxelot, line 678: unrelated to this PR, but this panic reason doesn't make much sense to me. Since the tx could be completely valid according to the spec and still encounter this error if a contract it called passed in the wrong block height here.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -420,26 +737,46 @@ where
 
         // validations passed, perform the mutations
 
-        self.base_asset_balance_sub(amount)?;
+        base_asset_balance_sub(self.balances, self.memory, self.amount_coins_to_send)?;
```
Voxelot, line 740: this should subtract from a contract balance if the context is internal. Although this should be done in a separate PR.

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -122,20 +122,7 @@ where
...
-                        } else if l < 2 {
-                            (l, false)
-                        } else {
-                            (0, true)
-                        }
-                    },
-                    r!(b),
-                    r!(c),
-                )?;
+                self.alu_boolean_overflow(a.into(), alu::exp, r!(b), r!(c))?;
```
Voxelot, line 125: Looks like @Dentosal's bugfix was lost in a merge?

freesig: Isn't it in the alu::exp?

Voxelot: ah you're right, my bad

Voxelot (approved):

PR #356, opened 2023-02-16
fix: Encode SMT error keys

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-core/issues/527

When traversing a Merkle tree (during SMT update or delete, during BMT root calculation, etc.) we must access the left or right child of each node along the path. Accessing children is liable to cause a `ChildError` (e.g., `ChildNotFound`, `NodeIsLeaf`), and we must be able to report the key of the invalid child. Currently this is not possible for all keys because some key types, like `[u8; 32]` cannot be used in the error message.

This PR adds a wrapper for a key type that can be used in the `ChildError` error message called `ChildErrorKey`. `ChildErrorKey` implements different display strategies depending on the type of key. Specifically:
- A key of type `[u8; 8]` is displayed as a `u64` (big endian)
- A key of type `[u8; 32]` is displayed as a hex encoded string of the key value

This allows fuel-core to use fallible SMT functions, like `update`, and to generate an error object with the expected key when an error happens.


fuel-merkle/src/common/node.rs
```diff
@@ -28,17 +30,37 @@ pub type ChildResult<T: ParentNode> = Result<T, ChildError<T::Key, T::Error>>;
...
     #[cfg_attr(feature = "std", error("Node is a leaf with no children"))]
     NodeIsLeaf,
     #[cfg_attr(feature = "std", error(transparent))]
     Error(E),
 }
 
-impl<Key, E> From<E> for ChildError<Key, E> {
+impl<Key, E> From<E> for ChildError<Key, E>
+where
+    for<'a> ChildErrorKey<&'a Key>: fmt::Display,
```
xgreenx, line 47:
```suggestion
    ChildErrorKey<Key>: fmt::Display,
```

I guess it should work too and we can simplify it for all implementations

bvrooman: If I understand correctly, you're suggesting to use an owned type `Key` rather than `&Key`. Then, we change the constraints from:
 ```
 for<'a> ChildErrorKey<&'a Key>: fmt::Display,
 ```
 to just:
 ```
 ChildErrorKey<Key>: fmt::Display,
 ```

That could work, but with the caveat that we would need adjust `ChildError` to something like this:
```
pub enum ChildError<Key, E>
where
    Key: Copy,
    ChildErrorKey<&'a Key>: fmt::Display,
{
    #[cfg_attr(feature = "std", error("Child with key {} was not found in storage", ChildErrorKey(*.0)))]
    ChildNotFound(Key),
    ...
}
```

Note the `Copy` constraint and the dereference `ChildErrorKey(*.0)`. This is because `Key` is accessed (`.0`) as a reference in the error message. We can do that, but I figured this way we avoid the extra copy.

xgreenx: Yeah, I missed that `.0` is a reference=)

But it can be done in a better way via the trait. [Here](https://github.com/bvrooman/fuel-vm/pull/2) is an example.

But why do we use aliases instead of types from the `fuel-types`?
image

Our types already implement `fmt::Display` and you don't need to use custom implementation.

BTW, could you next time create a branch in the `FuelLabs` org?=) In this case, it would be simpler to suggest changes=D

bvrooman: Thanks for the suggestion. I like the simpler approach. 

I would like to use the `fuel-types` types, but I will do that in a separate PR. As Tommi [mentioned](https://github.com/FuelLabs/fuel-vm/pull/341#pullrequestreview-1284781805), going between different owned types (storage key <-> hash) introduces extra copies. I think I should implement this in a separate PR so we can discuss the impact more precisely.

Voxelot (approved):

Voxelot:
lgtm - waiting for @xgreenx to also approve

xgreenx (approved):

PR #358, opened 2023-02-20
Unit tests for op functions

freesig:
Adds at least one unit test per op functions but some have full coverage.


freesig:
Now have at least one unit test per op function.

fuel-vm/src/interpreter/alu/tests.rs
```diff
@@ -0,0 +1,189 @@
...
+    b: u128,
+    c: u128,
+) -> Result<(Word, CommonInput), RuntimeError> {
+    let common = AluCommonReg {
+        of: RegMut::new(&mut of),
+        err: RegMut::new(&mut err),
+        pc: RegMut::new(&mut pc),
+    };
+    let mut dest = 0;
+    alu_capture_overflow(&mut dest, Reg::new(&flag), common, u128::overflowing_add, b, c)
```
Voxelot, line 54: I feel like we should consider wrapping add more fully into it's own fn, as this is basically a copy/paste of what add does rather than testing the same exact add logic used by the interpreter..

freesig: Yeh but this is true for basically all the ops in alu. I guess the point is to more unit test the actual function `alu_capture_overflow`

Voxelot (approved):

PR #359, opened 2023-02-21
Benchmark prepare call and split registers

freesig:


freesig:
Too many conflicts to bother merging this. We know it was fast enough.

PR #370, opened 2023-03-01
CALL instruction recursion limit

Dentosal:
Since we currently implement CALL using recursion, it can overflow the process stack before exceeding the gas limit. Currently that occurs around 95 nested calls, so to be sure I've set the limit to 64 for now. It's hardcoded as a part of the interpreter, as changing that value in a config doesn't currently make much sense.

In the future it would be nice to reduce the stack allocations to a level where the limit of 256 could be reached. Even better would be to not use non-optimized recursion in the first place, and keep the state in a stack instead. I will be attempting a refactor for this in the future. When such refactor is complete, the recursion limit could controlled similarly to the gas limits.

This is a breaking change, although I'd assume that nobody is relying on the behavior this PR disallows.

### DIFF
--- a/fuel-vm/tests/flow.rs
+++ b/fuel-vm/tests/flow.rs
@@ -492,7 +492,7 @@ fn nested_call_limit() {
         maturity,
         script,
         script_data,
-        vec![input.clone()],
+        vec![input],
         vec![output],
         vec![],
     )
--- a/fuel-vm/tests/flow.rs
+++ b/fuel-vm/tests/flow.rs
@@ -461,30 +461,13 @@ fn nested_call_limit() {
     let input = Input::contract(rng.gen(), rng.gen(), rng.gen(), rng.gen(), contract);
     let output = Output::contract(0, rng.gen(), rng.gen());
 
-    let mut script_ops = vec![
-        op::movi(0x10, 0),
+    let script: Vec<u8> = vec![
+        op::gtf_args(0x10, RegId::ZERO, GTFArgs::ScriptData),
         op::call(0x10, RegId::ZERO, 0x10, RegId::CGAS),
         op::ret(RegId::ONE),
-    ];
-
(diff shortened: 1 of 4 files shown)


fuel-vm/tests/flow.rs
```diff
@@ -416,6 +416,110 @@ fn revert_from_call_immediately_ends_execution() {
...
+    let input = Input::contract(rng.gen(), rng.gen(), rng.gen(), rng.gen(), contract);
+    let output = Output::contract(0, rng.gen(), rng.gen());
+
+    let mut script_ops = vec![
+        op::movi(0x10, 0),
+        op::call(0x10, RegId::ZERO, 0x10, RegId::CGAS),
+        op::ret(RegId::ONE),
+    ];
+
+    let script: Vec<u8> = script_ops.clone().into_iter().collect();
```
Voxelot, line 470: FYI, you can use the GTF opcode now to get the script data length dynamically instead of needing to create 2 versions of the transaction.

fuel-vm/tests/flow.rs
```diff
@@ -416,6 +416,110 @@ fn revert_from_call_immediately_ends_execution() {
...
+        script,
+        vec![],
+        vec![input.clone()],
+        vec![output],
+        vec![],
+    )
+    .into_checked(height, &params, client.gas_costs())
+    .expect("failed to generate checked tx");
+
+    let script_data_offset = client.tx_offset() + tx.transaction().script_data_offset();
```
Voxelot, line 484: FYI, you can use the GTF opcode now to get the script data length dynamically instead of needing to create 2 versions of the transaction.

Dentosal: I simply copied another case and modified it to work. I'll update this.

Voxelot (approved):

Issue #348, opened 2023-02-13
Remove all unsafe code


freesig:
# Problem
The current implementation of the fuel VM (and dependencies in the same repo) contains unsafe code which could potentially lead to security vulnerabilities and undermine the reliability and integrity of fuel core.

# Solution
Remove all unsafe in favor of safe rust code for all crates in this repo.

Removing unsafe code from the vm will provide significant benefits in terms of security, reliability, maintainability, and confidence in the system. The small performance improvements provided by unsafe code are not worth the risks and trade-offs associated with it.

If a code path shows up as particularly slow (in a real benchmark) and the only option is to use unsafe code we need to develop safe apis on top so the unsafe code is contained. We also need to very thoroughly test the unsafe code using something like [loom](https://github.com/tokio-rs/loom) or other tools to make sure the correctness constraints are upheld in the unsafe code. It should not be up to the API consumer to uphold the constraints as this scatters potentially subtle bugs all throughout out code base.

PR #371, opened 2023-03-03
Removing almost all unsafe code

freesig:
It is rebased https://github.com/FuelLabs/fuel-vm/pull/365

Closes https://github.com/FuelLabs/fuel-vm/issues/348


xgreenx (approved):
The `fuel-tx` part looks good to me.
I didn't try to dive deeply into the `fuel-vm` part. I fixed issues that noticed there, but someone also should check it=)

fuel-asm/src/panic_reason.rs
```diff
@@ -566,3 +165,22 @@ impl From<core::array::TryFromSliceError> for PanicReason {
...
+    fn test_u8_panic_reason_round_trip() {
+        for i in 0..0x25 {
+            let reason = PanicReason::from(i);
+            let i2 = reason as u8;
+            assert_eq!(i, i2);
+        }
+        for i in 0x25..=255 {
+            let reason = PanicReason::from(i);
+            let i2 = reason as u8;
+            assert_eq!(0x25, i2);
```
Voxelot, line 183: should this be asserting that the panic reason is `UnknownPanicReason`? Otherwise the test is basically the same besides the ranges.

xgreenx: Yes, `0x25` is `UnknownPanicReason` and it checks that any number more than `0x25` will be `UnknownPanicReason`.

Dentosal (approved):
One small nit: This uses both `warn` and `deny` lints for `unsafe_code`. Is there a reason to not just use one of them all the time? Also, if some crate doesn't contain any unsafe at all anymore, we could use `forbid` lint instead, as that cannot be `allow`'d afterwards, so it's easy to check which crates are completely safe Rust.

LGTM in any case, this is a big improvement.

PR #373, opened 2023-03-06
Retryable messages `fuel-vm` part

xgreenx:
`fuel-vm` part of the https://github.com/FuelLabs/fuel-core/issues/946

- Removed the `message_id` field from the message.
- Removed `Output::Message`.
- Re-worked messages to be spendable as a common coin and retryable.
- Added rules regarding `sum_inputs` and `sum_data_messages`. Updated the logic of the VM to only use `SumDataMessages` for `RuntimeBalances`.


fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -667,6 +666,11 @@ pub(crate) fn timestamp(
     inc_pc(pc)
 }
 
+// TODO: Register b is the address of the begin of the data and register c
```
xgreenx, line 669: It is a bug and we need to fix it in the follow-up PR

fuel-tx/src/transaction/types/input.rs
```diff
@@ -63,14 +63,16 @@ impl AsField<Vec<u8>> for Vec<u8> {
...
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Input {
     CoinSigned(CoinSigned),
     CoinPredicate(CoinPredicate),
     Contract(Contract),
-    MessageSigned(MessageSigned),
-    MessagePredicate(MessagePredicate),
+    DepositCoinSigned(DepositCoinSigned),
+    DepositCoinPredicate(DepositCoinPredicate),
+    MetadataSigned(MetadataSigned),
```
Voxelot, line 74: Why are we changing the name from message to metadata? We don't have anything in the specs referring to metadata inputs so this is confusing.

Voxelot: changed to MessageCoin_ and MessageData_

fuel-vm/src/interpreter/diff/tests.rs
```diff
@@ -204,11 +202,12 @@ fn reset_vm_memory() {
 
 #[test]
 fn reset_vm_txns() {
```
Voxelot, line 204: this test should clarify what is being tested about reset

Voxelot (approved):

PR #376, opened 2023-03-07
Make call instruction implementation non-recursive

Dentosal:
After this, we should be able to freely increase or remove `VM_MAX_NESTED_CALLS`

### DIFF
--- a/fuel-vm/src/interpreter/executors/main.rs
+++ b/fuel-vm/src/interpreter/executors/main.rs
@@ -2,7 +2,7 @@ use crate::checked_transaction::{Checked, IntoChecked};
 use crate::consts::*;
 use crate::context::Context;
 use crate::crypto;
-use crate::error::{Bug, BugId, BugVariant, InterpreterError, PredicateVerificationFailed, RuntimeError};
+use crate::error::{Bug, BugId, BugVariant, InterpreterError, PredicateVerificationFailed};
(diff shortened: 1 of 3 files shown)


fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -92,47 +92,6 @@ impl<T> Interpreter<PredicateStorage, T> {
...
-    Tx: ExecutableTransaction,
-{
-    pub(crate) fn run_call(&mut self) -> Result<ProgramState, RuntimeError> {
-        loop {
-            if self.registers[RegId::PC] >= VM_MAX_RAM {
-                return Err(PanicReason::MemoryOverflow.into());
-            }
-
-            let state = self.execute().map_err(|e| {
-                e.panic_reason()
```
bvrooman, line 107: Do you not need to carry this mapping over to the refactor?

Dentosal: No. The old code was wrapping the panic reason in a RuntimeError and then later undoing that (via implicit `?`). Now that we're combining `run_program` and `run_call`, the error can be propagated directly.

Voxelot (approved):

xgreenx (approved):

Voxelot (approved):

PR #381, opened 2023-03-13
Relative jump opcodes

Dentosal:
- [x] Integration tests
- [x] Fuel-specs PR: https://github.com/FuelLabs/fuel-specs/pull/470


fuel-vm/src/interpreter/flow.rs
```diff
@@ -200,38 +196,50 @@ pub(crate) fn revert(append: AppendReceipt, current_contract: Option<ContractId>
...
+        JumpMode::RelativeBackwards => pc
+            .checked_sub(
+                addr.saturating_add(1)
+                    .saturating_add(offset)
+                    .saturating_mul(Instruction::SIZE as Word),
+            )
+            .ok_or_else(|| RuntimeError::Recoverable(PanicReason::MemoryOverflow))?,
+    };
+
+    if addr > VM_MAX_RAM - 1 {
```
Voxelot, line 237: [nit]

```suggestion
    if addr >= VM_MAX_RAM {
```

Voxelot (dismissed):

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -303,31 +304,67 @@ where
...
                 self.gas_charge(self.gas_costs.jne)?;
                 let (a, b, c) = jne.unpack();
-                self.jump_not_equal(r!(a), r!(b), r!(c))?;
+                self.jump(r!(a) != r!(b), JumpMode::Absolute, r!(c), 0)?;
+            }
+
+            Instruction::JMPF(jmpf) => {
+                self.gas_charge(self.gas_costs.jmpf)?;
+                let (a, offset) = jmpf.unpack();
+                self.jump(true, JumpMode::RelativeForwards, r!(a), offset.into())?;
```
xgreenx, line 337: `offset` could be part of the `JumpMode`

Voxelot (approved):

fuel-vm/src/interpreter/executors/instruction.rs
Dentosal: So could `addr`, but all modes share it, so then we would create `struct JumpArgs { addr, offset, mode }`. I don't feel like this makes much sense.

xgreenx: But it will remove passing `0` for `Absolute` mode=) The `addr` is used in all modes, so we don't need to move it into 
`JumpMode`

Dentosal: I rewrote this to use a builder pattern, that seems cleaner

fuel-vm/src/interpreter/flow.rs
```diff
@@ -201,38 +197,50 @@ pub(crate) fn revert(append: AppendReceipt, current_contract: Option<ContractId>
...
-    if j > VM_MAX_RAM - 1 {
-        Err(PanicReason::MemoryOverflow.into())
-    } else {
-        *pc = j;
 
-        Ok(())
+    let addr = match mode {
+        JumpMode::Absolute => is.saturating_add(addr.saturating_add(offset).saturating_mul(Instruction::SIZE as Word)),
+        JumpMode::RelativeForwards => pc.saturating_add(
+            addr.saturating_add(1)
```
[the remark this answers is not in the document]
Dentosal: Added a comment explaining that zero-length relative jumps don't make any sense

Issue #97, opened 2022-03-24
set receipt root on serialized tx in memory


Voxelot:
In order to keep the serialized representation of the tx consistent with it's malleated state, set the `receipt_root` on the serialized form of the transaction as well. Keeping the serialized representation of the tx will allow it to be easily copied and reused for merkle hashing later on.

https://github.com/FuelLabs/fuel-vm/pull/93#discussion_r830499552

Voxelot:
related: #125

PR #383, opened 2023-03-14
Cache receipts in in-memory BMT during execution

### LINKED ISSUE #311: Caching of receipts merkle tree
Instead of computing the receipts root ephemerally from scratch each time we append a new receipt, we should consider using an in-memory merkle tree state that is cached on the VM instance. This would make appending receipts cheaper as the new root would computed from an existing tree instead of an ephemeral one.

This should be relatively straightforward except for the new state-diffing feature, which may justify writing a custom merkle tree in-memory backend to make diffing easier.

bvrooman:
Related issues:
- Closes #311 

Introduces `ReceiptsCtx` which ties together VM receipts and the corresponding Merkle root. The `ReceiptsCtx` exposes a simple interface that keeps the list of receipts and the binary Merkle tree in sync:
- `push`: push a receipt to the context and update the Merkle root
- `root`: get the current root

The `ReceiptsCtx` also supports inverting receipts from a VM state diff. Because BMTs do not natively support modifications (only appends), we recalculate the Merkle root from all leaves when inverting a receipt.


fuel-vm/src/interpreter/internal/message_tests.rs
```diff
@@ -2,6 +2,7 @@ use crate::interpreter::memory::Memory;
 
 use super::*;
 use fuel_tx::Create;
+use fuel_types::bytes::SerializableVec;
```
bvrooman, line 5: Now `SerializableVec` is no longer found in `super::*`, so I must explicitly add it here.

fuel-merkle/src/binary/merkle_tree.rs
```diff
@@ -182,6 +182,11 @@ where
         Ok((root, proof_set))
     }
 
+    pub fn reset(&mut self) {
```
Voxelot, line 185: does this need to alter storage at all?

fuel-vm/src/interpreter/receipts.rs
```diff
@@ -0,0 +1,170 @@
...
+        self.receipts_tree.reset();
+        self.receipts.clear();
+    }
+
+    pub fn root(&self) -> Bytes32 {
+        self.receipts_tree.root().into()
+    }
+
+    /// Get a mutable lock on this context
+    pub fn lock(&mut self) -> ReceiptsCtxMut {
```
Voxelot, line 30: Tried to think of another way this could be done, but this seems to strike the best balance between minimizing allocs and safety when rewinding diffs.

fuel-vm/src/tests/predicate.rs
```diff
@@ -236,9 +236,9 @@ fn gas_used_by_predicates_is_deducted_from_script_gas() {
 
     let mut client = MemoryClient::default();
     client.transact(tx_with_predicate);
-    let receipts_with_predicate = client.receipts().expect("Expected receipts").to_vec();
+    let receipts_with_predicate = client.receipts().expect("Expected receipts").as_ref().to_vec();
```
Voxelot, line 239: Unnecessary change

```suggestion
    let receipts_with_predicate = client.receipts().expect("Expected receipts").to_vec();
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -236,9 +236,9 @@ fn gas_used_by_predicates_is_deducted_from_script_gas() {
...
     client.transact(tx_without_predicate);
-    let receipts_without_predicate = client.receipts().expect("Expected receipts").to_vec();
+    let receipts_without_predicate = client.receipts().expect("Expected receipts").as_ref().to_vec();
```
Voxelot, line 241: Unnecessary change

```suggestion
    let receipts_without_predicate = client.receipts().expect("Expected receipts").to_vec();
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -270,7 +270,7 @@ fn gas_used_by_predicates_causes_out_of_gas_during_script() {
     let mut client = MemoryClient::default();
 
     client.transact(tx_without_predicate);
-    let receipts_without_predicate = client.receipts().expect("Expected receipts").to_vec();
+    let receipts_without_predicate = client.receipts().expect("Expected receipts").as_ref().to_vec();
```
Voxelot, line 273: Unnecessary change

```suggestion
    let receipts_without_predicate = client.receipts().expect("Expected receipts").to_vec();
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -301,7 +301,7 @@ fn gas_used_by_predicates_causes_out_of_gas_during_script() {
         .expect("Predicate check failed");
 
     client.transact(tx_with_predicate);
-    let receipts_with_predicate = client.receipts().expect("Expected receipts").to_vec();
+    let receipts_with_predicate = client.receipts().expect("Expected receipts").as_ref().to_vec();
```
Voxelot, line 304: Unnecessary change

```suggestion
    let receipts_with_predicate = client.receipts().expect("Expected receipts").to_vec();
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -336,7 +336,7 @@ fn gas_used_by_predicates_more_than_limit() {
     let mut client = MemoryClient::default();
 
     client.transact(tx_without_predicate);
-    let receipts_without_predicate = client.receipts().expect("Expected receipts").to_vec();
+    let receipts_without_predicate = client.receipts().expect("Expected receipts").as_ref().to_vec();
```
Voxelot, line 339: Unnecessary change

```suggestion
    let receipts_without_predicate = client.receipts().expect("Expected receipts").to_vec();
```

fuel-vm/src/tests/memory.rs
```diff
@@ -109,7 +109,7 @@ fn test_stack_and_heap_cannot_overlap(offset: u64, cause_error: bool) {
     let mut vm = Transactor::new(storage, Default::default(), Default::default());
     vm.transact(tx);
 
-    let mut receipts = vm.receipts().unwrap().to_vec();
+    let mut receipts = vm.receipts().unwrap().as_ref().to_vec();
```
Voxelot, line 112: Unnecessary change

```suggestion
    let mut receipts = vm.receipts().unwrap().to_vec();
```

fuel-vm/src/tests/memory.rs
```diff
@@ -164,7 +164,7 @@ fn dynamic_call_frame_ops() {
     let mut vm = Transactor::new(storage, Default::default(), Default::default());
     vm.transact(tx);
 
-    let receipts = vm.receipts().unwrap().to_vec();
+    let receipts = vm.receipts().unwrap().as_ref().to_vec();
```
Voxelot, line 167: Unnecessary change

```suggestion
    let receipts = vm.receipts().unwrap().to_vec();
```

fuel-vm/src/interpreter/receipts.rs
```diff
@@ -0,0 +1,170 @@
...
+        ReceiptsCtxMut::new(self)
+    }
+
+    /// Recalculates the Merkle root of the receipts from scratch. This should
+    /// only be used when the list of receipts has been mutated externally.
+    fn recalculate_root(&mut self) {
+        self.receipts_tree.reset();
+        let receipts = self.as_ref().clone();
```
Voxelot, line 38: maybe we should add a todo here about removing clone once `to_bytes()` no longer requires `&mut self`

fuel-vm/src/interpreter/receipts.rs
```diff
@@ -0,0 +1,170 @@
...
+        for mut receipt in receipts {
+            self.receipts_tree.push(receipt.to_bytes().as_slice())
+        }
+    }
+}
```
Voxelot, line 43: might be good to add a `pub fn len(&self) -> usize` method here to avoid all the extra `.as_ref().len()` casting.

fuel-vm/src/tests/outputs.rs
```diff
@@ -303,7 +303,10 @@ fn variable_output_set_by_external_transfer_out() {
             && asset_id == asset_id
     ));
 
-    assert!(receipts.iter().any(|r| matches!(r, Receipt::TransferOut { .. })));
+    assert!(receipts
+        .as_ref()
```
Voxelot, line 307:
```suggestion
```

fuel-vm/src/tests/outputs.rs
```diff
@@ -379,7 +382,10 @@ fn variable_output_not_set_by_external_transfer_out_on_revert() {
     ));
 
     // TransferOut receipt should not be present
-    assert!(!receipts.iter().any(|r| matches!(r, Receipt::TransferOut { .. })));
+    assert!(!receipts
+        .as_ref()
```
Voxelot, line 386:
```suggestion
```

fuel-vm/src/tests/outputs.rs
```diff
@@ -457,7 +463,10 @@ fn variable_output_set_by_internal_contract_transfer_out() {
     let output = Output::variable(owner, transfer_amount, asset_id);
 
     assert_eq!(output, outputs[0]);
-    assert!(receipts.iter().any(|r| matches!(r, Receipt::TransferOut { .. })));
+    assert!(receipts
+        .as_ref()
```
Voxelot, line 467:
```suggestion
```

fuel-vm/src/tests/outputs.rs
```diff
@@ -536,5 +545,8 @@ fn variable_output_not_increased_by_contract_transfer_out_on_revert() {
     ));
 
     // TransferOut receipt should not be present
-    assert!(!receipts.iter().any(|r| matches!(r, Receipt::TransferOut { .. })));
+    assert!(!receipts
+        .as_ref()
```
Voxelot, line 549:
```suggestion
```

fuel-vm/src/tests/test_helpers.rs
```diff
@@ -23,7 +23,7 @@ pub fn run_script(script: Vec<Instruction>) -> Vec<Receipt> {
         .into_checked(0, &ConsensusParameters::DEFAULT, client.gas_costs())
         .expect("failed to generate a checked tx");
     client.transact(tx);
-    client.receipts().expect("Expected receipts").to_vec()
+    client.receipts().expect("Expected receipts").as_ref().to_vec()
```
Voxelot, line 26:
```suggestion
    client.receipts().expect("Expected receipts").to_vec()
```

fuel-vm/src/util.rs
```diff
@@ -446,7 +446,7 @@ pub mod test_helpers {
     ) {
         let receipts = client.transact(checked_tx);
 
-        let panic_found = receipts.iter().any(|receipt| {
+        let panic_found = receipts.as_ref().iter().any(|receipt| {
```
Voxelot, line 449:
```suggestion
        let panic_found = receipts.iter().any(|receipt| {
```

Voxelot (commented):
some minor nits, dunno why clippy doesn't catch these

Voxelot (dismissed):

fuel-merkle/src/binary/merkle_tree.rs
bvrooman: It actually doesn't because all reads and writes are determined only by the `leaves_count` and `head`. 

However, it does leave the storage in a kind of "dirty" state. It's currently not possible with this interface, but theoretically, with future modifications, we could end up in a position where we request a node that should be empty, and the dirty state returns erroneous data from before the reset. We could introduce another storage trait, e.g. `StorageReset` that we can implement on databases or in-memory storage that allows for implementation specific reset strategies. Then, we can simply call `storage.reset()` here to make sure we have cleaned up entirely.

bvrooman:
@Voxelot I pushed a commit to address your other comments, but dismissed your review. Please have another look when you get a chance.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -239,11 +237,7 @@ where
...
 
-            let receipts_root = if self.receipts().is_empty() {
-                EMPTY_RECEIPTS_MERKLE_ROOT.into()
-            } else {
-                crypto::ephemeral_merkle_root(self.receipts().iter().map(|r| r.clone().to_bytes()))
-            };
+            let receipts_root = self.receipts.root();
 
             // TODO optimize
             if let Some(script) = self.tx.as_script_mut() {
```
xgreenx, line 243: Could you remove "TODO" below, please? It seems https://github.com/FuelLabs/fuel-vm/issues/97 is solved=)

xgreenx (approved):

Voxelot (approved):

PR #386, opened 2023-03-15
Predicate Estimation VM part

ghost:
[Predicate Estimation Spec](https://github.com/FuelLabs/fuel-specs/blob/master/src/vm/index.md#predicate-estimation)


Voxelot:
1. We determined that modifying transaction fields during `IntoChecked` could cause edge cases where CheckedTransactions could later be invalid when re-checked (due to GTF support of the predicate_gas_used fields)
2. Make a new estimate predicates API on transactions that internally converts it into a basic checked tx, estimates the predicate gas, and then mutates the predicate gas used fields on the original tx
3. Then the user can turn it into a checked transaction later, which will only verify the fields and not perform any mutations.
4. `TransactionFee::checked_from_values` should statically price the base tx cost based on the total sum of all `predicate_gas_used` fields on the inputs.

fuel-vm/src/checked_transaction/types.rs
```diff
@@ -180,4 +205,22 @@ pub mod script {
             Ok(Checked::basic(self, metadata))
         }
     }
+
+    // impl EstimatePredicates for Script {
```
Dentosal, line 209: Why is this commented out?

fuel-tx/src/transaction/types/create.rs
```diff
@@ -396,6 +396,11 @@ mod field {
                     .zip(input.predicate_len().map(bytes::padded_len_usize))
             })
         }
+
+        #[inline(always)]
+        fn inputs_predicate_gas_used_at(&self, idx: usize) -> Option<Word> {
```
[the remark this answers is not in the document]
freesig:
```suggestion
        fn inputs_predicate_gas_used_offset_at(&self, idx: usize) -> Option<Word> {
```
nit
```suggestion
        fn inputs_predicate_gas_used_at(&self, idx: usize) -> Option<Word> {
```

fuel-tx/src/transaction/types/create.rs
```diff
@@ -396,6 +396,11 @@ mod field {
...
+            self.inputs().get(idx).and_then(|input| input.predicate_gas_used())
```
freesig, line 402: Isn't this supposed to be the location in memory of the predicate? This is returning the value itself

fuel-tx/src/transaction/types/script.rs
```diff
@@ -371,6 +371,11 @@ mod field {
                     .zip(input.predicate_len().map(bytes::padded_len_usize))
             })
         }
+
+        #[inline(always)]
+        fn inputs_predicate_gas_used_at(&self, idx: usize) -> Option<Word> {
+            self.inputs().get(idx).and_then(|input| input.predicate_gas_used())
```
freesig, line 377: This is also returning the value instead of the offset.

fuel-tx/src/transaction.rs
```diff
@@ -500,6 +500,9 @@ pub mod field {
 
         /// Returns predicate's offset and length of the `Input` at `idx`, if any.
         fn inputs_predicate_offset_at(&self, idx: usize) -> Option<(usize, usize)>;
+
+        /// Returns predicate's offset and length of the `Input` at `idx`, if any.
+        fn inputs_predicate_gas_used_at(&self, idx: usize) -> Option<Word>;
```
freesig, line 505: Yeh this comment indicates that it's a memory offset not a value

fuel-vm/src/checked_transaction/builder.rs
```diff
@@ -8,7 +8,7 @@ use fuel_types::BlockHeight;
 /// Extension trait for [`fuel_tx::TransactionBuilder`] adding finalization methods
 pub trait TransactionBuilderExt<Tx>
 where
-    Tx: IntoChecked,
+    Tx: IntoChecked + Clone,
```
freesig, line 11: Is this Clone actually needed?

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,52 +44,113 @@ impl<T> Interpreter<PredicateStorage, T> {
...
                 return Err(PredicateVerificationFailed::False);
             }
 
-            remaining_gas = vm.registers[RegId::GGAS];
+            if vm.registers[RegId::GGAS] != 0 {
+                return Err(PredicateVerificationFailed::GasMismatch);
+            }
+            cumulative_gas_used = cumulative_gas_used
+                .checked_add(gas_used)
+                .expect("cumulative gas overflow");
```
freesig, line 88: We don't want expects in production code. This should return an error instead.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,52 +44,113 @@ impl<T> Interpreter<PredicateStorage, T> {
     /// This is not a valid entrypoint for debug calls. It will only return a `bool`, and not the
     /// VM state required to trace the execution steps.
     pub fn check_predicates<Tx>(
-        checked: Checked<Tx>,
+        checked: &Checked<Tx>,
```
freesig, line 47: It's usually best practice to take an owned value when it get's cloned below as it tells the user of this api that the value is used.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,52 +44,113 @@ impl<T> Interpreter<PredicateStorage, T> {
...
     {
         if !checked.transaction().check_predicate_owners(&params) {
             return Err(PredicateVerificationFailed::InvalidOwner);
         }
 
         let mut vm = Interpreter::with_storage(PredicateStorage::default(), params, gas_costs);
 
+        let mut cumulative_gas_used: Word = 0;
+
+        vm.init_predicate(checked.clone());
```
freesig, line 63: Was there a reason to move this before the following line?

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,52 +44,113 @@ impl<T> Interpreter<PredicateStorage, T> {
...
-
-        vm.init_predicate(checked);
-
         for predicate in predicates {
             // VM is cloned because the state should be reset for every predicate verification
             let mut vm = vm.clone();
 
-            vm.context = Context::Predicate { program: predicate };
-            vm.set_remaining_gas(remaining_gas);
+            let gas_used = predicate.gas_used();
```
freesig, line 75: Doesn't this still need to account for the total transaction gas limit?

freesig (changes requested):
Nice job. A few things to address. Will also continue my review

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,52 +44,113 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+    /// inputs.
+    ///
+    /// The storage provider is not used since contract opcodes are not allowed for predicates.
+    /// This way, its possible, for the sake of simplicity, it is possible to use
+    /// [unit](https://doc.rust-lang.org/core/primitive.unit.html) as storage provider.
+    ///
+    /// # Debug
+    ///
+    /// This is not a valid entrypoint for debug calls. It will only return a `bool`, and not the
+    /// VM state required to trace the execution steps.
```
freesig, line 105: This comment seems unrelated?

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,52 +44,113 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+    pub fn estimate_predicates<Tx>(
```
freesig, line 106: I would move this out of the `Interpreter` and make it a stadalone function as it doesn't take self anyway. Then it can have unit tests

freesig (changes requested):

fuel-tx/src/tests/offset.rs
```diff
@@ -93,6 +93,17 @@ fn common_parts_create_and_script<Tx: Buildable>(tx: &Tx, bytes: &[u8], cases: &
             assert_eq!(predicate_data, predicate_data_p);
         }
 
+        if let Some(predicate_gas_used) = i.input_predicate_data() {
+            cases.predicate_data_coin = cases.predicate_data_coin || i.is_coin() && !predicate_gas_used.is_empty();
+            cases.predicate_data_message =
+                cases.predicate_data_message || i.is_message() && !predicate_gas_used.is_empty();
+
+            let ofs = input_ofs + i.predicate_data_offset().expect("input contains predicate data");
```
freesig, line 101: This is pointing to the same offset as above. That seems wrong?

freesig (changes requested):

fuel-tx/src/transaction/types/create.rs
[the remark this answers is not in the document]
mitch-fuel: I removed this function completely, and refactored it to iterate over the inputs, from which predicate_gas_used can be directly accessed for all predicates.

fuel-vm/src/checked_transaction/builder.rs
mitch-fuel: removed

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: removed and returning an error now

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: fixed, reverted back to format before a mutable reference was used to both check and estimate predicates via a boolean passed into this function that was removed

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -62,38 +62,41 @@ impl<T> Interpreter<PredicateStorage, T> {
...
-            .collect();
+        for (idx, input) in checked.transaction().inputs().iter().enumerate() {
+            if let Some(predicate) = RuntimePredicate::from_tx(&params, checked.transaction(), idx) {
+                // VM is cloned because the state should be reset for every predicate verification
+                let mut vm = vm.clone();
 
-        for predicate in predicates {
-            // VM is cloned because the state should be reset for every predicate verification
-            let mut vm = vm.clone();
+                let gas_used = input.predicate_gas_used().unwrap();
```
freesig, line 70: Is this unwrap supposed to be here?

freesig (changes requested):

fuel-tx/src/tests/offset.rs
freesig: The function `input_predicate_data` does not return `predicate_gas_used`

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: given that the check_predicates also does not operate on self, do you still feel this should be refactored out?

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: modified to be relevant

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: I added a check to the total gas used, but we can't know the total gas used, only that the predicate gas used for a single predicate matches the input field. We can do a static check before execution to ensure that the total predicate gas used for all inputs is less than the transaction gas limit.

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: yes, this field is not present on all inputs so could be None

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: can you clarify what you are referencing?

fuel-tx/src/transaction.rs
mitch-fuel: removed the function and refactored

fuel-tx/src/transaction/types/script.rs
mitch-fuel: removed the function and refactored

fuel-tx/src/transaction/types/create.rs
mitch-fuel: removed the function and refactored

fuel-vm/src/interpreter/executors/main.rs
freesig: But this is production code, we don't want to panic the whole client because of this right?

fuel-vm/src/interpreter/executors/main.rs
freesig: Maybe not but it should at least have a unit test

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -48,46 +50,110 @@ impl<T> Interpreter<PredicateStorage, T> {
...
-            .collect();
+        let mut cumulative_gas_used: Word = 0;
 
-        // Since we reuse the vm objects otherwise, we need to keep the actual gas here
-        let tx_gas_limit = checked.transaction().limit();
-        let mut remaining_gas = tx_gas_limit;
+        vm.init_predicate(checked.clone());
 
-        vm.init_predicate(checked);
+        for (idx, input) in checked.transaction().inputs().iter().enumerate() {
```
freesig, line 65:
```suggestion
        for (idx, input) in checked.transaction().inputs().iter().enumerate().filter(|(_, input)| matches!(input, Input::CoinPredicate(_) | Input::MessageCoinPredicate(_) | Input::DataPredicate(_))  {
```

fuel-vm/src/interpreter/executors/main.rs
freesig: It was moved above another line but that line has since been removed.

fuel-vm/src/checked_transaction.rs
```diff
@@ -153,7 +155,7 @@ impl<Tx: IntoChecked> Borrow<Tx> for Checked<Tx> {
 /// Performs checks for a transaction
 pub trait IntoChecked: FormatValidityChecks + Sized {
     /// Metadata produced during the check.
-    type Metadata: Sized;
+    type CheckedMetadata: Sized + Clone;
```
freesig, line 158: this require clone now?

mitch-fuel: yes, otherwise we can't clone the checked transaction

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: fixed

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: there is now a specific test in predicate.rs that addresses a transaction that fails predicate checking before estimation and succeeds after estimation

fuel-tx/src/tests/offset.rs
mitch-fuel: this was mistakenly implemented as I added the predicate_gas_used field, it has been removed, I also noticed that the GTF was not fully implemented for predicate_gas_used but now it is

freesig (approved):
Nice work. All looks good to me now. Can @xgreenx you give this a quick look before it's merged as I'm reviewing with a bit of a jet lagged brain right now :)

xgreenx (changes requested):
I didn't finish review, it only a part=)

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -49,6 +51,8 @@ impl ConsensusParameters {
         max_storage_slots: 255,
         max_predicate_length: 1024 * 1024,
         max_predicate_data_length: 1024 * 1024,
+        //TODO: set this to a reasonable value
```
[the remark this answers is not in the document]
mitch-fuel: done

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -48,46 +50,121 @@ impl<T> Interpreter<PredicateStorage, T> {
...
         }
 
         let mut vm = Interpreter::with_storage(PredicateStorage::default(), params, gas_costs);
 
-        // Needed for now because checked is only freed once the value is collected into a Vec
-        #[allow(clippy::needless_collect)]
-        let predicates: Vec<_> = (0..checked.transaction().inputs().len())
-            .filter_map(|i| RuntimePredicate::from_tx(&params, checked.transaction(), i))
-            .collect();
+        let mut cumulative_gas_used: Word = 0;
```
xgreenx, line 61: It seems you don't need to change the old logic. You can return it back, and at the end of the loop add:

```rust
 if vm.registers[RegId::GGAS] != 0 {
     return Err(PredicateVerificationFailed::GasMismatch);
 }
 cumulative_gas_used = cumulative_gas_used
     .checked_add(gas_used)
     .ok_or_else(|| PredicateVerificationFailed::OutOfGas)?;
```

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -48,46 +50,121 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+                }
+                cumulative_gas_used = cumulative_gas_used
+                    .checked_add(gas_used)
+                    .ok_or_else(|| PredicateVerificationFailed::OutOfGas)?;
             }
+        }
 
-            remaining_gas = vm.registers[RegId::GGAS];
+        if cumulative_gas_used > checked.transaction().limit() {
+            return Err(PredicateVerificationFailed::OutOfGas);
```
xgreenx, line 101: I think we want to return another error here, more user-friendly

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -48,46 +50,121 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+    {
+        let mut vm = Interpreter::with_storage(PredicateStorage::default(), params, gas_costs);
+
+        let tx_gas_limit: u64 = transaction.limit();
+
+        let predicate_gas_limit: u64 = if tx_gas_limit > params.max_gas_per_predicate {
+            params.max_gas_per_predicate
+        } else {
+            tx_gas_limit
+        };
```
xgreenx, line 137: It is only estimation, not validation. We can use gas limit=)

```suggestion
        let predicate_gas_limit: u64 = params.max_gas_per_predicate;
```

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -48,46 +50,121 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+
+        let transaction_clone = transaction.clone();
```
xgreenx, line 139: You don't need to create a clone here. Instead of `iter()` you can iterate manually with `for i in 0..transaction.inputs().len()`. It will pass the borrow checker=)

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -48,46 +50,121 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+            if let Some(predicate) = RuntimePredicate::from_tx(&params, &transaction_clone, idx) {
+                vm.init_predicate_estimation(transaction_clone.clone(), balances.clone());
+                vm.context = Context::PredicateEstimation { program: predicate };
+                vm.set_gas(predicate_gas_limit);
+
+                if !matches!(vm.verify_predicate()?, ProgramState::Return(0x01)) {
+                    return Err(PredicateVerificationFailed::False);
+                }
+
+                let gas_used: u64 = tx_gas_limit
```
xgreenx, line 151:
```suggestion
                let gas_used: u64 = predicate_gas_limit
```

fuel-vm/src/interpreter/initialization.rs
```diff
@@ -70,15 +71,32 @@ where
...
-    <Tx as IntoChecked>::Metadata: CheckedMetadata,
+{
+    /// Initialize the VM for a predicate context
+    pub fn init_predicate_estimation(&mut self, transaction: Tx, balances: InitialBalances) -> bool {
+        self.context = Context::PredicateEstimation {
+            program: Default::default(),
+        };
+
+        let mut transaction = transaction;
+        transaction.prepare_init_predicate();
```
xgreenx, line 82: We need to zero the `Input.gas_used` field for estimation but not for verification, so you need to use another method here

fuel-vm/src/interpreter/initialization.rs
```diff
@@ -70,15 +71,32 @@ where
...
+    }
+}
+
+impl<S, Tx> Interpreter<S, Tx>
+where
+    Tx: ExecutableTransaction,
+    <Tx as IntoChecked>::CheckedMetadata: CheckedMetadata,
 {
     /// Initialize the VM for a predicate context
     pub fn init_predicate(&mut self, checked: Checked<Tx>) -> bool {
```
xgreenx, line 94:
```suggestion
    /// Initialize the VM for a verification predicate context
    pub fn init_predicate_verification(&mut self, checked: Checked<Tx>) -> bool {
```

fuel-vm/src/interpreter/metadata/tests.rs
```diff
@@ -42,7 +42,7 @@ fn test_get_transaction_field() {
     assert_eq!(result, 0);
 }
 
-#[test_case(Context::Predicate { program: Default::default() }, 2 => (); "can fetch inside predicate")]
+#[test_case(Context::PredicateVerification { program: Default::default() }, 2 => (); "can fetch inside predicate verification")]
```
xgreenx, line 45: I think we need to update all tests to check `PredicateEstimation` too

fuel-vm/src/tests/predicate.rs
```diff
@@ -105,10 +124,188 @@ fn get_verifying_predicate() {
...
+        predicate,
+        vec![],
+        rng.gen(),
+    );
+
+    builder.add_input(input);
+
+    let mut transaction = builder.finalize();
+
+    //unestimated transaction should fail as it's predicates are not estimated
```
xgreenx, line 281:
```suggestion
    // unestimated transaction should fail as it's predicates are not estimated
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -105,10 +124,188 @@ fn get_verifying_predicate() {
...
+    assert!(transaction
+        .clone()
+        .into_checked(Default::default(), &params, &GasCosts::default())
+        .is_err());
+
+    //create checked transaction to get access to balances from metadata
```
xgreenx, line 287:
```suggestion
    // create checked transaction to get access to balances from metadata
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -105,10 +124,188 @@ fn get_verifying_predicate() {
...
+
+    let metadata: &CheckedMetadata = unestimated_checked.metadata();
+
+    let balances: InitialBalances = InitialBalances {
+        non_retryable: metadata.clone().non_retryable_balances,
+        retryable: Some(metadata.clone().retryable_balance),
+    };
+
+    Interpreter::<PredicateStorage>::estimate_predicates(&mut transaction, balances, params, GasCosts::default())
+        .expect("Should successfully estimate predicates");
```
xgreenx, line 301: I guess you can call `transaction.estimate_predicates` instead of duplicating the logic

fuel-vm/src/tests/predicate.rs
```diff
@@ -105,10 +124,188 @@ fn get_verifying_predicate() {
             op::ret(0x10),
         ];
 
-        assert!(execute_predicate(predicate, vec![], idx as usize));
+        assert!(execute_predicate(predicate, vec![], idx as usize, false));
     }
 }
 
+#[test]
+fn gas_used_by_predicates_causes_out_of_gas_during_script() {
```
xgreenx, line 132: Could you move this test to the old place, please to not shuffle the commit history?=)

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: i agree, this looks like a useful simplification as the tx_gas_limit could need to be adjusted after estimating predicates

fuel-vm/src/tests/predicate.rs
mitch-fuel: this was meant to explicitly test the call to the estimate_predicates function at Tommi's request, otherwise testing only occurs at the transaction.estimate_predicates level and not at the function level.

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -536,6 +565,46 @@ impl ConsensusParameters {
...
+            gas_price_factor,
+            gas_per_byte,
+            max_message_data_length,
+            chain_id,
+        }
+    }
+
+    /// Replace the max gas per predicate with the given argument
+    pub const fn with_max_gas_per_predicate(self, max_gas_per_predicate: u64) -> Self {
+        let Self {
```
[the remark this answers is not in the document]
mitch-fuel: would you like to do this for every function in this file? this is the pattern for every other function in this file

fuel-tx/src/transaction/types/input.rs
```diff
@@ -412,6 +438,16 @@ impl Input {
         }
     }
 
+    pub fn predicate_gas_used(&self) -> Option<Word> {
+        match self {
+            Input::CoinPredicate(CoinPredicate { predicate_gas_used, .. })
+            | Input::MessageCoinPredicate(MessageCoinPredicate { predicate_gas_used, .. })
+            | Input::MessageDataPredicate(MessageDataPredicate { predicate_gas_used, .. }) => Some(*predicate_gas_used),
+            Input::CoinSigned(_) | Input::MessageCoinSigned(_) | Input::MessageDataSigned(_) => Some(0),
```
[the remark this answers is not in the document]
mitch-fuel: is it not reasonable  for us to have the gas used be zeroed for every non-predicate value as they have zero gas used?

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: I believe I tried that initially and Brandon and I modified it to use the mutable iterator. I'll investigate this further

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: the RuntimePredicate returned here does not have the gas_used returned, only the range of the predicate itself, this complexity led to the change in the logic to more closely match the iteration similar to estimate_predicates

mitch-fuel: are you commenting on the removal of the predicate collection in favor of iterating over all inputs or simply the cumulative_gas_used logic?

fuel-vm/src/tests/predicate.rs
freesig: This was meant to be a unit test of `fn estimate_predicates`. Can it go in `main/tests.rs`?

fuel-vm/src/interpreter/executors/main.rs
mitch-fuel: added CumulativePredicateGasExceededTxGasLimit

fuel-vm/src/tests/predicate.rs
mitch-fuel: returned, not sure what happened here. I may have copy pasted then removed the wrong one

Voxelot:
> ..
> 4. `TransactionFee::checked_from_values` should statically price the base tx cost based on the total sum of all `predicate_gas_used` fields on the inputs.

This is still missing. Without it, the gas used by predicates wont get charged properly and might be refunded to the user via change.

mitch-fuel:
> > ..
> > 4. `TransactionFee::checked_from_values` should statically price the base tx cost based on the total sum of all `predicate_gas_used` fields on the inputs.
> 
> This is still missing. Without it, the gas used by predicates wont get charged properly and might be refunded to the user via change.

Adding this in now.

fuel-vm/src/checked_transaction.rs
freesig: Ok so then we can remove this clone?

xgreenx (dismissed):
The PR is ready for the final review @freesig @Voxelot 

> ..
> 4. `TransactionFee::checked_from_values` should statically price the base tx cost based on the total sum of all `predicate_gas_used` fields on the inputs.

We already use `Transaction::gas_limit` to charge the user, and we have requirements that the cumulative gas of predicates <= than the gas limit. So we don't need to update the logic.

Also I [updated](https://github.com/FuelLabs/fuel-specs/pull/487) the specification according to changes in this PR

fuel-tx/src/transaction/types/input.rs
```diff
@@ -66,6 +66,23 @@ impl AsFieldFmt for u8 {
...
+        Some(self)
+    }
+}
+
+impl AsFieldFmt for u64 {
+    fn fmt_as_field(&self, f: &mut Formatter) -> fmt::Result {
+        f.write_str(self.to_string().as_str())
+    }
+}
+
```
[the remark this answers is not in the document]
Voxelot: Is this resolved?

fuel-tx/src/transaction/types/input/message.rs
```diff
@@ -152,7 +161,11 @@ where
 {
     /// It is empty, because specification says nothing:
```
Voxelot, line 162: Is this comment still relevant? Looks like this fn is no longer empty

fuel-vm/src/checked_transaction.rs
```diff
@@ -184,22 +186,56 @@ pub trait CheckPredicates: Sized {
...
+}
+
 impl<Tx: ExecutableTransaction> CheckPredicates for Checked<Tx>
 where
-    Self: Clone,
     <Tx as IntoChecked>::Metadata: crate::interpreter::CheckedMetadata,
 {
     fn check_predicates(mut self, params: &ConsensusParameters, gas_costs: &GasCosts) -> Result<Self, CheckError> {
         if !self.checks_bitmask.contains(Checks::Predicates) {
             // TODO: Optimize predicate verification to work with references where it is possible.
```
Voxelot, line 201: is this todo resolved?

fuel-tx/src/transaction/types/input.rs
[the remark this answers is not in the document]
xgreenx: Yep, now we make them always empty

fuel-vm/src/checked_transaction.rs
xgreenx: We still close `gas_costs`

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,50 +61,148 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+        let balances = checked.metadata().balances();
+        Self::run_predicate(PredicateRunKind::Verifying(tx), balances, params, gas_costs)
+    }
+
+    /// Initialize the VM with the provided transaction and check all predicates defined in the
+    /// inputs. Set the predicate_gas_used to be the actual gas consumed during execution for
+    /// each predicate.
+    ///
+    /// The storage provider is not used since contract opcodes are not allowed for predicates.
+    /// This way, its possible, for the sake of simplicity, it is possible to use
```
Voxelot, line 82:
```suggestion
    /// This way, for the sake of simplicity, it is possible to use
```

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -42,50 +61,148 @@ impl<T> Interpreter<PredicateStorage, T> {
...
+                        }
+                    },
+                }
             }
+        }
 
-            remaining_gas = vm.registers[RegId::GGAS];
+        match kind {
+            PredicateRunKind::Verifying(tx) => {
+                if cumulative_gas_used > tx.limit() {
```
Voxelot, line 193: Shouldn't `cumulative_gas_used` also factor in the min_fee? Otherwise this limit check is only checking the total gas used by predicates against the total limit for the whole transaction.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -106,6 +223,16 @@ where
         let root = contract.root();
         let storage_root = Contract::initial_state_root(storage_slots.iter());
         let id = contract.id(salt, &root, &storage_root);
+        let mut cumulative_predicate_gas: Word = 0;
+
+        // Calculate cumulative predicate gas for inputs
+        for input in create.inputs() {
```
Voxelot, line 229: Why duplicate this calculation & logic here when it's already known from the `Checked<Create>` `metadata().gas_used_by_predicates`?

xgreenx: Because we don't have a `Checked` transaction during execution=) We can try to store it inside the `Interpreter` but I think it is simpler to calculate it than propagate

Voxelot (approved):

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -128,7 +231,10 @@ where
             .deploy_contract_with_id(salt, storage_slots, &contract, &root, &id)
             .map_err(InterpreterError::from_io)?;
 
-        let remaining_gas = create.limit();
+        let remaining_gas = create
```
Voxelot, line 234: Could we deduct the predicate gas before deploying the contract to save some extra computation in case it fails?

Voxelot (dismissed):

xgreenx (dismissed):

Voxelot (dismissed):

Voxelot (approved):

xgreenx:
Merged with https://github.com/FuelLabs/fuel-vm/pull/460

PR #388, opened 2023-03-15
Make `double_key!` public

ControlCplusControlV:
Need to make `double_key!` public in order to resolve [2 TODO's](https://github.com/search?q=repo%3AFuelLabs%2Ffuel-core%20double_key&type=code) currently in `fuel-core`


fuel-vm/src/storage.rs
```diff
@@ -160,5 +160,7 @@ macro_rules! double_key {
     };
 }
 
+pub(crate) use double_key;
```
xgreenx, line 163: You need to export macro=)

ControlCplusControlV: Oh that is only within the same crate my bad

xgreenx (approved):

PR #391, opened 2023-03-16
Use `#![deny(unused_crate_dependencies)]` instead of `udeps`

xgreenx:
The CI [fails](https://github.com/FuelLabs/fuel-vm/actions/runs/4437744090/jobs/7787856328?pr=380) with the latest nightly Rust for `udeps`. This PR removes the usage of the `udeps` and replaces it with `unused_crate_dependencies`. Because of specific how linting works, I moved all `tests` folders into the `src` folder and compile integration tests as one module. Also, it improved the compilation time for tests([see](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#integration-tests) for details why). Also found several unused dependencies with `unused_crate_dependencies`


fuel-crypto/Cargo.toml
```diff
@@ -22,16 +22,10 @@ serde = { version = "1.0", default-features = false, features = ["derive"], opti
 sha2 = { version = "0.10", default-features = false }
 zeroize = { version = "1.5", features = ["derive"] }
 
-# Note that, while we don't use this dependency directly, we must specify it in
-# order to enable the transitive dependency's "js" feature which is required
-# for supporting the wasm32-unknown-unknown target.
-[target.'cfg(target_arch = "wasm32")'.dependencies]
```
xgreenx, line 28: I'm only trying to figure out this part. @mitchmindtree Added it before=) He may know if we still need it or not. Usually, the `js` feature is required if you try to use the `std` code in the WASM.

`cargo check --target wasm32-unknown-unknown -p fuel-crypto --no-default-features` compiles successfully for me

Voxelot: I think this was potentially to help address the wasm issues the indexer was having? Does it still compile to wasm with `--features random` enabled?

Voxelot:
> Another bonus of all tests in one module is that it compiles tests for me much faster. I would say 10 seconds versus 1 min=)

Makes sense, that's why we use a custom test harness in fuel-core so we don't have to link rocksdb into dozens of binary targets for each integ test module.

fuel-crypto/Cargo.toml
```diff
@@ -22,16 +22,10 @@ serde = { version = "1.0", default-features = false, features = ["derive"], opti
...
-getrandom = { version = "0.2.8", default-features = false, features = ["js"] }
-
 [dev-dependencies]
 bincode = { version = "1.3", default-features = false }
 criterion = "0.3"
-fuel-crypto = { path = ".", default-features = false, features = ["random"] }
+fuel-crypto = { path = ".", default-features = false, features = ["random", "serde"] }
```
Voxelot, line 28: should this be only enabled via the serde feature flag?

fuel-crypto/Cargo.toml
xgreenx: > Does it still compile to wasm with --features random enabled?

Yeah, it compiles with `cargo build --target wasm32-unknown-unknown -p fuel-crypto --no-default-features --features random`

fuel-crypto/Cargo.toml
xgreenx: Remove it=)

fuel-tx/Cargo.toml
```diff
@@ -23,9 +23,9 @@ serde = { version = "1.0", default-features = false, features = ["alloc", "deriv
 serde_json = { version = "1.0", default-features = false, features = ["alloc"], optional = true }
 
 [dev-dependencies]
-bincode = { version = "1.3", default-features = false }
+bincode = { workspace = true }
 fuel-crypto = { workspace = true, default-features = false, features = ["random"] }
-fuel-tx = { path = ".", features = ["builder", "random"] }
+fuel-tx = { path = ".", features = ["builder", "random", "serde"] }
```
Voxelot, line 28: Also should only be enabled when the serde feature is turned on

```suggestion
fuel-tx = { path = ".", features = ["builder", "random"] }
```

xgreenx: Added
```rust
#[cfg(feature = "serde")]
mod bytes;
#[cfg(feature = "serde")]
mod display;
#[cfg(not(feature = "serde"))]
use bincode as _;
```

PR #393, opened 2023-03-17
add dynamic cfe and cfs

Voxelot:
Related spec change: https://github.com/FuelLabs/fuel-specs/pull/471

Adds two new opcodes for dynamic call frame manipulation: `CFE` & `CFS`


fuel-asm/src/lib.rs
```diff
@@ -247,6 +247,10 @@ impl_instructions! {
     0x91 CFEI cfei [Imm24]
     "Shrink the current call frame's stack by an immediate value."
     0x92 CFSI cfsi [Imm24]
+    "Extend the current call frame's stack"
```
Voxelot, line 250: Added these to the end to avoid breaking existing bytecode outputs.

freesig (approved):
Looks good

PR #403, opened 2023-03-21
bump to 0.28.1

Voxelot:
unblock publishing


xgreenx (approved):
Are you sure that we need to do that? I guess we can still publish a previous version=)

Voxelot:
Since fuel-types already got published, I thought the publish job would fail since the commit dates wouldn't match.

PR #405, opened 2023-03-21
Make verification for binary merkle tree public

xgreenx:


bvrooman:
Is this because we need BMT proofs in production now?

The thing is, the `verify` algorithm that is in the `test-helpers` is not really production ready. It is simply a loose translation of a standard Go implementation into Rust. The `test-helpers::verify` function exists to check the _correctness_ of the proof sets generated with BMT `prove`, but it doesn't have the same optimizations or considerations as a production-ready implementation.

If you need to expose the verify algorithm, can I suggest that you copy it instead of move it? This way, we can iterate on the public version to use in production, and we still have a `test-helpers` version to check the _correctness_ of the public version.

xgreenx:
I need it for unit tests right now, but I'm sure it would be helpful for the SDK users=)

Copy-pasting means that I need to bring `sum_node` and other functions too.

fuel-merkle/src/binary/verify.rs
```diff
@@ -0,0 +1,165 @@
+use crate::binary::node_sum;
+use crate::common::Bytes32;
+use crate::common::ProofSet;
+
+pub fn verify(root: &Bytes32, proof_set: &ProofSet, proof_index: u64, num_leaves: u64) -> bool {
```
xgreenx, line 5: @bvrooman Do you have ideas on how we can specify the data of the leaf?=)

xgreenx: It seems the first item in the set is a "data"
image

bvrooman: With this algorithm, it expects that the first item in the proof set is the leaf that you are proving. 
In your above screenshot, we get item 0 from `proof_set` using the `height` (since `height = 0`), and continue to accumulate the hashes from there.

bvrooman (dismissed):
If you're simply using this for unit tests in the mean time, that is fine. We can iterate on the implementation.

Voxelot (approved):

Issue #372, opened 2023-03-04
Chain ID 


Voxelot:
- [x] Add `chain_id: u64` to consensus parameters
- [x] Change predicate owner checks to a separate predicate ID function instead of contract ID, which includes the `chain_id` (see https://github.com/FuelLabs/fuel-specs/pull/462)
- [x] Update the transaction id hash to include chain_id for chain specific signatures

Voxelot:
related: https://github.com/FuelLabs/fuel-specs/issues/340

PR #406, opened 2023-03-22
Chain ID included in transaction ID's and predicate owners

Voxelot:
related to: https://github.com/FuelLabs/fuel-specs/pull/462, [#372](https://github.com/FuelLabs/fuel-vm/issues/372)

- hashes chain id into predicate owner
- hashes chain id into transaction id
- add GM op for fetching the chain id in the vm


fuel-tx/src/builder.rs
```diff
@@ -276,25 +276,25 @@ impl<Tx: Buildable> TransactionBuilder<Tx> {
     }
 
     #[cfg(feature = "std")]
-    pub fn _finalize(&mut self) -> Tx {
+    pub fn _finalize(&mut self, parameters: &ConsensusParameters) -> Tx {
```
xgreenx, line 279: It is better if the builder stores the `ConsensusParameters` inside. In this case, we don't need to propagate the parameter everywhere and break the API. 

We can use `Builder.set_consensus_parametrs` in tests where we want to use it. The same pattern that we use in the `fuel-core` where we have `Option<ConsensusParameters>` field=)

fuel-tx/src/transaction/types/input.rs
```diff
@@ -566,23 +566,29 @@ impl Input {
         compute_message_id(sender, recipient, nonce, amount, data)
     }
 
-    pub fn predicate_owner<P>(predicate: P) -> Address
+    pub fn predicate_owner<P>(predicate: P, params: &ConsensusParameters) -> Address
```
xgreenx, line 569: Hmm, maybe we need to notify other teams about that explicitly

fuel-tx/src/transaction/types/script.rs
```diff
@@ -102,7 +103,12 @@ impl crate::UniqueIdentifier for Script {
         clone.outputs_mut().iter_mut().for_each(Output::prepare_sign);
         clone.witnesses_mut().clear();
 
-        fuel_crypto::Hasher::hash(clone.to_bytes().as_slice())
+        let mut hasher = Hasher::default();
+        // chain ID
+        hasher.input(params.chain_id.to_be_bytes());
+        // transaction bytes
+        hasher.input(clone.to_bytes().as_slice());
+        hasher.finalize()
```
xgreenx, line 111: [nit] It Would be nice to move this code in the `compute_tx_id` function and reuse in 3 places=)

xgreenx (dismissed):

fuel-tx/src/transaction/types/input.rs
Voxelot: yep - it's a big change

xgreenx (approved):

Issue #389, opened 2023-03-16
Truncate large byte fields in debug output


Voxelot:
There are several fields on transactions which can get quite large, such as when a creation transaction with a big contract submitted. We should truncate large fields like witness or code when debug formatting transactions to avoid clogging logs and terminal outputs.

PR #413, opened 2023-03-28
Truncate large binary arrays when debug logging

Voxelot:
fixes: https://github.com/FuelLabs/fuel-vm/issues/389

When debug formatting types like Witness, script, script_data, predicate, predicate_data or message data, it will be truncated to at most 16 characters. This addresses an issue where large bytecode deployments causes the log ingestion by fluentbit/elastic to be scrambled.

Before:

`witness: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`

After:

`witness: "0x00000000000..."`


fuel-tx/src/transaction/types/witness.rs
```diff
@@ -8,12 +9,15 @@ use rand::{
...
 
+use derivative::Derivative;
 #[cfg(feature = "std")]
 use std::io;
 
-#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[derive(Derivative, Default, Clone, PartialEq, Eq, Hash)]
+#[derivative(Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Witness {
```
xgreenx, line 19: Maybe we want to do the same for the `Contract` struct?

xgreenx (commented):
Hmm, maybe we also should do that for the `data` field in the `Receipt`

Issue #147, opened 2022-06-13
Ownership check shouldn't fail for empty memory range


vlopes11:
For a range `$m[x, 0]`, the ownership check fails here since `x < x + 0` will never be met

https://github.com/FuelLabs/fuel-vm/blob/52cbe36086cd09e2bbc0e89eb8009f272bd48850/src/interpreter/memory.rs#L230

Technically speaking, this isn't an illegal range, but it might lead to undefined behavior if used in some opcodes.

Example: `mcp $rA $rA $zero`

If this is considered a legal range, then we will

https://github.com/FuelLabs/fuel-vm/blob/52cbe36086cd09e2bbc0e89eb8009f272bd48850/src/interpreter/memory.rs#L395

It clearly violates the constraints of [std](https://doc.rust-lang.org/std/index.html)::[ptr](https://doc.rust-lang.org/std/ptr/index.html)::[copy_nonoverlapping](https://doc.rust-lang.org/std/ptr/fn.copy_nonoverlapping.html#), but then the compiler might be smart enough to translate this into `noop` since the length is `0`.

We could alternatively add an extra check before any non-overlapping copy, but then we are increasing the execution cost as result of an edge case that should be `NOOP`

vlopes11:
Update: the example is illegal since `mcp` shouldn't overlap.

https://github.com/FuelLabs/fuel-specs/blob/fcfe18a585a465c750fd939b0fd85a30255d6c00/specs/vm/instruction_set.md?plain=1#L954

Maybe there isn't a single legal case for such ranges?

adlerjohn:
This isn't so much a bug as much as the specification doesn't allow for this behavior. I don't think it's necessarily desirable to add for the aforementioned reasons, namely performance.

PR #416, opened 2023-03-30
Ownership test of a validly-positioned zero-length range should succeed

Dentosal:
This case was broken, and untested. Fixed the bug, and added unit and integration tests so that it wont reappear.

See https://github.com/FuelLabs/fuel-vm/issues/147 for discussion.


fuel-vm/src/tests/memory.rs
```diff
@@ -181,3 +218,157 @@ fn dynamic_call_frame_ops() {
...
+    ops.push(op::movi(0x10, count + 1));
+    ops.push(op::logd(0, 0, RegId::HP, 0x10));
+    ops.push(op::ret(RegId::ONE));
+
+    let vm = setup(ops);
+    let vm: &Interpreter<MemoryStorage, Script> = vm.as_ref();
+
+    if let Some(Receipt::LogData { data, .. }) = vm.receipts().first() {
+        let c = count as usize;
+        assert_eq!(data.len(), c + 1);
```
Voxelot, line 258: Might be helpful to have some comments on these assertions to explain what's being tested.

Voxelot (approved):

PR #418, opened 2023-04-05
Add tests for cb, bhsh and bhei instructions

Dentosal:
There were no tests for these instructions. The tests added here feel a bit redundant, as they essentially just copy the implementation of the instructions themselves. On the hand, at least this makes sure they work

### DIFF
--- a/fuel-vm/src/tests/blockchain.rs
+++ b/fuel-vm/src/tests/blockchain.rs
@@ -1508,9 +1508,7 @@ fn timestamp_works() {
 }
 
 #[rstest::rstest]
-fn block_height_works(
-    #[values(0, 1, 2, 10, 100)] current_height: u32,
-) {
+fn block_height_works(#[values(0, 1, 2, 10, 100)] current_height: u32) {
     let current_height: BlockHeight = current_height.into();
 
     let mut client = MemoryClient::default();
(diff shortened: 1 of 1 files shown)


fuel-vm/src/tests/blockchain.rs
```diff
@@ -1506,3 +1506,131 @@ fn timestamp_works() {
...
+    let receipts = client.transact(tx);
+    let Some(Receipt::LogData { data, .. }) = receipts.first() else {
+        panic!("expected log receipt");
+    };
+
+    assert_eq!(data, &*expected);
+}
+
+#[rstest::rstest]
+fn block_proposer_works() {
```
Voxelot, line 1600: This should really be coinbase, as the proposer can use a different wallet for collecting fees (i.e. cold storage)

Dentosal: So is the spec wrong in this case? https://github.com/FuelLabs/fuel-specs/blob/master/src/vm/instruction_set.md#cb-block-proposer-address

mitch-fuel: if this will always be the coinbase, then is seems like the short text of the opcode needs to be changed

Voxelot: Yeah there's not a lot of benefit to having the block proposer address, John and I previously discussed that this should just be the coinbase address associated with the coinbase transaction on the block. We could add another opcode for block proposer later if we want to.

Voxelot: Made a PR for the specs here: https://github.com/FuelLabs/fuel-specs/pull/477

Dentosal: Updated PR

Voxelot (approved):

PR #419, opened 2023-04-07
Add convenience API for getting transaction ID without ConsensusParameters

Voxelot:
Since checked transactions are guaranteed to have computed all relevant metadata, this PR adds an infallible API to enable fetching the cached transaction ID for Checked\<Tx\> types.

This cleans up a lot of places in https://github.com/FuelLabs/fuel-core/pull/1107 where the consensus parameters would have to be routed even though the transaction id is already known (mainly executor and txpool since they use checked transactions).


mitch-fuel (approved):
Looks good

PR #424, opened 2023-04-14
Add wide math operations

Dentosal:
Adds wideint instructions as described in https://github.com/FuelLabs/fuel-specs/pull/483

TODO:

- [x] Bit shift error behaviors
- [x] Unit tests
- [x] Integration tests
- [ ] Gas benchmarking
- [x] Pick the fastest crate for each operation. For instance `ethnum` does some operations faster than `primitive_types`, but doesn't offer 512bit multiplication
- [ ] Update the requirements traceability document
- [ ] Maybe tune down the macro usage?


Dentosal:
Benchmarking results for the most expensive 256-bit operations:

Operation     | `ethnum`  | `primitive_types`
-------------|------------|------------------
Div                 | 44.93ns     |  13.73ns
Mod               | 45.35ns     | 13.66ns
checked_mul  | 4.98ns     | 9.10ns
full_mul          | -                | 8.55ns


To summarize , multiplication is faster for `ethnum`, but division is faster for `primitive_types`. However, `primitive_types` multiplication is slower because it computes the full multiplication result every time, even when it overflows. [The implementation](https://github.com/paritytech/parity-common/blob/910097e9e8c1c8aeb86a4da5e5b84ed890f0a2e0/uint/src/uint.rs#L252) uses manual loop unrolling. I'd like to submit a PR there, but that's low priority compared to making this work first.

Ideally we could use only the `primitive_types` library. It's API is not as clean as `ethnum` but operations like `mul+div` are not possible with `ethnum`.

fuel-vm/src/gas/default_gas_costs.rs
```diff
@@ -77,6 +78,20 @@ pub fn default_gas_costs() -> GasCostsValues {
...
+        wdml: 2,
+        wqml: 4,
+        wddv: 4,
+        wqdv: 8,
+        wdmd: 5,
+        wqmd: 10,
+        wdam: 2,
+        wqam: 4,
+        wdmm: 4,
+        wqmm: 8,
```
freesig, line 94: Are these actually generated from our benchmarks?

freesig (approved):
Nice job!

fuel-vm/src/gas/default_gas_costs.rs
Dentosal: Those are from microbenchmarks. I should probably do a full benchmarking run using fuel-core gas benchmarking as well.

freesig: Yeh they all need to be based on the same benchmark but I doubt it's a huge issue for math operations. We will have to re do all the gas soon anyway

mitch-fuel (dismissed):
LGTM

fuel-vm/src/gas/default_gas_costs.rs
mitch-fuel: It will definitely be done before it reaches anywhere that would be accessible to the public

Dentosal: I'm currently working on adding the instructions to fuel-core-benches. Do you think we should just merge this as-is and adjust the costs later?

fuel-vm/src/tests/wideint.rs
```diff
@@ -0,0 +1,654 @@
+use ethnum::U256;
```
Voxelot, line 1: looks like we're missing tests for WDMD?

Dentosal: Added those. I feel like adding copy-pasted tests for macro-generated code isn't too useful, but still nice to have for future proofing.

fuel-vm/src/interpreter/alu/muldiv.rs
```diff
@@ -0,0 +1,86 @@
...
+                ..
+            },
+            mut w,
+        ) = split_registers(&mut self.registers);
+        let dest = &mut w[ra.try_into()?];
+
+        let (result, overflow) = muldiv(lhs, rhs, divider);
+
+        if overflow != 0 && !is_wrapping(flag.into()) {
+            println!("!!");
```
Voxelot, line 34:
```suggestion
```

fuel-vm/src/interpreter/alu/muldiv.rs
```diff
@@ -0,0 +1,86 @@
...
+            return Err(PanicReason::ArithmeticOverflow.into());
+        }
+
+        *of = overflow;
+        *err = 0;
```
Voxelot, line 39: > If the [`F_UNSAFEMATH`](./index.md#flags) flag is set, an `ALU` operation that would have panicked will instead set `$err` to `true`.

https://github.com/FuelLabs/fuel-specs/blob/813faf0b94bc45cd5e9b55826b57831d27bed394/src/vm/instruction_set.md?plain=1#L144

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+            /// Drops higher half of the value and converts to `u128` or `ethnum::U256`
+            fn [<truncate_from_prim_ $t:lower>](value: primitive_types::[<$wider_t:upper>]) -> $t {
+                const S: usize = ::core::mem::size_of::<$t>();
+                let mut buffer = [0u8; 2 * S];
+                value.to_little_endian(&mut buffer);
+                let truncated: [u8; S] = buffer[..S].try_into().unwrap_or_else(|_| unreachable!());
+                $t::from_le_bytes(truncated)
+            }
+
+            impl<S, Tx> Interpreter<S, Tx>
```
Voxelot, line 50: should we be concerned with macro hygiene here?

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+                    };
+
+                    let lhs = [<to_prim_ $t:lower>](lhs);
+                    let rhs = [<to_prim_ $t:lower>](rhs);
+
+                    let result = match lhs.checked_div(rhs) {
+                        Some(d) => [<from_prim_ $t:lower>](d),
+                        None => {
+                            if is_unsafe_math(flag.into()) {
+                                $t::default() // Zero
```
Voxelot, line 177: should set `*err = 1` also

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+                            } else {
+                                return Err(PanicReason::ErrorFlag.into());
```
Voxelot, line 179: `ErrorFlag` is typically used when something unexpected happens (i.e. a bug or incorrectly validated transaction). Since a user could trigger this behavior without any sort of intervention on our end, I'd go with a different kind of panic reason like `ArithmeticOverflow` or something more specific.

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+
+                    let lhs = [<to_prim_ $t:lower>](lhs);
+                    let rhs = [<to_prim_ $t:lower>](rhs);
+
+                    let result = match lhs.checked_div(rhs) {
+                        Some(d) => [<from_prim_ $t:lower>](d),
+                        None => {
+                            if is_unsafe_math(flag.into()) {
+                                $t::default() // Zero
+                            } else {
```
Voxelot, line 178: if `is_wrapping == true`, should this behave like `overflowing_div` and set `*of = 1` without an error?

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+                    // RHS is only indirect if the flag is set
+                    let rhs: $t = if args.indirect_rhs {
+                        $t::from_be_bytes(read_bytes(&self.memory, c)?)
+                    } else {
+                        c.into()
+                    };
+
+                    let (wrapped, overflow) = [<op_overflowing_ $t:lower>](lhs, rhs, args);
+
+                    if overflow && !is_wrapping(flag.into()) {
```
Voxelot, line 101: should not panic if `F_UNSAFEMATH` is set

> If the [F_UNSAFEMATH](https://github.com/FuelLabs/fuel-vm/pull/424/index.md#flags) flag is set, an ALU operation that would have panicked will instead set $err to true.

https://github.com/FuelLabs/fuel-specs/blob/813faf0b94bc45cd5e9b55826b57831d27bed394/src/vm/instruction_set.md?plain=1#L144

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+                    // RHS is only indirect if the flag is set
+                    let rhs: $t = if args.indirect_rhs {
+                        $t::from_be_bytes(read_bytes(&self.memory, c)?)
+                    } else {
+                        c.into()
+                    };
+
+                    let (wrapped, overflow) = $t::overflowing_mul(lhs, rhs);
+
+                    if overflow && !is_wrapping(flag.into()) {
```
Voxelot, line 138: should not panic if `F_UNSAFEMATH` is set

> If the [F_UNSAFEMATH](https://github.com/FuelLabs/fuel-vm/pull/424/index.md#flags) flag is set, an ALU operation that would have panicked will instead set $err to true.

https://github.com/FuelLabs/fuel-specs/blob/813faf0b94bc45cd5e9b55826b57831d27bed394/src/vm/instruction_set.md?plain=1#L144

fuel-vm/src/interpreter/alu/muldiv.rs
```diff
@@ -0,0 +1,86 @@
...
+            },
+            mut w,
+        ) = split_registers(&mut self.registers);
+        let dest = &mut w[ra.try_into()?];
+
+        let (result, overflow) = muldiv(lhs, rhs, divider);
+
+        if overflow != 0 && !is_wrapping(flag.into()) {
+            println!("!!");
+            return Err(PanicReason::ArithmeticOverflow.into());
```
Voxelot, line 35: should not panic if `F_UNSAFEMATH` is set

> If the [F_UNSAFEMATH](https://github.com/FuelLabs/fuel-vm/pull/424/index.md#flags) flag is set, an ALU operation that would have panicked will instead set $err to true.

https://github.com/FuelLabs/fuel-specs/blob/813faf0b94bc45cd5e9b55826b57831d27bed394/src/vm/instruction_set.md?plain=1#L144

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+
+                    let lhs: $t = $t::from_be_bytes(read_bytes(&self.memory, b)?);
+                    let rhs: $t = $t::from_be_bytes(read_bytes(&self.memory, c)?);
+                    let modulus: $t = $t::from_be_bytes(read_bytes(&self.memory, d)?);
+
+                    let result: $t = if modulus == 0 {
+                        if is_unsafe_math(flag.into()) {
+                            $t::default() // Zero
+                        } else {
+                            return Err(PanicReason::ErrorFlag.into());
```
Voxelot, line 210: Should probably use a more descriptive panic reason other than `ErrorFlag`.

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -0,0 +1,365 @@
...
+                        }
+                    } else {
+                        // Use wider types to avoid overflow
+                        let lhs = [<to_wider_prim_ $t:lower>](lhs);
+                        let rhs = [<to_wider_prim_ $t:lower>](rhs);
+                        let modulus = [<to_wider_prim_ $t:lower>](modulus);
+
+                        // Trunacte never loses data as modulus is still in domain of the original type
```
Voxelot, line 218:
```suggestion
                        // Truncate never loses data as modulus is still in domain of the original type
```

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: Non-wideint ALU operations use ErrorFlag for any error condition, including divion by zero. I should probably update them as well, if we're going to that direction.

fuel-vm/src/interpreter/alu/muldiv.rs
Dentosal: I think the spec is worded incorrectly. Unsafemath doesn't supress panic when an overflow occurs. Instead, F_WRAPPING controls this behavior. Currently F_UNSAFEMATH is only used for zero divisor, modulus and root,logarithm of zero or with a base <= 1, and (incorrectly, see https://github.com/FuelLabs/fuel-vm/issues/426) bit shifts over 2<sup>32</sup> bits.

I'll make a PR to clarify the spec. I feel like we could even combine the flags since no instruction uses both of them. We could even combine $of and $err registers into one as well. But the easiest would be to simply make the spec clear about the difference.

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: Dividing by zero is an arithmetic error and not an overflow.  For instance the Rust function [`u64::overflowing_div`](https://doc.rust-lang.org/std/primitive.u64.html#method.overflowing_div) does the following:

> Returns a tuple of the divisor along with a boolean indicating whether an arithmetic overflow would occur. Note that for unsigned integers overflow never occurs, so the second value is always false.
> ### Panics
> This function will panic if rhs is 0.

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: Please clarify. I'm not sure what the potential issue would be, as the macros are expanded below in the same file.

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: Good catch

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: This is consistent with other ALU behaviour. Would it make sense to update those as well?

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: See below

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: Done, the new flag is named `ArithmeticError`

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: See below

fuel-vm/src/interpreter/alu/wideint.rs
Dentosal: Updated.

freesig (approved):

PR #435, opened 2023-05-02
Allow empty range in the checked memory

xgreenx:
The `SMO` opcode may pass empty array with zero size.

### DIFF
--- a/fuel-vm/src/constraints.rs
+++ b/fuel-vm/src/constraints.rs
@@ -97,8 +97,7 @@ impl CheckedMemRange {
         Self::new_inner(address as usize, size, constraint)
     }
 
-    /// Create a new memory range, checks that the range is not empty
-    /// and that it fits into the constraint.
+    /// Create a new memory range, checks that the range fits into the constraint.
     fn new_inner(address: usize, size: usize, constraint: core::ops::Range<Word>) -> Result<Self, RuntimeError> {
         if size == 0 {
             return Ok(Self(address..address));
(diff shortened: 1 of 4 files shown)


fuel-vm/src/interpreter/blockchain/smo_tests.rs
```diff
@@ -74,7 +74,7 @@ impl Default for Input {
         msg_data_len: 0,
         amount_coins_to_send: 0,
         ..Default::default()
-    } => Err(RuntimeError::Recoverable(PanicReason::MemoryOverflow))
+    } => matches Ok(Output { .. })
     ; "Call abi can't be zero"
```
Voxelot, line 78:
```suggestion
    ; "message data can be zero-length"
```

fuel-vm/src/constraints.rs
```diff
@@ -94,15 +94,19 @@ impl CheckedMemRange {
         if constraint.end > VM_MAX_RAM {
             return Err(Bug::new(BugId::ID009, BugVariant::InvalidMemoryConstraint).into());
         }
-        Self::new_inner(address, size, constraint)
+        Self::new_inner(address as usize, size, constraint)
     }
 
     /// Create a new memory range, checks that the range is not empty
```
Voxelot, line 100: This comment is no longer correct.

Does this change break the specs for other opcodes unrelated to SMO, such as MCP? I didn't see anything specifically disallowing this from the specs on cursory glance.

cc @Dentosal

Dentosal: This shouldn't break anything there. Zero-length MCP is allowed by the spec, and should work correctly in the VM ([test case](https://github.com/FuelLabs/fuel-vm/blob/62d35e57f69d31d8bfa014e4f2f200e8e9e4fbd0/fuel-vm/src/tests/memory.rs#L272)).

Voxelot (approved):

PR #442, opened 2023-05-12
added methods to convert between ContractId and AssetId

SwayStar123:
its much simpler to do `asset_id.into()` than `ContractId::new(*asset_id)`

### DIFF
--- a/fuel-types/src/array_types.rs
+++ b/fuel-types/src/array_types.rs
@@ -309,6 +309,18 @@ impl AssetId {
     pub const BASE: AssetId = AssetId::zeroed();
 }
 
+impl From<ContractId> for AssetId {
+    fn from(contract_id: ContractId) -> Self {
+        AssetId::new(*contract_id)
+    }
+}
+
+impl From<AssetId> for ContractId {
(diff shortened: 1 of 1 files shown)


Voxelot:
This wouldn't be as convenient as .into or .from, but you could make a new method like convert that will generically work across all compatible array types automatically.

If you put the following code into the primary impl block of the `key_methods` macro, you will be able to convert between any of these array types:

```
            pub fn convert<T: From<[u8; $s]>>(self) -> T {
                let val: [u8; $s] = self.into();
                T::from(val)
            }
```

xgreenx:
The trait `From` is usually used for objects with the same meaning but different representations. I don't like the idea of converting `ContracttId` into `AssetId` by this trait.

I agree with Brandon that it is better to add a new explicit method like `convert` or `from_contract_id`.

Also, with the upcoming multi-token inside of one contract feature, a method like `from_contract_id` will have the next signature:

```rust
fn from_contract_id(contract_id: ContractId, nonce: AssetIdNonce) -> Self { ... }
```

PR #444, opened 2023-05-12
Addressing minor issues

xgreenx:
Fixes https://github.com/FuelLabs/fuel-vm/issues/445

### DIFF
--- a/fuel-vm/src/constraints.rs
+++ b/fuel-vm/src/constraints.rs
@@ -105,7 +105,7 @@ impl CheckedMemRange {
 
         let (end, of) = address.overflowing_add(size);
         let range = address..end;
-        if of || !constraint.contains(&((range.end - 1) as Word)) {
+        if of || range.is_empty() || !constraint.contains(&((range.end - 1) as Word)) {
             return Err(PanicReason::MemoryOverflow.into());
         }
         Ok(Self(range))
--- a/fuel-vm/src/constraints.rs
+++ b/fuel-vm/src/constraints.rs
@@ -99,13 +99,13 @@ impl CheckedMemRange {
 
     /// Create a new memory range, checks that the range fits into the constraint.
     fn new_inner(address: usize, size: usize, constraint: core::ops::Range<Word>) -> Result<Self, RuntimeError> {
-        if size == 0 {
-            return Ok(Self(address..address));
-        }
-
         let (end, of) = address.overflowing_add(size);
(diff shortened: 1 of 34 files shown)


mitch-fuel (dismissed):
looks good, but I just noticed that this is the same failure that I created an [issue](https://github.com/FuelLabs/fuel-vm/issues/445) for today

mitch-fuel (approved):
looks good, great job getting the test fix in as well

fuel-vm/src/constraints.rs
```diff
@@ -99,13 +99,13 @@ impl CheckedMemRange {
...
-        if size == 0 {
-            return Ok(Self(address..address));
-        }
-
         let (end, of) = address.overflowing_add(size);
         let range = address..end;
-        if of || !constraint.contains(&((range.end - 1) as Word)) {
+
+        if of
+            || range.start != 0 && !constraint.contains(&((range.start - 1) as Word))
```
Voxelot, line 106:
```suggestion
            || range.start != 0 && !constraint.contains(&((range.start) as Word))
```

fuel-vm/src/constraints.rs
```diff
@@ -99,13 +99,13 @@ impl CheckedMemRange {
...
+            || range.end != 0 && !constraint.contains(&((range.end - 1) as Word))
```
Voxelot, line 107: If I have a constraint of 1..5, but my range is 0..5, would this allow constraint check to be incorrectly skipped because the range.start == 0?

fuel-vm/src/tests/blockchain.rs
```diff
@@ -545,9 +545,9 @@ fn ldc_reason_helper(cmd: Vec<Instruction>, expected_reason: PanicReason, should
     } = receipts.get(0).expect("No receipt")
     {
         assert_eq!(
-            &expected_reason,
+            &Some(expected_reason),
```
Voxelot, line 548: Makes me wonder if the panic receipt should really just have a non-optional panic reason, instead of instruction result. Seems strange to have the panic reason optional on a panic receipt.

Voxelot (changes requested):
blocking pr with change request since the memory constraint validation is unsafe.

Voxelot (approved):
Nice work! TryFrom for panic reasons makes more sense than using placeholder values that don't make sense.

Issue #434, opened 2023-05-02
Use 16kib leaf sizes when calculating code roots


Voxelot:
Spec change: https://github.com/FuelLabs/fuel-specs/pull/476

Improve the performance of calculating code roots by using a larger leaf size in the merkle tree to 16kib.

- Chunk bytecode and predicate code by 16kib chunks.
- Zero pad chunks out to the nearest multiple of 8 bytes if they are < 16kib. For example: if a predicate is only 500bytes, it should be padded out to 504 bytes no 16 kib.

freesig:
@Voxelot What is the purpose of the padding?

bvrooman:
> Zero pad chunks out to the nearest multiple of 8 bytes if they are < 16kib. For example: if a predicate is only 500bytes, it should be padded out to 508 bytes no 16 kib.

To be clear: 500 is already not divisible by 8; i.e. `500 mod 8 = 4`. Do you mean that 500 bytes would be zero padded to _504_ bytes, since `504 mod 8 = 0`?

Voxelot:
> @Voxelot What is the purpose of the padding?

To preserve the original padding we used of 8 bytes. This is likely somewhat arbitrary though.

Issue #385, opened 2023-03-15
Large contracts take a very long time to merkleize


Voxelot:
While we aim to support contract sizes of up to 16mb, this takes an excess of 20s+ to deploy onto a node due to the contract root calculation. The contract root exists to provide a mechanism to prove that the code being executed is actually part of the contract. However, the root is computed by chunking every opcode into a BMT leaf, which ends up causing millions of sha256 hash operations for a 16mb contract.

Our options are:
- Reduce max contract size to a level that is reasonable for merklizing
- Use larger chunks for the contract root (e.g. 32+ bytes / 4+ opcodes per leaf, maybe 1KiB chunks would be ok for ethereum fraud proofs?)
- Parallelize computing of contract roots (e.g. rayon)
- Investigate a more efficient way of computing the root when all leaves are present (e.g. regular merkle tree instead of BMT since it's not going to be appended to)
- Verify contract root on checked txs, and avoid re-verifying it during the `deploy` executor in fuel-vm (still potentially impacts our consensus/sync time)

Voxelot:
Some initial benchmarking shows that using grouping ops into larger chunks when merkleizing makes a fairly substantial impact on performance:

When merklizing 16kib of contract code
```
word sized (8) chunks          time:   [1.1178 s 1.1211 s 1.1245 s]
quad sized (32) chunks         time:   [234.35 ms 235.35 ms 236.45 ms]
1 kib sized (1024) chunks      time:   [11.627 ms 11.633 ms 11.639 ms]
page sized (4096) chunks       time:   [8.5189 ms 8.5468 ms 8.5737 ms]
```

ControlCplusControlV:
Is there a reason to use `rayon`? AFAIK you could spawn a tokio task for each contract root, and since they are very efficient and have minimal overhead being slightly heavier than a [goroutine](https://www.reddit.com/r/rust/comments/lg0a7b/benchmarking_tokio_tasks_and_goroutines/)

mitchmindtree:
Maybe a bit late in the game for this, but is it maybe worth seeing if it's feasible to swap out sha256 for blake3? https://crates.io/crates/blake3.

They seem to boast an order of magnitude improvement in performance in single-threaded environments, and it looks like it's also parallel-friendly so could potentially get even greater wins.

[...]

PR #450, opened 2023-05-18
feat: Update contract leaf size

bvrooman:
Related issue:
- Closes https://github.com/FuelLabs/fuel-vm/issues/434
- Closes https://github.com/FuelLabs/fuel-vm/issues/385


fuel-tx/src/contract.rs
```diff
@@ -193,4 +194,24 @@ mod tests {
         let default_root = Contract::default_state_root();
         insta::assert_debug_snapshot!(default_root);
     }
+
```
Voxelot, line 197: should we have a unit test to verify padding behavior?

xgreenx:
I think it is a good idea to add a new test to verify padding. The change for me looks good, could you also update the [changelog](https://github.com/FuelLabs/fuel-vm/pull/455) please?

xgreenx (commented):
Still waiting for updated changelog=D

fuel-tx/src/contract.rs
```diff
@@ -29,25 +45,17 @@ impl Contract {
         B: AsRef<[u8]>,
     {
         let mut tree = BinaryMerkleTree::new();
+        let mut bytes = bytes.as_ref().to_vec();
```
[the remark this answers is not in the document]
bvrooman: Yep, that's a good point. I updated the algorithm to only allocate and pad only if the last leaf is not already a multiple of 8.

bvrooman:
> Still waiting for updated changelog=D

Will update now

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and `Input::predicate_owner` use `ChainId` instead of the `ConsensusParameters`.
 It is a less strict requirement than before because you can get `ChainId` 
 from `ConsensusParameters.chain_id`, and it makes the API cleaner. 
 It affects all downstream functions that use listed methods.
+- [#450](https://github.com/FuelLabs/fuel-vm/pull/450): The Merkle root of a contract's code is now calculated by partitioning the code into chunks of 16 KiB, instead of 8 bytes. If the last leaf is not a full 16 KiB, it is padded with `0` up to the nearest multiple of 8 bytes.
```
xgreenx, line 30:
```suggestion
- [#450](https://github.com/FuelLabs/fuel-vm/pull/450): The Merkle root of a contract's code is now calculated by partitioning the code into chunks of 16 KiB, instead of 8 bytes. If the last leaf is not a full 16 KiB, it is padded with `0` up to the nearest multiple of 8 bytes. It affects the `ContractId` and `PredicateId` calculations, breaking all code used hardcoded values.
```

xgreenx (dismissed):

fuel-tx/src/contract.rs
```diff
@@ -29,25 +45,20 @@ impl Contract {
...
+            // should be zero-padded rounding up to the nearest multiple of 8
+            // bytes.
+            let len = leaf.len();
+            if len == LEAF_SIZE || len % MULTIPLE == 0 {
+                tree.push(leaf);
+            } else {
+                let padding_size = next_multiple::<MULTIPLE>(len);
+                let mut padded_leaf = vec![PADDING_BYTE; padding_size];
+                padded_leaf[0..len].clone_from_slice(leaf);
+                tree.push(padded_leaf.as_ref());
```
xgreenx, line 59: You can do that without vector=) You can create array `[PADDING_BYTE; MULTIPLE]` and only copy into first several bytes=)

bvrooman: I'm not sure that it's enough to create `[PADDING_BYTE; MULTIPLE]`, since your padding size will be some factor of `MULTIPLE` (i.e., `n * MULTIPLE`) if the remainder is already greater than `MULTIPLE`. We don't know the size at compile time, so I use a vector to accommodate that unknown size.

xgreenx: I meant the following=) But I don't know which is faster, your variant or mine=)
image

bvrooman: Do you have a preference on which version to use? The advantage of the vector approach is that it will have a smaller average memory allocation than the array approach (`0` to `LEAF_SIZE - 1` vs. constant `LEAF_SIZE`). The advantage of the array approach is, of course, that it allocates on the stack, rather than the heap.

xgreenx: We need to decide it based on the benchmarks=) Could you do that, please?

bvrooman: Yep, I will benchmark the different strategies and see what kind of difference it makes

bvrooman: I did some benchmarking using Criterion. I compared the two variations of the algorithm. 

The first version v1 (using `vec!`):
```rust
...
let padding_size = next_multiple::<MULTIPLE>(len);
let mut padded_leaf = vec![PADDING_BYTE; padding_size];
padded_leaf[0..len].clone_from_slice(leaf);
tree.push(padded_leaf.as_ref());
```

The second version V2 (using an array):
```rust
...
let padding_size = next_multiple::<MULTIPLE>(len);
let mut padded_leaf = [PADDING_BYTE; LEAF_SIZE];
padded_leaf[0..len].clone_from_slice(leaf);
tree.push(padded_leaf[..padding_size].as_ref());
```

The bench setup is as follows:
```rust
const LEAF_SIZE: usize = 1024 * 16;

fn bench(c: &mut Criterion) {
    let mut rng = StdRng::seed_from_u64(0xF00D);
    let code_len = 3 * LEAF_SIZE + 100;
    let mut code = alloc::vec![0u8; code_len];
    rng.fill_bytes(code.as_mut_slice());

    let mut group = c.benchmark_group("contract");
    group.bench_with_input("root_from_code", &code, |b, code| {
        b.iter(|| Contract::root_from_code(code))
    });
}

criterion_group!(benches, bench);
criterion_main!(benches);
```

First, I ran the bench using v1. The output is as follows:
```
Running benches/bench.rs (/Volumes/Fuel/projects/FuelLabs/fuel-vm/target/release/deps/bench-7aacf08cf5869b41)
contract/root_from_code time:   [152.83 µs 153.27 µs 153.76 µs]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
```

Second, I ran the bench using v2. the output is as follows:
```
Running benches/bench.rs (/Volumes/Fuel/projects/FuelLabs/fuel-vm/target/release/deps/bench-7aacf08cf5869b41)
contract/root_from_code time:   [152.03 µs 152.39 µs 152.77 µs]
                        change: [-0.2665% +0.1694% +0.6561%] (p = 0.47 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
```

I went back and forth between the two implementations, and most of further results were generally consistent with these initial results.

In conclusion, it seems that there is negligible difference in performance between the two implementations.

I can also include the benchmark code in this PR if that would be helpful. Just let me know and I can commit it here.

xgreenx: Could you also try `let code_len = 4 * LEAF_SIZE - 1;` please?

If the results are the same, then I prefer a variant with an array=)

bvrooman: Using `let code_len = 4 * LEAF_SIZE - 1;`, the results are:

v1:
```
Running benches/bench.rs (/Volumes/Fuel/projects/FuelLabs/fuel-vm/target/release/deps/bench-7aacf08cf5869b41)
contract/root_from_code time:   [200.90 µs 201.42 µs 201.97 µs]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

v2:
```
Running benches/bench.rs (/Volumes/Fuel/projects/FuelLabs/fuel-vm/target/release/deps/bench-7aacf08cf5869b41)
contract/root_from_code time:   [200.95 µs 201.41 µs 201.86 µs]
                        change: [-0.1991% +0.1953% +0.6080%] (p = 0.34 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
```

Yes, the results appear to be the same, even when changing the partial leaf size.

In that case, I will update the code to use the array version.

Please let me know if you want me to include the bench code as well and I will clean it up and commit it for review.

xgreenx: If there is nothing special, then maybe it is better not to keep this benchmark because the chance that we need it again is low=)

xgreenx (approved):

fuel-tx/src/contract.rs
bvrooman: Sounds good. I've updated the code now to use the array version, and it should be good to go.

PR #454, opened 2023-05-22
Do not hex-format primitives when format doesn't require it

Dentosal:
Uses [`is_human_readable` feature of serde](https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable). At least `postcard` which we're currently using for db benefits from this behavior. Works towards https://github.com/FuelLabs/fuel-core/issues/1085


Voxelot:
Should we have unit / snapshot tests to verify that roudtrip serialization works when is_human_readable is enabled / disabled?

Dentosal:
> Should we have unit / snapshot tests to verify that roudtrip serialization works when is_human_readable is enabled / disabled?

Yep. Those are already tested for in fuel-core, but it makes sense to have them tested here as well.

Voxelot (approved):

fuel-types/src/array_types.rs
```diff
@@ -266,7 +266,16 @@ macro_rules! key_methods {
                 S: serde::Serializer,
             {
                 use alloc::format;
-                serializer.serialize_str(&format!("{:x}", &self))
+                use serde::ser::SerializeTuple;
+                if serializer.is_human_readable() {
+                    serializer.serialize_str(&format!("{:x}", &self))
+                } else {
+                    let mut seq = serializer.serialize_tuple($s)?;
+                    for elem in self.0 {
```
xgreenx, line 274: The problem here and during deserialization is that you will iterate over each element and serialize/deserialize it separately. It is much slower than copying bytes into/from the slice.

We need to optimize that part for our primitives because we use them everywhere.

Maybe the optimizer already is clever enough to optimize it for arrays, but we need to proof it with benchmarks. 

For example, you can reuse implementation form the [`OptimizedContract`](https://github.com/FuelLabs/fuel-core/pull/1141/files/7153736da40d85a05c43305774b88821f55db846#diff-3d74057e7300bbc8bb721a964761e63eb79e1737a3b696744ffa6dbe9f829a6f) to compare with your implementation

Dentosal: The current way is what serde docs suggests to do with fixed size arrays.

You're right that the iteration here is slower than a slice copy; and indeed switching to `serialize_bytes` speeds this up quite a bit (4x by my microbenchmark). However, with the currently used `postcard` format, `serialize_bytes` prefixes the array with it's length in the serialized form, making it larger than necessary. Other formats (like `bincode`) do not do this, but bincode is much slower otherwise. I wish there was a serde format that had both of these figured out, but I haven't seen that yet.

In any case, I think having a single extra byte encoded for each of the primitives isn't ideal, but we cannot afford the slow down that comes with doing it right, so I'm updating this to use `serialize_bytes`.

xgreenx: Could you also try to bench it with `zerovec`, maybe for `serialize_bytes`, it doesn't use length

Dentosal: Yep, I'll do that later today

Dentosal: So after reading through zerovec docs, it's not apparent to me how `zerovec` would help us. `zerovec` is not a serialization format, but rather a set of types to assist with zero-copy deserialization using serde formats. Converting these types to use `zerovec` would make them less ergonomic to use, and creating new instances would require allocation, as `zerovec` doesn't do stack-allocated arrays.

What we could do instead is to use `Cow<'a, [u8; SIZE]>` for the array inside the type instead. This means that the type takes a lifetime argument. Sadly that breaks over 200 different locations where the types are used, and I don't necessarily feel like spending a whole day just adding lifetimes to them in hopes that it would be useful. That would also break things like `AsMut` trait impl, which is used in many many places as well.

These types are so cheap to copy that zero-copying them makes almost no sense anyways; I think we should instead focus on using zero-copy for the expensive types like `Contract`.

Voxelot: One extra byte at the start is definitely an improvement over what we had in the past. It's also common in other formats like scale as well.

xgreenx: In the `SCALE`, they don't put additional bytes for arrays=) Because the size is known

Dentosal: What is `SCALE`?

xgreenx: [Substrate](https://github.com/paritytech/parity-scale-codec) library for serialization and deserialization

Voxelot: It seems like the 0-prefixing is outside of our control from the serde impl perspective. We don't really have a way around this since we use postcard downstream, and so avoiding this overhead would require using an entirely different serde adapter (which is outside the scope of this PR imo)

xgreenx (commented):
Could you also update the CHANGELOG.md, please?

fuel-types/Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde = { version = "1.0", default-features = false, features = ["derive", "allo
 bincode = { workspace = true }
```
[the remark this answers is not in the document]
xgreenx: Hmm, where do we use `bincode`? I can't find any usage in the `fuel-core`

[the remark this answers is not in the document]
Dentosal: Oh apparently we don't anymore. In that case the tests should be updated as well. I'll do a follow-up PR, since that's out-of-scope for this.

Voxelot (approved):

xgreenx (approved):

PR #457, opened 2023-05-23
Inputs have at least one spendable coin or message

xgreenx:


Voxelot (approved):
This will definitely need some assistance with upgrading downstream, as I assume there are many places in sway / forc testing that run plain scripts without worrying about all the validity rules like this.

PR #458, opened 2023-05-23
Sort storage slots on create transactions automatically

### LINKED ISSUE #439: Transaction::Create requires that storage_slots is sorted
https://github.com/FuelLabs/fuel-tx/issues/153

Voxelot:
fixes #439 

This PR attempts a best effort to automatically sort storage slots when building create transactions. This is done in several places:

- When using the helper constructor for create transactions, the storage slots are sorted before constructing the transaction
- When mutating storage slots via the fields trait, a special wrapper is used to hold the mutable reference to storage slots which sorts when it's dropped.
- When using the `TransactionBuilder::create` helper fn, the storage slots are sorted before initializing the builder.


xgreenx:
I think we also need to add a new validation rule that checks that storage slots are sorted

Voxelot:
> I think we also need to add a new validation rule that checks that storage slots are sorted.

This should already be covered in the `check_without_signatures` for creation txs:

image

Will update the changelog

xgreenx (dismissed):

xgreenx (approved):

PR #459, opened 2023-05-24
Reuse witnesses for identical secrets when building transactions

### LINKED ISSUE #440: Witness signature: dedup by owner
https://github.com/FuelLabs/fuel-tx/issues/72

Voxelot:
This PR fixes #440 by reusing witness data for matching keys.

It also adds a breaking change by requiring add_unsigned_input for plain transaction (non-builder mode) to specify a witness index instead of naively adding a new witness for each input automatically.


xgreenx (commented):
Do we want to add a new validity rule: Each witness should be unique? It also will remove duplicate contracts from the transaction

Voxelot:
> Do we want to add a new validity rule: Each witness should be unique? It also will remove duplicate contracts from the transaction

We could, although that should go through the specs first.

xgreenx (dismissed):

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ Description of the upcoming release here.
 
 - [#456](https://github.com/FuelLabs/fuel-vm/pull/456): Added a new type - `ChainId` to represent the identifier of the chain. 
 It is a wrapper around the `u64`, so any `u64` can be converted into this type via `.into()` or `ChainId::new(...)`.
+- [#459](https://github.com/FuelLabs/fuel-vm/pull/459) Require witness index to be specified when adding an unsigned coin to a transaction.
+This allows for better reuse of witness data when using the transaction builder and helper methods to make transactions compact.
```
xgreenx, line 46:
```suggestion
- [#456](https://github.com/FuelLabs/fuel-vm/pull/456): Added a new type - `ChainId` to represent the identifier of the chain. 
It is a wrapper around the `u64`, so any `u64` can be converted into this type via `.into()` or `ChainId::new(...)`.

- [#459](https://github.com/FuelLabs/fuel-vm/pull/459) Require witness index to be specified when adding an unsigned coin to a transaction.
This allows for better reuse of witness data when using the transaction builder and helper methods to make transactions compact.
```

xgreenx (approved):

fuel-tx/src/builder.rs
```diff
@@ -278,6 +281,20 @@ impl<Tx: Buildable> TransactionBuilder<Tx> {
         self
     }
 
+    /// Adds a secret to the builder, and adds a corresponding witness if it's a new entry
+    #[cfg(feature = "std")]
+    fn upsert_secret(&mut self, secret_key: SecretKey) -> u8 {
```
bvrooman, line 286: `upsert` - I had to look that one up 👍

PR #461, opened 2023-05-24
Paged memory allocation

### LINKED ISSUE #452: Add memory pages
Instead of allocation `VM_MAX_RAM` on each `Interpreter`, dynamically allocate pages(for example, 16 KB) of memory and charge user only for those pages.

The tricky part is all memory range checks, increasing the stack size on `op::aloc`, loading of the contract, and so on. 
The maximum memory should still be `VM_MAX_RAM` and all previous rules should apply. 

During implementation decide: to use the same vector for heap and stack or have two separate memories.

The memory addressing should be dynamically translated into/from stack/heap.

Dentosal:
Closes #452. Spec PR https://github.com/FuelLabs/fuel-specs/pull/489.

Still has lots of unimplemented details.


fuel-vm/src/constraints.rs
```diff
@@ -181,16 +16,28 @@ pub struct InstructionLocation {
...
+    pub const fn offset(&self) -> u64 {
+        self.offset
     }
 }
 
-impl<const LEN: usize> DerefMut for CheckedMemConstLen<LEN> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
+impl<S, Tx> Interpreter<S, Tx> {
+    pub(crate) fn current_location(&self) -> InstructionLocation {
```
xgreenx, line 37: Previously it was under `#[cfg(feature = "profile-gas")]` flag and didn't participate in the production. But now, it will. I think we need to put fields under this flag too(or use different `InstructionLocation` for each flag)

fuel-vm/src/interpreter/alu/muldiv.rs
```diff
@@ -16,29 +16,17 @@ where
...
-            SystemRegisters {
-                flag,
-                mut of,
-                mut err,
-                pc,
-                ..
-            },
-            mut w,
-        ) = split_registers(&mut self.registers);
-        let dest = &mut w[ra.try_into()?];
```
xgreenx, line 29: Hmm, I don't know how it worked before, but for me `w[ra]`(where `w` is a writable registers without system registers) and `self.registers[ra]`(it is system registers and writable together) are different things. Because the lengths of arrays are different and with the same `ra`, they will point o different registers.

Is that a bug? If yes, what is the right version?

fuel-vm/src/interpreter/alu/wideint.rs
```diff
@@ -58,22 +57,19 @@ macro_rules! wideint_ops {
...
 
                     // RHS is only indirect if the flag is set
                     let rhs: $t = if args.indirect_rhs {
-                        $t::from_be_bytes(read_bytes(&self.memory, c)?)
+                        $t::from_be_bytes(self.mem_read_bytes(c)?)
                     } else {
                         c.into()
                     };
 
-                    *dest = [<cmp_ $t:lower>](lhs, rhs, args.mode);
```
xgreenx, line 74: The same here. What is the right variant?

fuel-vm/src/interpreter/balances.rs
```diff
@@ -91,14 +92,12 @@ impl RuntimeBalances {
         self.state.get(asset).map(Balance::value)
     }
 
-    fn set_memory_balance_inner(balance: &Balance, memory: &mut [u8; MEM_SIZE]) -> Result<Word, RuntimeError> {
+    fn set_memory_balance_inner(balance: &Balance, memory: &mut VmMemory) -> Result<Word, RuntimeError> {
```
xgreenx, line 95: If we are sure that `write_bytes` can't fail, then let's change the signature of the method too

```suggestion
    fn set_memory_balance_inner(balance: &Balance, memory: &mut VmMemory) ->  Word {
```

fuel-vm/src/interpreter/memory/operations.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub fn check_mem_access(&self, range: &MemoryRange) -> Result<(), RuntimeError> {
+        let stack = MemoryRange::try_new(0, self.registers[RegId::SP]).expect("Invalid SP value");
+        let heap = MemoryRange::try_new(self.registers[RegId::HP], VM_MAX_RAM - self.registers[RegId::HP])
+            .expect("Invalid HP value");
+
+        if range.len() > MEM_MAX_ACCESS_SIZE {
+            return Err(PanicReason::MemoryAccessSize.into());
+        }
+
+        if stack.contains_range(range) && heap.contains_range(range) {
```
xgreenx, line 128: How `range` can belong to the `heap` and to the `stack` simultaneously?

fuel-vm/src/interpreter/blockchain/smo_tests.rs
```diff
@@ -1,234 +0,0 @@
...
-        msg_data_ptr: 432,
-        msg_data_len: 10,
-        amount_coins_to_send: 20,
-        initial_balance: 20,
-        internal: true,
-        ..Default::default()
-    } => matches Ok(Output { external_balance: 20, internal_balance: 0, .. })
-    ; "spend all coins succesfully from internal context"
-)]
-fn test_smo(
```
Voxelot, line 182: why was this test removed? It has little to do with these memory changes and is more about verify how it affects runtime balances.

Voxelot:
Found a bug related to the issue I anticipated with this approach before we started implementing this. Because new pages are appended to the end of the heap buffer, the memory address translation isn't stable. 

I wrote a test which writes some data to the heap, logs it, allocates a new page, then tries to log the same data again. On master this test passes, but on this branch it fails.

Dentosal:
Nevermind

PR #465, opened 2023-05-31
Minor improvements during integration with `fuel-core`

xgreenx:
Added some helper functions.

### DIFF
--- a/fuel-tx/src/contract.rs
+++ b/fuel-tx/src/contract.rs
@@ -313,4 +313,15 @@ mod tests {
 
         assert_eq!(root, expected_root);
     }
+
+    #[test]
+    fn empty_contract_id() {
+        let contract = Contract::from(vec![]);
+        let salt = Salt::zeroed();
+        let root = contract.root();
+        let state_root = Contract::default_state_root();
+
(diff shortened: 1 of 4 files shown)


fuel-tx/src/contract.rs
```diff
@@ -32,6 +32,12 @@ fn next_multiple<const N: usize>(x: usize) -> usize {
 pub struct Contract(#[derivative(Debug(format_with = "fmt_truncated_hex::<16>"))] Vec<u8>);
 
 impl Contract {
+    /// The `ContractId` of the contract with empty bytecode, zero salt, and empty state root.
+    pub const EMPTY_CONTRACT_ID: ContractId = ContractId::new([
```
Voxelot, line 36: It would be nice if this was all calculated in a const block instead of being hardcoded. Is this already possible or are the fn's needed to generate this ID not easily usable in const? If so, we could consider using `lazy_static!` as well.

xgreenx: It is possible to constantly get `salt`, `root`, and `state_root`, but the hash function itself is not const. 

I added a unit test to verify the empty contract id.

Voxelot (approved):

PR #468, opened 2023-06-02
demonstrate bug in memory paging due to reversed order of memory on the heap

Voxelot:
This issue stems from the fact that allocations extend the heap array, but memory addressing is reversed.

Specifically this is related to how memory ranges are translated onto the heap:
Screen Shot 2023-06-01 at 5 34 47 PM

If new pages were prepended instead of appended to the heap array, this would likely work. Ie [`VecDeque::push_front`](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html#method.push_front)


Dentosal:
Should be fixed in 878bb396dbc04e976ec5ab683f6589b7fbbdf828. Added this test case to the original PR as well.

PR #467, opened 2023-06-01
feat: SMT update set

bvrooman:

PR #470, opened 2023-06-04
Add bulk SMT creation

xgreenx:
I came up with the idea of the algorithm. It creates a tree for 180ms instead of 1.4s (~8x faster) for 64k entries. So already is better than before=D

image

### DIFF
--- a/fuel-merkle/benches/smt.rs
+++ b/fuel-merkle/benches/smt.rs
@@ -223,23 +223,23 @@ fn sparse_merkle_tree(c: &mut Criterion) {
     let gen = || Some((random_bytes32(rng), random_bytes32(rng)));
     let data = std::iter::from_fn(gen).take(64000).collect::<Vec<_>>();
 
-    let l0 = Bytes32::default(); // left, left, left, left left, ...
-
-    let mut l1 = Bytes32::default();
-    l1[0..1].copy_from_slice(&[0b01000000]); // left, right, left, left, left, ...
-
-    let mut l3 = Bytes32::default();
-    l3[0..1].copy_from_slice(&[0b01001000]); // left, right, left, left, right, ...
-
-    let mut l2 = Bytes32::default();
-    l2[0..1].copy_from_slice(&[0b01100000]); // left, right, right, ...
-
-    let data = [
-        (l0, random_bytes32(rng)),
-        (l1, random_bytes32(rng)),
-        (l2, random_bytes32(rng)),
-        (l3, random_bytes32(rng)),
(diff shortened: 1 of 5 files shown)


bvrooman:
Nice job finding an algorithm. We can test it further with the existing SMT test suite to verify that it's correct.

xgreenx:
Yep, it was a brainstorm, and it works somehow. But I hope you can review it and make production ready(review the algorithm, integrate it into MerkleTree, and add more tests)=D

Or if it is not correct, maybe it will give you some ideas on how to implement that better.

bvrooman:
I reviewed your algorithm, and overall it is good!

 But I have two comments:

1. The algorithm assumes that the leaves will combine to form the root; this is not necessarily true. When you combine all the leaves, the result may be a node somewhere beneath the root. In this case, the algorithm returns the wrong root. I was able to create a test case that shows this failure:

```
    let l0 = Bytes32::default(); // left, left, left, left left, ...
    
    let mut l1 = Bytes32::default();
    l1[0..1].copy_from_slice(&[0b01000000]); // left, right, left, left, left, ...
    
    let mut l2 = Bytes32::default();
    l2[0..1].copy_from_slice(&[0b01100000]); // left, right, right, ...

    let mut l3 = Bytes32::default();
    l3[0..1].copy_from_slice(&[0b01001000]); // left, right, left, left, right, ...
    
    let data = [
        (l0, random_bytes32(rng)),
        (l1, random_bytes32(rng)),
        (l2, random_bytes32(rng)),
        (l3, random_bytes32(rng)),
    ];

    let input: BTreeMap<Bytes32, Bytes32> = BTreeMap::from_iter(data.into_iter());
    ...
```

Fortunately, that is easy to fix: We simply merge more placeholder nodes until we reach the root. I pushed a commit to fix that.

2. The algorithm uses a variable number of checks. Specifically, it checks `while !upcoming.is_empty() { ... }`, and the loop may pop from and push into `upcoming` multiple times. This is because it compares a given leaf or node to another leaf or node that is before or after it in the set to determine with whom it should be paired, and on which side. However, we can rely on some given knowledge: we have sorted the leaves by radix, and we can compare the paths of each leaf. With this information, we know how to pair the leaves/nodes, and on which one is the left and right. I think we can create a more deterministic loop with a smaller number of operations based on this information. 

Your algorithm gives a good starting point and I will be able to continue from here. Thanks for putting that together.

xgreenx:
Done in the https://github.com/FuelLabs/fuel-vm/pull/467

PR #475, opened 2023-06-05
chore: optimize smt common prefix count

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-core/issues/1143

This PR simplifies the calculation of the common prefix count between two byte arrays. This replaces the old method of reading each bit (and unwrapping) and comparing them. 

This is used during SMT calculations, such as `update` and `delete`, when calculating the "proximity" (how close two leaves/nodes are to each other).

### DIFF
--- a/fuel-merkle/src/common/msb.rs
+++ b/fuel-merkle/src/common/msb.rs
@@ -150,17 +150,26 @@ mod test {
 
     #[test]
     fn test_common_prefix_count_returns_count_of_common_bits_when_some_bits_match() {
-        let lhs_bytes: Bytes4 = [0b10101010, 0b10101010, 0b10101010, 0b10101010];
-        let rhs_bytes: Bytes4 = [0b10101010, 0b10101010, !0b10101010, 0b10101010];
+        let lhs_bytes: Bytes4 = [0b11111111, 0b11111111, 0b11111111, 0b11111111];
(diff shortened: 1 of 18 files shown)


fuel-merkle/src/common/msb.rs
```diff
@@ -39,12 +39,16 @@ impl<const N: usize> Msb for [u8; N] {
...
-                count += 1;
+        for (byte1, byte2) in self.iter().zip(other.iter()) {
+            let xor = byte1 ^ byte2;
+            if xor == 0 {
+                // The bytes are equal: Increment the prefix count by 8 bits
+                count += 8;
             } else {
+                // The bytes are not equal: Count the number of matching bits
+                // and increment the prefix count by this number
+                let leading_zeros = xor.leading_zeros();
```
bvrooman, line 50: Many architectures have intrinsics for counting leading zeroes, so this could be even faster than manual bit manipulation techniques.

Dentosal: I think this could be further simplified to

```rust
count += (byte1 ^ byte2).leading_zeros();
if byte1 != byte2 {
  break;
}
```

The compiler surely will figure it out.

bvrooman: Nice!

Dentosal (approved):

xgreenx (approved):

PR #479, opened 2023-06-08
Return the previous behavior of the data length GTF

xgreenx:
After https://github.com/FuelLabs/fuel-vm/pull/373 the GTF for data length is not working for `MessageCoinSigned` and `MessageCoinPredicate`


Dentosal (approved):
LGTM

Voxelot:
As a todo, we should ensure there is at least happy-path test coverage for each of these GTF flags.

PR #482, opened 2023-06-09
refactor: SMT path iter uses leaf key

### LINKED ISSUE #472: Sparse Merkel tree leafs/nodes may override other leafs/nodes
Sparse Merkel tree leafs/nodes may override other leafs/nodes because we use the same storage for the nodes and leafs.
Unit test [here](https://github.com/FuelLabs/fuel-vm/pull/469) shows the problem.

Thread from the slack:
@xgreenx : 

> Actually, regarding [this comment](https://github.com/FuelLabs/fuel-vm/pull/469#issuecomment-1574526210). In the [ink!](https://use.ink/) they decided to hash each storage key to preventing data structure manipulation(like Merkle Tree). Because the code is public and you know how it works inside, you can push the data to the contract and manipulate the internal state of the data structure. It may make them non-optimal and operations will take more time than expected.
>
> Maybe we want to think about it too. Because without it we have the bug like in [this](https://github.com/FuelLabs/fuel-vm/pull/469) PR=)

@bvrooman : 
> It sounds reasonable to hash storage keys. In general, it might be a good idea to avoid putting user input directly into the storage's primary keys like that.
>
> It sounds like we have two options:
> 1. Hash direct user input used as primary keys
> 1. Use separate storages for leaves and nodes2. 
>
> If option 1 is sufficient, that would be the quickest win. Option 2 is more involved, and may introduce other performance issues, but it allows us to avoid the extra hash and reduce the impact of collisions.

@xgreenx : 
> I think it is always good to have separate storage for leaves. In this case, we don’t need to worry about SMT, and because each storage has fewer entries, DB will work faster.
I think we definitely need to consider doing hashing from a security perspective. The problem is that we cannot iterate over the storage with this approach. But maybe we can check how substrate solved this problem(because I know they still can iterate over it even with hashing of the primary keys)

bvrooman:
Related issues:
 - Closes #472 

This PR removes the dependency on storage lookups for the leaf key during deletion. During deletion, we no longer need to look up leafs by leaf keys in storage because we can perform path iteration without the whole leaf: the leaf key is sufficient. This PR refactors the iterator to use a leaf key rather than the entire leaf.


bvrooman:
Converted back to draft, since I forgot to update the CHANGELOG. Will do that now!

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -124,15 +124,12 @@ where
...
-            // If the data is empty, this signifies a delete operation for the
-            // given key.
-            self.delete(key)?;
+            // Empty data signifies a no-op.
             return Ok(());
         }
 
         let leaf_node = Node::create_leaf(key, data);
         self.storage.insert(&leaf_node.hash(), &leaf_node.as_ref().into())?;
-        self.storage.insert(leaf_node.leaf_key(), &leaf_node.as_ref().into())?;
```
bvrooman, line 135: This was the line that introduced the potential for malicious collision

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -144,14 +141,15 @@ where
...
+    pub fn delete(&mut self, key: &Bytes32, data: &[u8]) -> Result<(), MerkleTreeError<StorageError>> {
         if self.root() == *Self::empty_root() {
             // The zero root signifies that all leaves are empty, including the
             // given key.
             return Ok(());
         }
 
-        if let Some(primitive) = self.storage.get(key)? {
+        let leaf_node = Node::create_leaf(key, data);
+        if let Some(primitive) = self.storage.get(&leaf_node.hash())? {
```
xgreenx, line 152: Actually, I checked the code, and it seems you don't even need a value. Because `self.path_set`  and `self.delete_with_path_set` uses only `leaf_key` inside the logic.

xgreenx (changes requested):

fuel-merkle/src/sparse/merkle_tree.rs
Voxelot: Does this mean we wouldn't need to pass the data here?

xgreenx: Yeap

bvrooman: Interesting, I can explore that further

bvrooman: Unfortunately, this doesn't work: We can't avoid the storage lookup here because we need to know if the deleted item is in the set of leaves or not. Whenever we call delete, we will modify all items along the leaf path. If we never added the item, we end up "reverting" a leaf path that was never actually changed, thus creating a false root. I got far enough into the changes in a PR here https://github.com/FuelLabs/fuel-vm/pull/483, but the tests with delete are failing for this reason.

We need to be able to short-circuit the delete call before modifying the nodes along the leaf path. As far as I am aware, the only way to do this is by checking the storage.

bvrooman: @xgreenx observed that we can check for the leaf by querying the path set, allowing us to properly check the delete 👍

xgreenx (dismissed):

bvrooman:
Hmmm, for whatever reason, `test_update_with_empty_data_performs_delete` still fails. I'm looking into it now. Still, we may not need this behaviour anyway, and we can likely stop supporting it.

xgreenx (approved):

PR #483, opened 2023-06-09
refactor: path iter with leaf key

bvrooman:


fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -134,27 +133,22 @@ where
...
 
-        let leaf_node = Node::create_leaf(key, data);
-        if let Some(primitive) = self.storage.get(&leaf_node.hash())? {
-            let primitive = primitive.into_owned();
-            let leaf_node: Node = primitive.try_into().map_err(MerkleTreeError::DeserializeError)?;
-            let (path_nodes, side_nodes): (Vec<Node>, Vec<Node>) = self.path_set(leaf_node.clone())?;
-            self.delete_with_path_set(&leaf_node, path_nodes.as_slice(), side_nodes.as_slice())?;
-        }
+        let (path_nodes, side_nodes): (Vec<Node>, Vec<Node>) = self.path_set(*key)?;
+        self.delete_with_path_set(key, path_nodes.as_slice(), side_nodes.as_slice())?;
```
bvrooman, line 151: The problem here is that we are always deleting along this path, even if there was never anything added. This is problematic because we will revert nodes along the path that weren't actually changed, causing our root to be incorrect.

bvrooman:
Nice fix!

PR #485, opened 2023-06-11
Changed the API of the `SparseMerkleTree` to accept `MerkleStorageKey`

xgreenx:
The `MerkleStorageKey` is hashed version of the raw storage key. Hashing prevents the caller from the SMT structure manipulations and removes the possibility of decreasing the performance of the SMT.

[Example of the usage in the `fuel-core`](https://github.com/FuelLabs/fuel-core/pull/1207)

It requires merging the change on the [specification side](https://github.com/FuelLabs/fuel-specs/pull/496) first.

It is a final issue on the VM side that we need to be able to close https://github.com/FuelLabs/fuel-core/issues/1143


fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -641,7 +641,7 @@ where
                 let (a, b, c, d) = call.unpack();
 
                 // Enter call context
-                self.prepare_call(r!(a), r!(b), r!(c), r!(d))?;
+                self.prepare_call(a, b, c, d)?;
```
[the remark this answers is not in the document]
xgreenx: Yep, I moved it to https://github.com/FuelLabs/fuel-vm/pull/487. It is here because without it the corresponding PR doesn't compile on `fuel-core` side=D

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -126,12 +177,16 @@ where
     TableType: Mappable<Key = Bytes32, Value = Primitive, OwnedValue = Primitive>,
     StorageType: StorageMutate<TableType, Error = StorageError>,
 {
-    pub fn from_set<I, D>(mut storage: StorageType, set: I) -> Result<Self, StorageError>
+    pub fn from_set<B, I, D>(mut storage: StorageType, set: I) -> Result<Self, StorageError>
     where
-        I: Iterator<Item = (Bytes32, D)>,
+        I: Iterator<Item = (B, D)>,
```
bvrooman, line 182: Would it be better to use `I: Iterator<Item = (MerkleTreeKey, D)>,`?

xgreenx: Yes, but it will break many places, so I moved this conversion into this method. It is okay that `from_set` accepts pure keys

CHANGELOG.md
```diff
@@ -17,17 +17,26 @@ Description of the upcoming release here.
 
 ### Fixed
 
-- [#482](https://github.com/FuelLabs/fuel-vm/pull/482): This PR address a security issue where updates to a Sparse Merkle Tree could deliberately overwrite existing leaves by setting the leaf key to the hash of an existing leaf or node. This is done by removing the insertion of the leaf using the leaf key.
+- [#482](https://github.com/FuelLabs/fuel-vm/pull/482): This PR address a security 
+    issue where updates to a Sparse Merkle Tree could deliberately overwrite existing
+    leaves by setting the leaf key to the hash of an existing leaf or node. This is 
+    done by removing the insertion of the leaf using the leaf key.
```
Voxelot, line 23:
```suggestion
    done by no longer inserting the leaf using the leaf key.
```

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -33,6 +33,57 @@ impl<StorageError> From<StorageError> for MerkleTreeError<StorageError> {
     }
 }
 
+/// The safe Merkle tree storage key prevents Merkle tree structure manipulations.
+/// The type contains only one constructor that hashes the storage key.
+#[derive(Debug, Clone, Copy)]
```
Voxelot, line 38: should we also include ops like `PartialEq` or `Display`? [derive_more](https://jeltef.github.io/derive_more/derive_more/) may be helpful for newtype ergonomics here.

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -33,6 +33,57 @@ impl<StorageError> From<StorageError> for MerkleTreeError<StorageError> {
...
+    pub fn new<B>(storage_key: B) -> Self
+    where
+        B: AsRef<[u8]>,
+    {
+        use digest::Digest;
+        let mut hash = sha2::Sha256::new();
+        hash.update(storage_key.as_ref());
+        let hash = hash
+            .finalize()
+            .try_into()
```
Voxelot, line 53: Do we really need a try_into here? I thought finalize returns a `Bytes32` already

Voxelot (dismissed):

fuel-merkle/src/sparse/merkle_tree.rs
Voxelot: We will also need to be able to bulk-insert pure keys for regenesis

fuel-merkle/src/sparse/merkle_tree.rs
bvrooman: `finalize` actually returns a `GenericArray` which we have to transform into a native array. We end up doing this `finalize().try_into()...` pattern in a few places because of this.

https://docs.rs/sha2/latest/sha2/trait.Digest.html#tymethod.finalize

Voxelot: ah ok, I assumed this was using the hasher from fuel-crypto

bvrooman (approved):

PR #489, opened 2023-06-13
Use nightly rustfmt with `.rustfmt.toml` from `fuel-core`

xgreenx:
It uses the `.rustfmt.toml` from https://github.com/FuelLabs/fuel-core/pull/1050 PR.


bvrooman (approved):
I didn't go over the whole diff, but I am trusting that the `fmt` gods did their job here. Everything still passes and nothing sticks out to me from a quick glance. In a couple of places, manual formatting for things like comments are out of place, but that's pretty minor. If we want this in the next release in time for the audit, I'm good to go with this change. But others should weigh in on this to decide if we will go with this now,  later, or forgo this entirely.

Voxelot (approved):
It's not really going to get any easier to make formatting changes as time goes on, so probably time to just rip the bandaid off now to improve our dev ux going forward.

PR #495, opened 2023-06-15
Some additional test to verify ecrecovery with `TxId` and coin witness

xgreenx:
@luizstacio is working on the predicate, and he tried to recover the signer public key from the witness and `TxId`. These tests verify the behavior for the script and predicate on the `fuel-vm` side.

### DIFF
--- a/fuel-vm/src/tests/crypto.rs
+++ b/fuel-vm/src/tests/crypto.rs
@@ -82,6 +82,141 @@ fn ecrecover() {
 
     assert!(success);
 }
+
+#[test]
+fn ecrecover_tx_id() {
+    let rng = &mut StdRng::seed_from_u64(2322u64);
+
+    let mut client = MemoryClient::default();
+
+    let gas_price = 0;
+    let gas_limit = 1_000_000;
+    let maturity = Default::default();
+    let height = Default::default();
+    let params = ConsensusParameters::default();
+    let gas_costs = GasCosts::default();
+
+    let secret = SecretKey::random(rng);
+    let public = secret.public_key();
+
+    #[rustfmt::skip]
+    let script = vec![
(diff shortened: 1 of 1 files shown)


fuel-vm/src/tests/crypto.rs
```diff
@@ -82,6 +82,141 @@ fn ecrecover() {
...
+    #[rustfmt::skip]
+    let script = vec![
+        // 0x21 is a address of the singer of the witness
+        op::gtf_args(0x20, 0x00, GTFArgs::ScriptData),
+        op::move_(0x21, 0x20),
+        // 0x22 is a witness - signature
+        op::gtf_args(0x22, 0x00, GTFArgs::WitnessData),
+        // TxId is stored in the first 32 bytes of the memory
+        // Store it into register 0x23
+        op::movi(0x23, 0),
```
Dentosal, line 111: Nit: registers are zeroed already on initialization

fuel-vm/src/tests/crypto.rs
```diff
@@ -82,6 +82,141 @@ fn ecrecover() {
...
+        // Allocate space for the recovered public key
+        // 0x10 contains the size of the public key = PublicKey::LEN
+        op::movi(0x10, PublicKey::LEN as Immediate18),
+        op::aloc(0x10),
+        op::move_(0x11, RegId::HP),
```
Dentosal, line 116: Nit: assigning to a register is not needed, `RegId::HP` could be used directly in place of `0x11` after this.

Dentosal (approved):
LGTM; I'm a bit surprised we didn't have tests of this already but it's good to add them now.

xgreenx:
It is to make the test case more clear=D

Issue #349, opened 2023-02-13
Improve the Usability of Op codes by adding argument names to op codes


freesig:
# Problem

The current Op codes use ra, rb, rc and rd as arguments, which can be confusing when using the ops.
It is hard to remember which argument is which when creating the op.

# Solution

Adding descriptive arguments such as sum, lhs, and rhs for `ADDI` to the Op codes will make them easier to understand and use.

# Example
```rust
impl_instructions! {
    "Adds two registers."
    0x10 ADDI addi [Sum: RegId, Lhs: RegId, Rhs: U12]
```
Which generates something like the following:
```rust
pub fn new(sum: RegId<u64>, lhs: RegId<u64>, rhs: Imm12<U12>) -> Self

pub fn sum(&self) -> RegId<u64>
pub fn lhs(&self) -> RegId<u64>
pub fn rhs(&self) -> Imm12<U12>
```
and a debug like:
```
ADDI { sum: RegId(16), lhs: RegId(17), rhs: U12(10) }
```

Voxelot:
This seems like it's not critical for audits

Dentosal:
Closeing this as completed. Feel free to open new issues for specific improvements.

PR #497, opened 2023-06-19
Add argument names to fuel-asm opcodes

Dentosal:
Work towards https://github.com/FuelLabs/fuel-vm/issues/349. This PR doesn't do externally any visible changes (except renaming some of the arguments), i.e. doesn't add new methods etc. yet. However, this completes hard part of the refactor, and allows more improvements in the future. It refactors the big `macro_rules! impl_instructions` block to smaller pieces so it's easier to modify in the future.


xgreenx (approved):
Nice job! Later we also need to update the actual places where we use these opcodes with the same naming, but it is another story=)

PR #498, opened 2023-06-28
Add push/pop instructions

Dentosal:
Experimenting with push/pop opcodes. See spec PR https://github.com/FuelLabs/fuel-specs/pull/499 and spec issue https://github.com/FuelLabs/fuel-specs/issues/407


fuel-vm/src/interpreter/memory.rs
```diff
@@ -287,6 +327,75 @@ where
...
+    segment: ProgramRegistersSegment,
+    bitmask: Imm24,
+) -> Result<(), RuntimeError> {
+    let bitmask = bitmask.to_u32();
+
+    // First compute the new stack pointer, as that's the only error condition
+    let count = bitmask.count_ones();
+    let stack_range = MemoryRange::new(*sp, (count as u64) * 8)?;
+    let new_sp = stack_range.words().end;
+    if new_sp > *hp {
```
Voxelot, line 345: If sp == hp then we have a collision. Maybe we could refactor some of the code used in `stack_pointer_overflow` to be reusable here?

```suggestion
    if new_sp >= *hp {
```

fuel-vm/src/interpreter/memory.rs
```diff
@@ -287,6 +327,75 @@ where
...
+    let bitmask = bitmask.to_u32();
+
+    // First compute the new stack pointer, as that's the only error condition
+    let count = bitmask.count_ones();
+    let size_in_stack = (count as u64) * 8;
+    let new_sp = sp
+        .checked_sub(size_in_stack)
+        .ok_or(PanicReason::MemoryOverflow)?;
+    if new_sp < *ssp {
+        return Err(PanicReason::MemoryOverflow.into())
```
Voxelot, line 380: Is this an overflow or MemoryOwnership error?

Dentosal: Overflow, since `MemoryOwnership` is would mean violation of the ownership rules. Stack and heap overlap isn't exactly that. Ideally we'd have another error here, called `OutOfMemory` or something like that. There are several other places where clarifying the panic reason would be nice, so I think leaving that to a follow-up PR would be the correct call in case we want to have that.

fuel-vm/src/interpreter/memory.rs
Dentosal: Fixed in 0e00edf6952dcfa2e779b1178e762c64d2f895ee. The current code style is poorly composable, so the result is a bit ugly, but at least they share the validation logic now.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -265,26 +315,104 @@ where
...
+) -> Result<(), RuntimeError> {
+    let bitmask = bitmask.to_u32();
+
+    // First update the new stack pointer, as that's the only error condition
+    let count = bitmask.count_ones();
+    let stack_range = MemoryRange::new(*sp, (count as u64) * 8)?;
+    try_update_stack_pointer(sp, ssp, hp, stack_range.words().end)?;
+
+    // Write the registers to the stack
+    let mut it = memory[stack_range.usizes()].chunks_exact_mut(8);
```
xgreenx, line 373: Is it possible to cross the heap here? Or is it not because we checked that inside of the `try_update_stack_pointer`?

fuel-vm/src/interpreter/memory.rs
```diff
@@ -265,26 +315,104 @@ where
...
+    for (i, reg) in program_regs.segment(segment).iter().enumerate() {
+        if (bitmask & (1 << i)) != 0 {
+            it.next().unwrap().copy_from_slice(&reg.to_be_bytes());
```
xgreenx, line 376: Could you use `expect`, please, with a description of why it is impossible?=)

fuel-vm/src/interpreter/memory.rs
```diff
@@ -265,26 +315,104 @@ where
...
+    pc: RegMut<PC>,
+    program_regs: &ProgramRegisters,
+    segment: ProgramRegistersSegment,
+    bitmask: Imm24,
+) -> Result<(), RuntimeError> {
+    let bitmask = bitmask.to_u32();
+
+    // First update the new stack pointer, as that's the only error condition
+    let count = bitmask.count_ones();
+    let stack_range = MemoryRange::new(*sp, (count as u64) * 8)?;
```
xgreenx, line 369:
```suggestion
    let stack_range = MemoryRange::new(*sp, count * 8)?;
```

fuel-vm/src/interpreter/memory.rs
```diff
@@ -265,26 +315,104 @@ where
...
+    pc: RegMut<PC>,
+    program_regs: &mut ProgramRegisters,
+    segment: ProgramRegistersSegment,
+    bitmask: Imm24,
+) -> Result<(), RuntimeError> {
+    let bitmask = bitmask.to_u32();
+
+    // First update the stack pointer, as that's the only error condition
+    let count = bitmask.count_ones();
+    let size_in_stack = (count as u64) * 8;
```
xgreenx, line 398:
```suggestion
    let size_in_stack = count * 8;
```

fuel-vm/src/interpreter/memory.rs
Dentosal: That will not work; `u32` doesn't implement `ToAddr` (unless we expand the implementation to cover it). The as-cast here is extending `u32` returned by `count_ones()`, and is complely harmless. I'm replacing it with a `.into()` cast anyway, so it looks less suspicious.

fuel-vm/src/interpreter/memory.rs
Dentosal: See above

xgreenx (approved):

PR #502, opened 2023-07-06
bug: Remove leaf data from BMT proofs

bvrooman:


fuel-merkle/src/binary/merkle_tree.rs
```diff
@@ -172,14 +172,6 @@ where
 
         let root_position = self.root_position();
         let leaf_position = Position::from_leaf_index(proof_index);
-        let primitive = self
-            .storage
-            .get(&leaf_position.in_order_index())?
-            .ok_or(MerkleTreeError::LoadError(proof_index))?
-            .into_owned();
-        let leaf_node = Node::from(primitive);
-        proof_set.push(*leaf_node.hash());
```
bvrooman, line 181: This code block is what added the hashed leaf data to the proof set. This is now removed, and the proof set no longer includes the leaf data.

fuel-merkle/test-helpers/src/binary/merkle_tree.rs
```diff
@@ -88,7 +89,14 @@ impl MerkleTree {
             current = current.take_next().unwrap();
         }
 
-        (self.root(), self.proof_set)
+        let root = self.root();
+
+        // The proof set starts with the hashed leaf data; remove the leaf data
+        // in order to conform with the proof set specification.
+        let mut proof_set = VecDeque::from(self.proof_set);
+        proof_set.pop_front();
```
bvrooman, line 97: This removes the hashed leaf data from the `test-helpers` reference version of the BMT proof generation. The reference Merkle tree algorithm used here is different than the one used in production. Because of this, we can only remove the leaf data from the proof after the proof is generated.

xgreenx (approved):

fuel-merkle/test-helpers/src/binary/verify.rs
```diff
@@ -1,42 +1,44 @@
 use crate::binary::{
```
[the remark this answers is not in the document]
bvrooman: I think that reusing the function here is worth considering. 
As part of that consideration, some factors include: 
- Do we have a production use case for `fuel_merkle::binary::verify` yet? If not, it might actually make more sense to keep only `test-helpers::binary::verify` instead. `test-helpers::binary::verify` is used specifically to verify proofs during tests (i.e. check the proofs are generated correctly), but `fuel_merkle::binary::verify` doesn't have a "real" use case, so it is currently just a maintenance burden.
- Can we simplify or improve the implementation of `fuel_merkle::binary::verify`? If so, we can use `test-helpers::binary::verify` to quickly check the correctness of any refactors to the production version.

PR #504, opened 2023-07-10
Bugfixes found during chatting with Auditors on July 7th

xgreenx:
Unified the check that the `ContractId` in inputs by `TouchedContracts` structure.
Check that the contract is in inputs for `CROO` and `CSIZ`. 
Fixed the bug with `total_code_size`.
Fixed the bug with incorrect gas charging per unit.

### DIFF
--- a/fuel-vm/src/interpreter.rs
+++ b/fuel-vm/src/interpreter.rs
@@ -461,3 +461,27 @@ impl CheckedMetadata for CreateCheckedMetadata {
         self.gas_used_by_predicates = gas_used;
     }
 }
+
+pub(crate) struct TouchedContracts<'vm, I> {
+    input_contracts: I,
+    panic_context: &'vm mut PanicContext,
+}
+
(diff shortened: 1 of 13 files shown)


fuel-vm/src/interpreter.rs
```diff
@@ -461,3 +461,27 @@ impl CheckedMetadata for CreateCheckedMetadata {
         self.gas_used_by_predicates = gas_used;
     }
 }
+
+pub(crate) struct TouchedContracts<'vm, I> {
```
bvrooman, line 465: I think the idea of centralizing the location where we check the list of input contracts is a good abstraction. However, I think `touch` is a bit of a misleading name. In the context of software, `touch` usually means to create a file/modify its timestamp (see https://www.geeksforgeeks.org/touch-command-in-linux-with-examples/). Here, you're using `touch` to panic if the file doesn't exist.
What about calling it something like `InputContracts<..>` and changing `touch` to something like `check`?

Voxelot (dismissed):

Voxelot (approved):

PR #505, opened 2023-07-11
Removed the `data` field from canonical serialization and deserialization of receipts

xgreenx:
- The `data` field is optional now and is not recovered during canonical deserialization. 
- The `len` field is not padded anymore.

ref https://github.com/FuelLabs/fuel-core/issues/1240


MitchTurner:
LGTM! Can I ask what was wrong with the padded `len`?

Voxelot (approved):

xgreenx:
Padding only `len` without padding `data` means that the `digest` also is incorrect because `digest` doesn't match the `len`. Also, the specification doesn't say that we need to pad it.

PR #511, opened 2023-07-14
Return correct PanicReason on memory-related panics

### LINKED ISSUE #38: Test memory boundaries for opcode execution
We are to use words as pairs of encoded instructions.

However, jump instructions are half-word. And we also might interrupt a program (e.g. debug) in a half-word

This means that incrementing by word might lead to invalid memory access in case the code is in the memory boundary.

We need to have a test case covering that - and, if applicable, fix the code so no panic occur.

Dentosal:
Closes #38. Refactors the error handling code of multiple functions to leverage `MemoryRange` even more. Changes multiple panic reasons to be more accurate.


fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -617,10 +616,6 @@ where
 
     pub(crate) fn run_program(&mut self) -> Result<ProgramState, InterpreterError> {
         loop {
-            if self.registers[RegId::PC] >= VM_MAX_RAM {
```
xgreenx, line 620: How do we know that it is not possible? Were you able to check all the places modifying the PC register?

Also maybe worth to put it under `debug_assert` condition for future changes

fuel-vm/src/interpreter/memory.rs
```diff
@@ -294,15 +319,9 @@ pub(crate) fn load_byte(
...
-    let bc = b.saturating_add(c) as usize;
-
-    if bc >= VM_MAX_RAM as RegisterId {
-        Err(PanicReason::MemoryOverflow.into())
-    } else {
-        *result = memory[bc] as Word;
-
-        inc_pc(pc)
-    }
+    let range = MemoryRange::new_overflowing_op(Word::overflowing_add, b, c, 1u64)?;
```
xgreenx, line 322:
```suggestion
    let range = MemoryRange::new_overflowing_op(Word::overflowing_add, b, c, 0u64)?;
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -668,28 +657,29 @@ impl<'vm, S, I> CodeCopyCtx<'vm, S, I> {
...
         let contract = CheckedMemConstLen::<{ ContractId::LEN }>::new(b)?;
-        let cd = checked_add_word(c, d)?;
 
-        if d > MEM_MAX_ACCESS_SIZE
-            || a > checked_sub_word(VM_MAX_RAM, d)?
-            || cd > VM_MAX_RAM
-        {
-            return Err(PanicReason::MemoryOverflow.into())
-        }
+        MemoryRange::new(a, d)?;
```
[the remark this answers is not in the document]
Dentosal: It would be nice, but it's possible to invalidate the checks before creating a range and using it, still imposing the responsibility to the caller. So, I think it's not that big of an improvement. https://github.com/FuelLabs/fuel-vm/pull/524 takes some minor steps to making things like this cleaner, and I'm hoping to continue to that direction in a future PR.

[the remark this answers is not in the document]
Dentosal: `try_memory...` still does the full ownership checks. The reason we do the check here as well is to preserve the panic reason precedence. Not checking it here would make `ContractNotFound` trigger before the ownership checks. Maybe that's not an issue, but I didn't want to break any backwards compatibility needlessly.

fuel-vm/src/interpreter/memory.rs
Dentosal: We are writing a single byte, why should this be `0`? That would mean an empty range, which then would not include the byte we're writing to.

fuel-vm/src/interpreter/executors/main.rs
Dentosal: Even better, I reworked the instruction fetching so that `Interpreter::execute` checks the memory boundary when doing an instruction fetch. This means we can close https://github.com/FuelLabs/fuel-vm/issues/38 as well.

fuel-vm/src/interpreter/crypto.rs
```diff
@@ -186,21 +182,10 @@ pub(crate) fn keccak256(
...
-
-    if a > checked_sub_word(VM_MAX_RAM, Bytes32::LEN as Word)?
-        || c > MEM_MAX_ACCESS_SIZE
-        || bc > MIN_VM_MAX_RAM_USIZE_MAX
-    {
-        return Err(PanicReason::MemoryOverflow.into())
-    }
-
-    let (a, b, bc) = (a as usize, b as usize, bc as usize);
+    let src_range = MemoryRange::new(b, c)?;
```
Voxelot, line 185: What happened to the check for `MEM_MAX_ACCESS_SIZE`?

Dentosal: `MIN_VM_MAX_RAM_USIZE_MAX` is not a meaningful constant to have. If `VM_MAX_RAM > usize::MAX`, the VM will not work anyways with the current memory handling; it's not possible to allocate more than `usize::MAX` bytes of memory. It should be removed completely in a follow-up PR.

As for `MEM_MAX_ACCESS_SIZE`, the check was likely a mistake. The tricky part here is that it's impossible to catch this with tests, as the configuration value for it is `MEM_MAX_ACCESS_SIZE = VM_MAX_RAM`. The rule against exceeding `VM_MAX_RAM` always trigger first. Before this PR, the `MaxMemoryAccess` error was completely unused. [I asked about this in slack](https://fuellabs.slack.com/archives/C01L1HUJSL8/p1689365109352789), and after re-reading the discussion now, I feel like we're removing the it completely.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -43,7 +43,7 @@ pub trait ToAddr {
 
 impl ToAddr for usize {
     fn to_addr(self) -> Result<usize, RuntimeError> {
-        if self >= MEM_SIZE {
+        if self > MEM_SIZE {
```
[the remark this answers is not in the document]
Dentosal: This is only used by `MemoryRange`, which in turn checks that we wont do that. This is only here to allow a zero-length range at the end of the VM memory.

fuel-vm/src/interpreter/crypto.rs
Dentosal: And both removed, in 84cfbca and f0e5367.

fuel-vm/src/interpreter/blockchain.rs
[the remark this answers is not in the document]
Dentosal: I don't know whether it matters which PanicReason is given as long as it's consistent, though, and the code certainly would be simpler if we didn't care.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -487,39 +479,28 @@ impl<'vm, S, I> LoadContractCodeCtx<'vm, S, I> {
...
-        let contract_id: &[u8; ContractId::LEN] = &self.memory
-            [contract_id..contract_id_end]
-            .try_into()
-            .expect("This can't fail, because we checked the bounds above.");
-
-        // Safety: Memory bounds are checked and consistent
-        let contract_id = ContractId::from_bytes_ref(contract_id);
-
-        self.input_contracts.check(contract_id)?;
+        self.input_contracts.check(&contract_id)?;
```
Voxelot, line 500: maybe we should perform this validation before filling the memory with zeros? It gives us a way to bail out of execution earlier before performing a potentially large memset.

fuel-vm/src/interpreter/blockchain.rs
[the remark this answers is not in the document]
Voxelot: The order of panic checks shouldn't really matter much as long as we can unit test the code to verify each panic condition.

Voxelot (approved):

xgreenx (approved):

PR #514, opened 2023-07-19
Decompose consensus params

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1243

The idea behind this PR was that `ConsensusParameters` was being passed to any function that needed a value in `ConsensusParameters`. We wanted the functions to define what they needed instead of giving them the whole kitchen sink.

The work included:
1. Removing `ConsensusParams` from all function signatures
1. Adding the required fields back to the function signatures
1. Grouping parameters that are used together into sub structs: e.g. `TxParameters`, `FeeParameters`
1. Fixing tests
1. Fixing tests
1. Fixing tests
1. Combining fields of large function signatures


MitchTurner:
In addition to tests not passing, There are a number of function signatures I'd like to improve before merging this. I'll try to document what those changes are here so we can track them.

fuel-tx/src/transaction/types/create.rs
```diff
@@ -237,25 +242,29 @@ impl FormatValidityChecks for Create {
     fn check_without_signatures(
         &self,
         block_height: BlockHeight,
-        parameters: &ConsensusParameters,
+        tx_params: &TxParameters,
```
xgreenx, line 245: If we need a combination of the parameters, it is better to pass only `ConsensusParameters`. It is not a problem if some fields are not used. The API should be simple for the user and transparent regarding requirements(like you don't need to pass `ConsensusParameters` if you only need a `ChainId`).

MitchTurner: For sure. I've combined them back together in some other places. Clippy just didn't catch this because it's at the limit (7). I can do the same here. Good catch.

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -178,27 +239,107 @@ impl Default for ConsensusParameters {
...
+        PredicateParameters::DEFAULT.max_gas_per_predicate;
+    pub const GAS_PRICE_FACTOR: u64 = FeeParameters::DEFAULT.gas_price_factor;
+    pub const GAS_PER_BYTE: u64 = FeeParameters::DEFAULT.gas_per_byte;
+
+    pub const CHAIN_ID: ChainId = ChainId::new(0);
+}
+
+/// A collection of parameters for convenience
+#[derive(Debug, Clone, Copy)]
+pub struct ConsensusParams<'a> {
```
MitchTurner, line 282: I can't help but think this should live in `fuel-vm` rather than here since nothing in `fuel-tx` uses this.

fuel-vm/src/checked_transaction.rs
```diff
@@ -186,40 +191,91 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
     fn into_checked(
         self,
         block_height: BlockHeight,
-        params: &ConsensusParameters,
-        gas_costs: &GasCosts,
+        consensus_params: ConsensusParams<'_>,
```
xgreenx, line 194: Do you want to move `GasCosts` to `ConsensusParams` in a separate PR? Because it was an initial reason for this PR(have one type that includes all data required for consensus verification)=D

fuel-vm/src/checked_transaction.rs
```diff
@@ -186,40 +191,91 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
...
+        chain_id: ChainId,
+        gas_costs: GasCosts,
     ) -> Result<Checked<Self>, CheckError>
     where
         Checked<Self>: CheckPredicates,
     {
-        self.into_checked_basic(block_height, params)?
-            .check_signatures(&params.chain_id)?
-            .check_predicates(params, gas_costs)
+        let check_predicate_params = CheckPredicateParams {
```
xgreenx, line 201: It is better to implement `From<ConsensusParams>` trait

fuel-vm/src/checked_transaction.rs
```diff
@@ -186,40 +191,91 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
...
 /// Provides predicate verification functionality for the transaction.
 #[async_trait::async_trait]
 pub trait CheckPredicates: Sized {
     /// Performs predicates verification of the transaction.
-    fn check_predicates(
-        self,
-        params: &ConsensusParameters,
-        gas_costs: &GasCosts,
-    ) -> Result<Self, CheckError>;
+    fn check_predicates(self, params: CheckPredicateParams) -> Result<Self, CheckError>;
```
xgreenx, line 273: I prefer `&CheckPredicateParams` more than `CheckPredicateParams` because we usually store this configuration somewhere. Plus, downstream code decides whether it needs to clone or not.

fuel-vm/src/interpreter.rs
```diff
@@ -109,7 +112,14 @@ pub struct Interpreter<S, Tx = ()> {
     balances: RuntimeBalances,
     gas_costs: GasCosts,
     profiler: Profiler,
-    params: ConsensusParameters,
+    fee_params: FeeParameters,
```
xgreenx, line 115: I guess you want to use `InterpreterParams` here=)

fuel-vm/src/interpreter.rs
```diff
@@ -181,6 +191,12 @@ impl<S, Tx> Interpreter<S, Tx> {
     pub const fn profiler(&self) -> &Profiler {
         &self.profiler
     }
+
+    /// Modify the FeeParameters
+    pub fn with_fee_params(mut self, fee_params: FeeParameters) -> Self {
```
xgreenx, line 196: It should be specified during creation of the `Interpreter`

fuel-vm/src/interpreter/debug.rs
```diff
@@ -47,12 +47,20 @@ where
...
     let gas_limit = 1_000_000;
     let height = Default::default();
-    let params = ConsensusParameters::default();
+    let tx_params = TxParameters::default();
+    let predicate_params = PredicateParameters::default();
+    let script_params = ScriptParameters::default();
+    let contract_params = ContractParameters::default();
+    let fee_params = FeeParameters::default();
+    let chain_id = ChainId::default();
+    let gas_costs = GasCosts::default();
```
xgreenx, line 63: It adds a lot of overhead=) if `ConsensusParams` works with values instead, you don't need to do it

fuel-vm/src/util.rs
```diff
@@ -449,11 +503,17 @@ pub mod test_helpers {
             &mut self,
             checked: Checked<Create>,
         ) -> anyhow::Result<StateTransition<Create>> {
-            let mut transactor = Transactor::new(
-                self.storage.clone(),
-                self.params,
-                self.gas_costs.clone(),
-            );
+            let interpreter_params = InterpreterParams {
```
xgreenx, line 506: It is better if `Self` contains `InterpreterParams` instead=)

fuel-tx/src/tests/offset.rs
```diff
@@ -466,7 +469,8 @@ fn iow_offset() {
             let bytes = tx.to_bytes();
 
             let mut tx_p = tx.clone();
-            tx_p.precompute(&ConsensusParameters::DEFAULT.chain_id)
+            let chain_id = ChainId::new(0);
+            tx_p.precompute(&chain_id)
```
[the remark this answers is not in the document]
MitchTurner: Done.

fuel-tx/src/tests/valid_cases/input.rs
```diff
@@ -66,10 +70,10 @@ fn input_coin_message_signature() {
...
             f(&mut tx, &public);
 
-            tx.sign_inputs(&secret, &ConsensusParameters::DEFAULT.chain_id);
-            keys.iter().for_each(|sk| {
-                tx.sign_inputs(sk, &ConsensusParameters::DEFAULT.chain_id)
-            });
+            let chain_id = ChainId::new(0);
+
+            tx.sign_inputs(&secret, &chain_id);
+            keys.iter().for_each(|sk| tx.sign_inputs(sk, &chain_id));
```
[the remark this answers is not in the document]
MitchTurner: Done.

fuel-tx/src/tests/valid_cases/input.rs
```diff
@@ -648,7 +772,11 @@ fn transaction_with_duplicate_contract_inputs_is_invalid() {
         .add_output(o)
         .add_output(p)
         .finalize()
-        .check_without_signatures(Default::default(), &Default::default())
+        .check_without_signatures(
```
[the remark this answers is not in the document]
MitchTurner: This is a trait method. It gets used in the other impls of this method.

But it sounds like you want `chain_id` as part of `ConsensusParams` so I think it will get removed in that case anyway.

fuel-tx/src/transaction/consensus_parameters.rs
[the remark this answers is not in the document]
MitchTurner: It does make it harder to work with.

I had reasons I preferred references, but I'm not really convinced by them anymore, so I'll just get rid of the refs. Especially since all of them are `Copy`.

fuel-vm/src/checked_transaction.rs
MitchTurner: Okay. So, `GasCosts` is included on the `TransactionBuilderExt`:
```
fn finalize_checked(
        &mut self,
        height: BlockHeight,
        gas_costs: GasCosts,
    ) -> Checked<Tx>;
```

Which implies that, even though the transaction builder might hold the consensus params, it might not hold the `GasCosts`. 

That might just be an artifact and it can hold the `GasCosts` too.

MitchTurner: I'll try that. 

I'm not really sure why the `TransactionBuilderExt` trait exists at all. I don't see anything that depends on it. I wonder why it can't just be a method on the `TransactionBuilder` directly.

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -178,27 +239,107 @@ impl Default for ConsensusParameters {
...
+    pub tx_params: &'a TxParameters,
+    pub predicate_params: &'a PredicateParameters,
+    pub script_params: &'a ScriptParameters,
+    pub contract_params: &'a ContractParameters,
+    pub fee_params: &'a FeeParameters,
+}
+
+impl<'a> ConsensusParams<'a> {
+    /// Constructor for the `ConsensusParams` with Standard values
+    pub fn standard() -> Self {
```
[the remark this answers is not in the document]
MitchTurner: Hmm. Just added `GasCosts` as well. `GasCosts` doesn't have a `pub const DEFAULT`, it relies on the `default` which isn't `const`. So I don't think we can make `standard` const.

fuel-vm/src/util.rs
MitchTurner: I added the `ConsesensusParams` to `Self` and used the `into`.

fuel-vm/src/interpreter/debug.rs
MitchTurner: I missed this one. Adding now.

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -5,15 +5,197 @@ use fuel_types::{
...
+    GasUnit,
+};
+
 const MAX_GAS: u64 = 100_000_000;
 
+/// A collection of parameters for convenience
+#[derive(Debug, Clone)]
+pub struct ConsensusParams {
+    // TODO: This should be pub(crate) for the `fuel_tx` tests, but for some reason the
+    // tests can't see the fields with `pub(crate)`
```
Dentosal, line 23: Integration tests cannot see `pub(crate)` items, as they are outside the crate. Typically you'd do a public re-export behind `test-helpers` feature flag, but I don't see much point in not having this as a part of the public interface anyways.

fuel-vm/src/transactor.rs
```diff
@@ -122,18 +127,18 @@ where
...
-    /// Consensus parameters
-    pub const fn params(&self) -> &ConsensusParameters {
-        self.interpreter.params()
-    }
-
+    // /// Consensus parameters
+    // pub const fn params(&self) -> &ConsensusParameters {
+    //     self.interpreter.params()
+    // }
+    //
```
Dentosal, line 134: Commented-out code that likely should be removed

Dentosal (changes requested):
One todo and some commented out code should be cleaned up, but other than that LGTM

fuel-tx/src/transaction/consensus_parameters.rs
MitchTurner: Yeah. That's a good point. There are probably places outside of `fuel-tx` that could use this as well. I was referring to something in `fuel-tx/src/tests`, not in the integ tests.

I usually prefer getters/setters over accessing fields directly, but I'm not going to worry about it right now.

Dentosal (approved):

PR #524, opened 2023-07-25
Fix a crash in CCP and zero-filling issues in CCP and LDC

### LINKED ISSUE #523: Crash in CCP instruction
CCP is not checking that the given range `c..c+d` is in range for the contract. Crash here: https://github.com/FuelLabs/fuel-vm/blob/20f1e9b888e359b9d1ac45214673e97277370543/fuel-vm/src/interpreter/blockchain.rs#L692

Needs a check and a test case.

Dentosal:
Fixes #523. ~~Partial overlap with #511, but the order doesn't matter much.~~

Addresses the following issues:
* CCP instruction could crash when given out-of-contract-bounds input.
* CCP instruction would only zero memory when asked to copy more bytes than the contract has. This is a bug.
* LDC instruction would revert when asked more to load more bytes than the contract has. The spec requires zero-filling here.

As a side not, the logic around `$fp->code_size` seems quite complicated and it looks like it's not used for anything. Maybe we could remove it?


MitchTurner:
Could you give me the expected behavior for each of the bugs this PR is addressing? For example "out-of-contract-bounds input" sounds like it should result in unhappy behavior, if not "crash" what should it do?

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -487,67 +485,46 @@ impl<'vm, S, I> LoadContractCodeCtx<'vm, S, I> {
...
-        let memory_offset = ssp as usize;
-        let memory_offset_end = checked_add_usize(memory_offset, length)?;
+        let memory_offset = ssp;
 
         // Validate arguments
-        if memory_offset_end >= *self.hp as usize
-            || contract_id_end as Word > VM_MAX_RAM
-            || length > MEM_MAX_ACCESS_SIZE as usize
-            || length > self.contract_max_size as usize
+        if memory_offset.saturating_add(length) > *self.hp
```
Voxelot, line 497: Why did this change from >= hp to > hp? Since this variable is only used in a method that says no_ownership_checks, wouldn't that mean this could allow a stack overflow?

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -673,30 +650,26 @@ impl<'vm, S, I> CodeCopyCtx<'vm, S, I> {
         I: Iterator<Item = &'vm ContractId>,
         S: InterpreterStorage,
     {
-        let contract = CheckedMemConstLen::<{ ContractId::LEN }>::new(b)?;
-        let cd = checked_add_word(c, d)?;
-
-        if d > MEM_MAX_ACCESS_SIZE
-            || a > checked_sub_word(VM_MAX_RAM, d)?
-            || cd > VM_MAX_RAM
-        {
```
Voxelot, line 682: Should we also relabel the registers from a b c d to meaningful names like LDC above?

fuel-vm/src/interpreter/memory.rs
```diff
@@ -140,6 +140,13 @@ impl MemoryRange {
         Self(start..end)
     }
 
+    /// Splits range at given relative offset. Panics if offset > range length.
+    pub fn split_at_offset(self, at: usize) -> (Self, Self) {
```
Voxelot, line 144: Maybe we should return an optional return type here instead of using a panicking assert?

fuel-vm/src/tests/blockchain.rs
```diff
@@ -607,26 +607,170 @@ fn ldc_contract_not_in_inputs() {
...
+        vec![
+            op::movi(0x10, 2048),
+            op::aloc(0x10),
+            op::movi(0x20, data_offset as Immediate18),
+            op::add(0x11, RegId::ZERO, 0x20),
+            op::movi(0x12, (contract_size + 1) as Immediate18),
+            op::movi(0x13, contract_size as Immediate18),
+            op::ccp(RegId::HP, 0x11, 0x12, 0x13),
+            op::addi(0x21, 0x20, ContractId::LEN as Immediate12),
+            op::meq(0x30, 0x21, RegId::HP, 0x13),
```
Voxelot, line 751: Instead of checking if the result of ccp is != contract bytecode, maybe we should have a positive check that the ccp result zero filled as expected? Otherwise this test could still pass even if ccp just returned a bunch of garbage when the offset is > contract length.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -673,30 +650,26 @@ impl<'vm, S, I> CodeCopyCtx<'vm, S, I> {
...
 
-        let (a, c, d) = (a as usize, c as usize, d as usize);
-        let cd = cd as usize;
-
-        let contract = ContractId::from_bytes_ref(contract.read(self.memory));
+        self.input_contracts.check(&contract_id)?;
 
-        self.input_contracts.check(contract)?;
+        let contract =
+            super::contract::contract(self.storage, &contract_id)?.into_owned();
```
xgreenx, line 663: It seems we can remove `.into_owned()` because we don't do that for LDC

fuel-vm/src/interpreter/memory.rs
xgreenx: It seems you use this function only in one place, so maybe it is better to move this code there instead.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -581,3 +588,23 @@ pub(crate) fn write_bytes<const COUNT: usize>(
...
+    memory: &mut [u8; MEM_SIZE],
+    src: &[u8],
+    dst_addr: A,
+    src_offset: usize,
+    len: B,
+) -> Result<(), RuntimeError> {
+    let range = MemoryRange::new(dst_addr, len)?;
+
+    let src_end = range.len().min(src.len());
+    let data = src.get(src_offset..src_end).unwrap_or_default();
```
xgreenx, line 604: If the size of the `src`(contact) is `1000`, `len` is `8`, and `offset` is `999`, then you will have `999..8` instead of `999..1000`.

Maybe it is better to add tests for this function to cover all cases.

Also, do we need to worry about padding and alignment in the memory here?

fuel-vm/src/interpreter/memory.rs
Dentosal: The API here mirrors [`slice::split_at` from `core`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at), including the panic behavior. The code is written here instead of the call site since it's a useful abstraction, and IMO makes the code more readable.

fuel-vm/src/interpreter/blockchain.rs
Dentosal: This was fixed and cleaned up in #511. I merged `master` to this branch just now.

fuel-vm/src/interpreter/blockchain.rs
Dentosal: Done in 4323732

fuel-vm/src/interpreter/blockchain.rs
Dentosal: Done in df03c07

fuel-vm/src/interpreter/memory.rs
Dentosal: Fixed and added tests in 6c5313e. This function doesn't care about padding, it must be computed before it wanted. Alignment is not a concept FuelVM uses, so I'm not sure what you're asking here.

fuel-vm/src/tests/blockchain.rs
Dentosal: Good catch. Fixed in 0c3bb19.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -649,36 +629,40 @@ struct CodeCopyCtx<'vm, S, I> {
...
     pub(crate) fn code_copy(
         mut self,
-        a: Word,
-        b: Word,
-        c: Word,
-        d: Word,
+        dst_addr: Word,
+        contract_id_addr: Word,
+        contract_offset: Word,
+        length: Word,
```
xgreenx, line 635: Do we need to worry about padding here? Or is it the responsibility of the `CCP` user?

xgreenx (approved):
Looks good to me=) Thank you for fixing that! @Voxelot Could you have a look please one more time?

fuel-vm/src/interpreter/blockchain.rs
Voxelot: hmm it's not in the specs. I guess since this isn't meant to be executable code it's not so relevant here.

Voxelot (approved):

PR #527, opened 2023-07-26
Follow-up clean ups and fixed for the VM initialization

xgreenx:
Some follow-up improvements for the https://github.com/FuelLabs/fuel-vm/pull/514.

Also, during fixing, it found that the VM was initialized incorrectly for the predicate case. The balances should be zero.

image

### DIFF
--- a/fuel-tx/src/transaction/consensus_parameters.rs
+++ b/fuel-tx/src/transaction/consensus_parameters.rs
@@ -32,13 +32,26 @@ pub struct ConsensusParameters {
 
 impl Default for ConsensusParameters {
     fn default() -> Self {
-        Self::standard(ChainId::default())
+        Self::standard()
     }
 }
 
 impl ConsensusParameters {
-    /// Constructor for the `ConsensusParameters` with Standard values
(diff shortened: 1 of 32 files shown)


fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -32,13 +32,26 @@ pub struct ConsensusParameters {
...
+            script_params: ScriptParameters::DEFAULT,
+            contract_params: ContractParameters::DEFAULT,
+            fee_params: FeeParameters::DEFAULT,
+            chain_id: ChainId::default(),
+            gas_costs: GasCosts::default(),
+        }
+    }
+
+    /// Constructor for the `ConsensusParameters` with Standard values around `ChainId`.
+    pub fn standard_with_id(chain_id: ChainId) -> Self {
```
bvrooman, line 54: Would a builder pattern be more idiomatic here? I.e., something like:
```rust
let parameters = ConsensusParameters::standard().chain_id(chain_id);
```

bvrooman (approved):

PR #529, opened 2023-07-26
Initializing WASM synchronously

arboleya:
Ensures that the WASM binary is initialized [synchronously](https://github.com/rollup/plugins/tree/master/packages/wasm/#synchronous-modules) so that it is available right after being imported.


xgreenx (approved):

arboleya:
I saw that warning but went with it since everything worked consistently on my tests. As you suggested, I will run more in-depth tests at the file level and report back here. 

If we have to force async initialization, we'll need something like:

```ts
asm.init().then(() => asm.movi(...))
```

@xgreenx Please hold the release a bit more, so I can get to the bottom of this.

PR #531, opened 2023-07-27
Fix a crash in UtxoId::from_str and TxPointer::from_str with multibyte characters

### LINKED ISSUE #521: Decoding of UtxoId with multibyte characters can panic
The following query panics:

```graphql
{ 
 coin(utxoId:  "0x00😎"  )  {  utxoId  } 
}
```

This is caused by careless use of string indexing here:

https://github.com/FuelLabs/fuel-vm/blob/651351f74b0b3f112fd9676fefed9e077be71ee6/fuel-tx/src/transaction/types/utxo_id.rs#L104-L118

We should probably use [`#[deny(clippy::string_slice)]`](https://rust-lang.github.io/rust-clippy/master/index.html#/string_slice) to avoid these issues in the future.

Dentosal:
Fixes #521. Adds clippy lints to disallow raw string slicing, so this wont happen again so easily.

### DIFF
--- a/fuel-tx/src/lib.rs
+++ b/fuel-tx/src/lib.rs
@@ -4,6 +4,7 @@
 // Wrong clippy convention; check
 // https://rust-lang.github.io/api-guidelines/naming.html
 #![allow(clippy::wrong_self_convention)]
+#![deny(clippy::string_slice)]
 #![deny(unused_crate_dependencies)]
 #![deny(unsafe_code)]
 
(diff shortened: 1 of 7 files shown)


fuel-tx/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 // Wrong clippy convention; check
 // https://rust-lang.github.io/api-guidelines/naming.html
 #![allow(clippy::wrong_self_convention)]
+#![deny(clippy::string_slice)]
```
Voxelot, line 7: should we also add this to the `fuel-types` lib.rs?

Voxelot (approved):

fuel-tx/src/lib.rs
Dentosal: I'll do that as a follow-up, as I'm alreaady doing parsing changes as a part of #533

PR #533, opened 2023-07-27
Use custom derive ser/de instead of std::io

### LINKED ISSUE #443: Add `no_std` canonical serialization and deserialization
The initial PR: https://github.com/FuelLabs/fuel-tx/pull/190

With our own canonical serialization/deserialization we can avoid the usage of `io::Read`/`io::Write` and be able to compile with `no_std` mode.

Also, it can remove code duplication and simplify the support of serialization for types.

Dentosal:
Closes #443. Initial work migrated from https://github.com/FuelLabs/fuel-tx/pull/190, and the improved and adapted to the current system.

This PR replaces hand-written type serialization code with a macro-derived ones. It removes a lot of error-prone boilerplate code and makes it easier to add new types. The old implementation using `io::{Read, Write}` was not `no_std` compatible, but the new generated code is. The new implmentation also allows computing many sizes at compile time.

The new serialization framework is called `canonical`. It's used similarly to `serde`:

```rust
use fuel_types::canonical::{Serialize, Deserialize};

#[derive(Serialize, Deserialize)]
pub struct ExampleStruct {
    inner: InnerStruct,
    dynamically_sized: Vec<u8>,
    primitive: u64,
    #[canonical(ignore)]
    primitive: u64,
}
```

Several datatype-level attributes are also accepted:, expecially for enums.

### structs

https://github.com/FuelLabs/fuel-vm/blob/4af7c6312561576498acdca2f62c2235c94ce2bd/fuel-derive/src/attribute.rs#L59-L61

### enums

https://github.com/FuelLabs/fuel-vm/blob/4af7c6312561576498acdca2f62c2235c94ce2bd/fuel-derive/src/attribute.rs#L81-L95


fuel-asm/src/lib.rs
```diff
@@ -6,7 +6,7 @@
 #![deny(clippy::string_slice)]
 #![deny(missing_docs)]
 #![deny(unsafe_code)]
-#![deny(unused_crate_dependencies)]
+// #![deny(unused_crate_dependencies)]
```
Voxelot, line 9: Can we re-enable this now that the impl work is done?

fuel-derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "fuel-derive"
+version = "0.20.0"
```
Voxelot, line 3:
```suggestion
version = { workspace = true }
```

fuel-derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+authors = ["Fuel Labs <contact@fuel.sh>"]
+categories = ["cryptography::cryptocurrencies", "data-structures"]
+edition = "2021"
+homepage = "https://fuel.network/"
+keywords = ["blockchain", "cryptocurrencies", "fuel-vm", "vm"]
+license = "Apache-2.0"
+repository = "https://github.com/FuelLabs/fuel-tx"
```
Voxelot, line 10: version = { workspace = true }
authors = { workspace = true }
categories = ["cryptography::cryptocurrencies", "data-structures"]
edition = { workspace = true }
homepage = { workspace = true }
keywords = ["blockchain", "cryptocurrencies", "fuel-vm", "vm"]
license = "Apache-2.0"
repository = { workspace = true }

fuel-derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+description = "FuelVM (de)serialization derive macros for `fuel-tx`."
```
Voxelot, line 11:
```suggestion
description = "FuelVM (de)serialization derive macros for `fuel-vm` data structures."
```

fuel-derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[lib]
+proc-macro = true
+
+[dependencies]
+quote = "1"
+syn = { version = "2", features = ["full"] }
+proc-macro2 = "1"
+synstructure = "0.13"
```
Voxelot, line 20: newline
```suggestion
synstructure = "0.13"

```

fuel-tx/Cargo.toml
```diff
@@ -14,13 +14,16 @@ description = "FuelVM transaction."
 derivative = { version = "2.2.0", default-features = false, features = ["use_core"] }
 fuel-asm = { workspace = true, default-features = false }
 fuel-crypto = { workspace = true, default-features = false }
+fuel-derive = { path = "../fuel-derive" }
```
Voxelot, line 17: This should be a workspace dep
```suggestion
fuel-derive = { workspace = true }
```

fuel-tx/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 // https://rust-lang.github.io/api-guidelines/naming.html
 #![allow(clippy::wrong_self_convention)]
 #![deny(clippy::string_slice)]
-#![deny(unused_crate_dependencies)]
+// #![deny(unused_crate_dependencies)]
```
Voxelot, line 8: can we re-enable this?

fuel-tx/src/transaction/types/create.rs
```diff
@@ -208,8 +195,8 @@ impl Chargeable for Create {
     }
 }
 
+#[cfg(feature = "std")]
```
Voxelot, line 198: There are some areas in the code below which are affected by this change:
image

fuel-types/Cargo.toml
```diff
@@ -11,7 +11,9 @@ repository = { workspace = true }
 description = "Atomic types of the FuelVM."
 
 [dependencies]
+fuel-derive = { path = "../fuel-derive" }
```
Voxelot, line 14: Should use a workspace dep

fuel-types/src/canonical.rs
```diff
@@ -0,0 +1,619 @@
...
+    fn decode_dynamic<I: Input + ?Sized>(&mut self, buffer: &mut I) -> Result<(), Error> {
+        for _ in 0..self.capacity() {
+            // Bytes - Vec<u8> it a separate case without unpadding for each element.
+            // It should unpadded at the end if is not % ALIGN
+            if T::UNALIGNED_BYTES {
+                let byte = buffer.read_byte()?;
+                // SAFETY: `UNALIGNED_BYTES` implemented set for `u8`.
+                let _self =
+                    unsafe { ::core::mem::transmute::<&mut Vec<T>, &mut Vec<u8>>(self) };
+                _self.push(byte);
```
Voxelot, line 382: why do we take the first byte and append it at the end?

fuel-types/src/lib.rs
```diff
@@ -3,11 +3,16 @@
...
-#![deny(unused_crate_dependencies)]
+// #![deny(unused_crate_dependencies)]
+
+// `fuel-derive` requires `fuel_types` import
+extern crate self as fuel_types;
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
 extern crate core;
+#[cfg(any(feature = "alloc", feature = "std"))]
```
Voxelot, line 14: isn't alloc always enabled by std? Seems like there should be a way to simplify this pattern.

fuel-derive/src/serialize.rs
```diff
@@ -0,0 +1,495 @@
...
+    s.add_bounds(synstructure::AddBounds::Fields)
+        .underscore_const(true);
+
+    let serialize = match s.ast().data {
+        syn::Data::Struct(_) => serialize_struct(&s),
+        syn::Data::Enum(_) => serialize_enum(&s),
+        _ => panic!("Can't derive `Serialize` for `union`s"),
+    };
+
+    // crate::utils::write_and_fmt(
```
Voxelot, line 488: Should this be removed?

fuel-types/src/canonical.rs
```diff
@@ -0,0 +1,619 @@
...
+        let len: u64 = self.len().try_into().expect("msg.len() > u64::MAX");
+        len.encode(buffer)
+    }
+
+    fn encode_dynamic<O: Output + ?Sized>(&self, buffer: &mut O) -> Result<(), Error> {
+        // Bytes - Vec<u8> it a separate case without padding for each element.
+        // It should padded at the end if is not % ALIGN
+        if T::UNALIGNED_BYTES {
+            // SAFETY: `UNALIGNED_BYTES` only set for `u8`.
+            let bytes = unsafe { ::core::mem::transmute::<&Vec<T>, &Vec<u8>>(self) };
```
Voxelot, line 347: Do we have any safety checks to ensure we aren't transmuting heap pointers here like `Box` etc?

fuel-types/src/canonical.rs
```diff
@@ -0,0 +1,619 @@
...
+impl SizeCalculator {
+    /// The number of written bytes.
+    pub fn size(self) -> usize {
+        self.0
+    }
+}
+
+impl Output for SizeCalculator {
+    fn write(&mut self, bytes: &[u8]) -> Result<(), Error> {
+        self.0 = self
```
Voxelot, line 508: How does this compare to the previous approach of using the memlayout + some manual logic for dynamic sizes? Are we incurring any new overhead by first traversing all serializable fields, converting each one into a slice, and then storing each of their lengths here?

Voxelot: If the process of converting a field into a `&[u8]` has overhead, then we end up paying this twice for every call to `to_bytes()` since we get the size first to initialize a vec with enough capacity.

Voxelot: Maybe we should consider calculating the static size of each type at compile time?

fuel-vm/src/constraints.rs
```diff
@@ -61,11 +64,10 @@ impl<T> CheckedMemValue<T> {
     /// Inspect a value of type `T` from memory.
     pub fn inspect(self, memory: &[u8; MEM_SIZE]) -> T
     where
-        T: std::io::Write + Default,
+        T: Deserialize + Default,
```
xgreenx, line 67:
```suggestion
        T: Deserialize,
```

fuel-vm/src/call.rs
```diff
@@ -5,29 +5,23 @@ use fuel_asm::{
...
     *,
 };
-use std::io::{
-    self,
-    Write,
-};
 
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
```
xgreenx, line 24:
```suggestion
#[derive(Deserialize, Serialize)]
```

fuel-vm/src/call.rs
```diff
@@ -75,81 +62,15 @@ impl Call {
...
-
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    fuel_types::canonical::Deserialize,
+    fuel_types::canonical::Serialize,
```
xgreenx, line 72:
```suggestion
    Deserialize,
    Serialize,
```

fuel-types/src/lib.rs
xgreenx: The main idea is that `canonical` s/d should be always available, we don't need to use `std` or `alloc` features.

fuel-types/src/canonical.rs
```diff
@@ -0,0 +1,619 @@
...
+impl_for_fuel_types!(Address);
+impl_for_fuel_types!(AssetId);
+impl_for_fuel_types!(ContractId);
+impl_for_fuel_types!(Bytes4);
+impl_for_fuel_types!(Bytes8);
+impl_for_fuel_types!(Bytes20);
+impl_for_fuel_types!(Bytes32);
+impl_for_fuel_types!(MessageId);
+impl_for_fuel_types!(Salt);
+impl_for_fuel_types!(Nonce);
```
xgreenx, line 237: You don't need to implement it manually. You can derive it because these types are defined in the same crate.

During the previous implementation, I defined traits in the `fuel-tx` which is why I should implemented them manually.

fuel-types/src/canonical.rs
xgreenx: We iterate over all bytes(we use `capacity`, not length to do that because we have an empty vector with a pre-defined capacity) and add them to the current vector. Everything is expected=)

fuel-types/src/canonical.rs
xgreenx: The `Box` works with raw pointers, so the memory layout for any `T` is the same. But maybe it is better to work directly with `&[u8]` to avoid any question in the future=)

fuel-types/src/canonical.rs
xgreenx: It is the safest way to know the final size of the object. But yeah, we can optimize it later by introducing `STATIC_PART_SIZE`.

I'm okay with this during the first iteration of this transition.

Dentosal: Static size *is* computed into `SIZE_STATIC` constant. The transition away from `SizeCalculator´ shoulnd't be too hard. I left that for a follow-up PR, but it might make sense to do it right now.

fuel-types/src/lib.rs
Voxelot: I think alloc might be required since the serialization uses vec

fuel-types/src/canonical.rs
Dentosal: Migrated to derived impl in 34e6a048

fuel-tx/src/transaction/consensus_parameters/gas.rs
```diff
@@ -1,7 +1,10 @@
 //! Tools for gas instrumentalization
 
-use alloc::sync::Arc;
 use core::ops::Deref;
+
+#[cfg(feature = "alloc")]
```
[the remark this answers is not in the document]
Dentosal: Yep. Done in df32bbb8. The CI isn't catching this, but I'm unsure whether adding even more feature/crate combinations is a good idea regarding speed.

fuel-types/src/canonical.rs
Dentosal: `T::UNALIGNED_BYTES` is only set for `u8` as the comment says. Types are also serialized bottom-up so this shouldn't be too much of an issue anyway. I'm still considering removing the `unsafe` usage, since I'm not sure that the performance impact will be meaningful.

fuel-derive/src/serialize.rs
Dentosal: Maybe? That's the main way to debug derive macros, so maybe I'll rather just clean it up a bit and document what it does instead.

fuel-types/src/lib.rs
Dentosal: I did more fine-grained gating, so it should be fixed now. Currently `alloc` is enabled by `std`, so maybe that could be simplified even more.

fuel-vm/src/constraints.rs
Dentosal: Done in df32bbb8

fuel-vm/src/call.rs
Dentosal: Done in f329eacf

fuel-vm/src/call.rs
Dentosal: Done in f329eacf

fuel-types/src/canonical.rs
Dentosal: Yep, this is indeed correct

fuel-derive/src/attribute.rs
```diff
@@ -0,0 +1,102 @@
...
+use syn::{
+    AttrStyle,
+    Attribute,
+    Meta,
+};
+use synstructure::BindingInfo;
+
+/// This holds top-level `canonical` attributes for a struct or an enum.
+#[derive(Debug)]
+pub struct TypedefAttrs(pub HashMap<String, TokenStream>);
```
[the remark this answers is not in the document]
Dentosal: I totally agree here. Sadly Rust doesn't have reflection, and manually writing code for all possible fields is not something I like. Maybe just a list of allowed attributes plus duplicate checking would be good enough.

fuel-asm/Cargo.toml
```diff
@@ -26,7 +27,8 @@ rstest = "0.16"
 [features]
 default = ["std"]
 typescript = ["wasm-bindgen", "wee_alloc"]
-std = ["serde?/default"]
+std = ["serde?/default", "fuel-types/std"]
+alloc = []
```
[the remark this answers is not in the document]
Dentosal: Not needed anymore, I think. I'll remove it.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- [#533](https://github.com/FuelLabs/fuel-vm/pull/533): Use custom serliazation for fuel-types to allow no_std compilation.
```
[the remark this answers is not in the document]
Dentosal: True.

fuel-tx/src/transaction/repr.rs
```diff
@@ -1,10 +1,13 @@
...
+    IntoPrimitive,
+    TryFromPrimitive,
+};
 
-#[cfg(feature = "std")]
-use fuel_types::Word;
+use crate::Transaction;
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, IntoPrimitive, TryFromPrimitive)]
```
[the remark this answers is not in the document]
Dentosal: I've never had any issues, and it would be interesting to know what issues we had.

CHANGELOG.md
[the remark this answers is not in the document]
Dentosal: Fixed in 912a2786

fuel-types/src/canonical.rs
Dentosal: It did make sense to do right now, as it required so much changes.

fuel-tx/src/transaction/types.rs
```diff
@@ -19,7 +19,7 @@ pub use utxo_id::UtxoId;
 pub use witness::Witness;
 
 #[cfg(feature = "std")]
-pub fn compute_transaction_id<T: fuel_types::bytes::SerializableVec + Clone>(
+pub fn compute_transaction_id<T: fuel_types::canonical::Serialize + Clone>(
```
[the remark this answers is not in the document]
Dentosal: Done in 5bbdb451

fuel-tx/src/transaction/types.rs
```diff
@@ -19,7 +19,7 @@ pub use utxo_id::UtxoId;
 pub use witness::Witness;
 
 #[cfg(feature = "std")]
```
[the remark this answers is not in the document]
Dentosal: Done in 5bbdb451

fuel-tx/src/transaction/consensus_parameters/gas.rs
```diff
@@ -141,13 +144,15 @@ impl GasUnit {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 /// Gas costings for every op.
 /// The inner values are wrapped in an [`Arc`]
 /// so this is cheap to clone.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg(any(feature = "alloc", feature = "std"))]
```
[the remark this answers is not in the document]
Dentosal: Done in 5bbdb451

fuel-asm/Cargo.toml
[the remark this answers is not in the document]
Dentosal: Removed in a24f9a7a1742d3043b446250034447343b92e9c9

fuel-derive/src/deserialize.rs
```diff
@@ -0,0 +1,193 @@
...
+            *buffer = buf;
+        }
+    } else {
+        quote! {
+            let raw_discr = <::core::primitive::u64 as fuel_types::canonical::Deserialize>::decode(buffer)?;
+        }
+    };
+
+    s.gen_impl(quote! {
+        gen impl fuel_types::canonical::Deserialize for @Self {
```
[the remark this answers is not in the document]
Dentosal: Done in 635dc638cc362ac52e79dcd7252f2e543fee9e62

fuel-derive/src/attribute.rs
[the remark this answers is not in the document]
Dentosal: Done in cf1870d4

fuel-derive/src/deserialize.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/// Derives `Deserialize` trait for the given `struct` or `enum`.
+pub fn deserialize_derive(mut s: synstructure::Structure) -> TokenStream2 {
+    s.bind_with(|_| synstructure::BindStyle::RefMut)
+        .add_bounds(synstructure::AddBounds::Fields)
+        .underscore_const(true);
+
+    match s.ast().data {
+        syn::Data::Struct(_) => deserialize_struct(&mut s),
```
bvrooman, line 207: Does this automatically support tuple structs as well?

fuel-derive/src/serialize.rs
```diff
@@ -0,0 +1,551 @@
...
+
+    // #[repr(int)] types have a known static size
+    let impl_size = if let Some(repr) = repr {
+        // Repr size, already aligned up
+        let repr_size: usize = match repr.as_str() {
+            "u8" => 8,
+            "u16" => 8,
+            "u32" => 8,
+            "u64" => 8,
+            "u128" => 16,
```
bvrooman, line 237: It looks like every unsigned integer type is going to be serialized to take up 8 bytes (except 16 for `u128`) - is that correct? Is this part of the specification? 
Maybe using constants here, with a mention of the spec, would make this a little more clear and a little more flexible in case the specification ever changes.

fuel-derive/src/utils.rs
```diff
@@ -0,0 +1,47 @@
...
+        Command,
+        Stdio,
+    },
+    sync::OnceLock,
+};
+
+static WORKSPACE: OnceLock<PathBuf> = OnceLock::new();
+
+pub fn workspace_dir() -> &'static PathBuf {
+    WORKSPACE.get_or_init(|| {
```
bvrooman, line 20: Nice use of the new `OnceLock` feature

bvrooman (commented):
I looked at about 35 files. I had some questions/minor suggestions for your discretion. I can keep reviewing later, but don't wait for me if you've already gotten a full review from the others.

fuel-derive/src/deserialize.rs
Dentosal: Yes. See syn documentation here: https://docs.rs/syn/latest/syn/enum.Data.html

fuel-derive/src/attribute.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// The struct is prefixed with the given byte.
+    /// Useful with`#[canonical(inner_discriminant)]`.
+    pub prefix: Option<TokenStream>,
+}
+
+impl StructAttrs {
+    pub fn parse(s: &synstructure::Structure) -> Self {
+        let mut attrs = parse_attrs(s);
+
+        let prefix = attrs.remove("prefix");
```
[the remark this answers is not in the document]
Dentosal: See field documentation for prefix:

https://github.com/FuelLabs/fuel-vm/blob/e048864f1a4b41e41b2f66f494614a6794e04bc0/fuel-derive/src/attribute.rs#L59-L61

So in practice it's used something like this:

```rust
#[derive(Serialize, Deserialize)]
#[canonical(prefix = 3u64)]
pub struct Prefixed3 {
  a: u64,
  b: u64,
}
```

Now `Prefixed3 { a: 5, b: 7 }` serializes to:

```
00 00 00 00 00 00 00 03 // prefix
00 00 00 00 00 00 00 05 // a
00 00 00 00 00 00 00 07 // b
```

This is especially useful with `#[canonical(inner_discriminant = PrefixType)]`:

```rust
#[derive(Serialize, Deserialize, TryFromPrimitive, IntoPrimitive)]
pub enum Discriminant {
  P1 = 1,
  P2 = 2,
}

#[derive(Serialize, Deserialize)]
#[canonical(prefix = Discriminant::P1)]
pub struct Prefixed1 {
  a: u64,
  b: u64,
}

#[derive(Serialize, Deserialize)]
#[canonical(prefix = Discriminant::P2)]
pub struct Prefixed2 {
  a: u64,
  data: Vec<u8>,
}


#[derive(Serialize, Deserialize]
#[canonical(inner_discriminant = Discriminant)]
pub enum Example {
  P1(Prefixed1),
  P2(Prefixed2),
}
```

Now when serializing `Example`, the discriminant isn't written to output, and instead the inner structs `Prefixed{1,2}` perform the write on serialization. When deserializing, the prefix is read to determine which variant to deserialize, but not consumed by the enum. The variant itself then re-reads the prefix.

This makes `Example` to not take more space than any of it's variants, when it can identify the variant by it's prefix instead. This is similar tricks that can be done with i.e. `union` in `C`. For our purposes this is useful since some structures must always start with a given identifier, and we would still like to warp over them using `enum`. I'm not sure if there's actually any reason those types are prefixed with their identifier, but I'd rather just replicate the old behavior exactly in this PR, and consider any changes in serialization behavior in follow-up PRs.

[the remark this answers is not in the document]
bvrooman: Great, thanks for the clear explanation. Is the prefix variable length or is it required to be a certain length, i.e. byte?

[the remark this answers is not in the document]
Dentosal: It's exactly one word, yes. It could be changed, but since that's the only use case now, I decided to keep it simple.

bvrooman (commented):
Can we now remove the `mem_layout` macro and related code in this PR?

fuel-tx/src/transaction/types/input/coin.rs
```diff
@@ -135,216 +130,64 @@ where
...
-        n += size;
-        if let Some(predicate_data_field) = self.predicate_data.as_mut_field() {
-            *predicate_data_field = predicate_data;
+        Coin {
+            utxo_id,
+            owner,
+            amount,
+            asset_id,
+            tx_pointer,
+            witness_index: (),
```
bvrooman, line 180: Can this line be replaced with `..Default::default()` for consistency with your changes below?

fuel-tx/src/transaction/types/input/message.rs
```diff
@@ -204,216 +197,64 @@ where
...
+        Message {
+            sender,
+            recipient,
+            amount,
+            nonce,
+            witness_index,
+            data,
+            predicate: (),
+            predicate_data: (),
+            predicate_gas_used: (),
```
bvrooman, line 221: Can these be replaced with `..Default::default()`?

bvrooman:
I've reviewed 66 files at this point - I will continue with the last stretch tomorrow.

fuel-derive/src/attribute.rs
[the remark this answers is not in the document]
Voxelot: Maybe a crate like [darling](https://crates.io/crates/darling) would be helpful?

fuel-derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
```
Voxelot, line 1: We should also have a README.md for this crate that include status badges for ci and crates.io

fuel-derive/manual_impl_tx_serde.rs
```diff
@@ -0,0 +1,77 @@
+impl fuel_types::canonical::SerializedSize for Transaction {
```
Voxelot, line 1: How does this file get used? I can't find it being imported anywhere with a quick `rg` search over the files.

fuel-derive/src/attribute.rs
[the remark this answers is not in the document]
Voxelot: According to the spec, aren't the prefixes generally supposed to be a u8? If word size prefixes matches how we serialized things before, that seems like a bug in the spec implementation.

image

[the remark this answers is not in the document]
Voxelot: Word size prefixes makes sense, given this:
image

fuel-types/src/canonical.rs
```diff
@@ -0,0 +1,688 @@
...
+
+        #[derive(Debug, PartialEq, Eq, TryFromPrimitive, IntoPrimitive)]
+        #[repr(u64)]
+        enum TestEnum3Discriminant {
+            P1 = 1,
+            P2 = 2,
+        }
+
+        #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+        #[canonical(inner_discriminant = TestEnum3Discriminant)]
```
Voxelot, line 650: should we also have a test for regular a non-inner discriminant, ie. like we use on Input?

```rust
#[canonical(discriminant = InputRepr)]
```

fuel-derive/manual_impl_tx_serde.rs
Voxelot: if it's just for local testing, maybe add a comment explaining what it's for

Voxelot:
What happens when an enum uses `#[canonical(inner_discriminant = Type)]`, but the corresponding structs of each variant don't use `#[canonical(prefix = Type::P1)]`? If we always need to specify the prefix along with discriminant, should we have some kind of check at compile time to ensure it's setup correctly?

fuel-types/src/canonical.rs
Dentosal: Yes. Fortunately it was even easier than that; I simply removed it completely, since it was only used when manually deserializing anyway. 1e14e956

fuel-derive/manual_impl_tx_serde.rs
Dentosal: It's not needed anymore. Removed.

fuel-derive/Cargo.toml
Dentosal: Added in b16268cb

fuel-tx/src/transaction/types/input/message.rs
Dentosal: 79d57179

fuel-tx/src/transaction/types/input/coin.rs
Dentosal: 79d57179

Voxelot:
Good call on removing discriminant, it was a little confusing how it was suppose to be used compared to inner_descriminant.

Voxelot (approved):
Wrote some local fuzz tests to check the ability of roundtrip serialization between this and the original encoding, and didn't find any issues. I think we're probably good to ship this.

Issue #510, opened 2023-07-14
Add more test cases for `test_transfer_output` and `test_transfer` tests


xgreenx:
The current test case set is poor(only one test case per unit test). We need at least to test internal and external context behaviors. But also, maybe we can find more test cases during more deep dive.

xgreenx:
Closed by https://github.com/FuelLabs/fuel-vm/pull/535

PR #535, opened 2023-07-28
Add more test cases for `transfer` and `transfer_output`

MitchTurner:
https://github.com/FuelLabs/fuel-vm/issues/510

### DIFF
--- a/fuel-vm/src/interpreter/contract.rs
+++ b/fuel-vm/src/interpreter/contract.rs
@@ -304,9 +304,6 @@ impl<'vm, S, Tx> TransferCtx<'vm, S, Tx> {
         };
 
         if let Some(source_contract) = internal_context {
-            dbg!(&source_contract);
-            dbg!(&asset_id);
-            dbg!(&amount);
             // debit funding source (source contract balance)
             balance_decrease(self.storage, &source_contract, &asset_id, amount)?;
         } else {
@@ -316,7 +313,6 @@ impl<'vm, S, Tx> TransferCtx<'vm, S, Tx> {
 
         // credit variable output
         let variable = Output::variable(to, amount, asset_id);
-        dbg!(&variable);
 
         set_variable_output(self.tx, self.memory, self.tx_offset, out_idx, variable)?;
 
--- a/fuel-vm/src/interpreter/contract.rs
+++ b/fuel-vm/src/interpreter/contract.rs
@@ -51,6 +51,7 @@ use fuel_types::{
     ContractId,
 };
 
+use crate::interpreter::memory::read_bytes;
 use std::borrow::Cow;
 
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


MitchTurner:
The existing test for for `transfer_output` was simply wrong. It just happened to work because so many of the test parameters were the same :(

fuel-vm/src/interpreter/contract.rs
```diff
@@ -365,6 +345,63 @@ impl<'vm, S, Tx> TransferCtx<'vm, S, Tx> {
     }
 }
 
+fn get_contract_id_from_memory(
```
xgreenx, line 348: You can actually replace all new functions with `CheckedMemConstLen` and `read`, it should be shorter and more readable and will unify the code

image

fuel-vm/src/interpreter/contract/tests.rs
```diff
@@ -53,19 +55,49 @@ fn test_contract_balance(b: Word, c: Word) -> Result<(), RuntimeError> {
...
+#[test_case(false, 0, 70, 32 => Err(RuntimeError::Recoverable(PanicReason::NotEnoughBalance)); "Cannot transfer from internal balance insufficient funds")]
+fn test_transfer(
+    external: bool,
+    contract_id_offset: Word,
+    transfer_amount: Word,
+    asset_id_offset: Word,
+) -> Result<(), RuntimeError> {
+    // Given
+
+    const ASSET_ID: [u8; AssetId::LEN] = [2u8; AssetId::LEN];
```
xgreenx, line 69: You can create the type itself. The same is applicable to `ContractId` and `Adress`=) And instead of `&ASSET_ID[..]` you can use `ASSET_ID.as_ref()`

```suggestion
    const ASSET_ID: AssetId = AssetId::new([2u8; AssetId::LEN]);
```

fuel-vm/src/interpreter/contract/tests.rs
```diff
@@ -53,19 +55,49 @@ fn test_contract_balance(b: Word, c: Word) -> Result<(), RuntimeError> {
...
 #[test_case(true, 0, 50, 32 => Ok(()); "Can transfer from external balance")]
-fn test_transfer(external: bool, a: Word, b: Word, c: Word) -> Result<(), RuntimeError> {
+#[test_case(false, 0, 50, 32 => Ok(()); "Can transfer from internal balance")]
+#[test_case(true, 0, 70, 32 => Err(RuntimeError::Recoverable(PanicReason::NotEnoughBalance)); "Cannot transfer from external balance insufficient funds")]
+#[test_case(false, 0, 70, 32 => Err(RuntimeError::Recoverable(PanicReason::NotEnoughBalance)); "Cannot transfer from internal balance insufficient funds")]
+fn test_transfer(
+    external: bool,
+    contract_id_offset: Word,
+    transfer_amount: Word,
+    asset_id_offset: Word,
```
xgreenx, line 65: It would be nice if you tested something with `contract_id_offset` and `asset_id_offset` arguments. Otherwise, it is strange why do we have them=D

fuel-vm/src/interpreter/contract.rs
MitchTurner: Went with
```
fn get_address_from_memory(
    offset: Word,
    memory: &[u8; MEM_SIZE],
) -> Result<Address, RuntimeError> {
    get_from_memory::<{ Address::LEN }, Address>(offset, memory)
}

fn get_from_memory<const SIZE: usize, T>(
    offset: Word,
    memory: &[u8; MEM_SIZE],
) -> Result<T, RuntimeError>
where
    T: From<[u8; SIZE]>,
{
    let range = CheckedMemConstLen::<SIZE>::new(offset)?;
    let bytes = range.read(memory);
    let value = T::from(*bytes);
    Ok(value)
}
```
Which I'm happy with.

BUT

I wish that `LEN` was a associated `const` for some trait and then we could let the type checker fill in the generics for us:
```
fn get_from_memory<T>(
    offset: Word,
    memory: &[u8; MEM_SIZE],
) -> Result<T, RuntimeError>
where
    T: WithSize
    T: From<[u8; <T as WithSize>::SIZE]>,
{
    let range = CheckedMemConstLen::<<T as WithSize>::SIZE>::new(offset)?;
    let bytes = range.read(memory);
    let value = T::from(*bytes);
    Ok(value)
}
```

I would add a trait for it, but then we'd need to import that trait everywhere we want the `LEN` and that's not worth it.

fuel-vm/src/interpreter/contract/tests.rs
MitchTurner: I added some simple tests that swap them. I think that's good enough because it shows that it actually uses memory. There are infinite other tests we could write including the offsets and none of them seem particularly interesting off the top of my head.

fuel-vm/src/interpreter/contract.rs
Dentosal: I find this even cleaner, as it's modular yet still readable:

```rust
let contract_id = ContractId::from(read_bytes(self.memory, contract_id_addr)?);
````

Comment (MitchTurner) [reply fuel-vm/src/interpreter/contract.rs:349]: Agreed. I added that change. I'm keeping the helper functions though because I prefer my code to read declaratively.

Comment (xgreenx) [reply fuel-vm/src/interpreter/contract.rs:349]: Hmm, but I think it is already descriptive enough: "Read bytes from memory with offset as ContractId"=)

```rust
let contract_id = ContractId::from(read_bytes(memory, offset)?);
```

fuel-vm/src/interpreter/contract/tests.rs
xgreenx: It would be nice to test some bad cases like memory overflow=)

fuel-vm/src/interpreter/contract.rs
MitchTurner: Okay. 2v1. I'll change it.

xgreenx (approved):

PR #546, opened 2023-08-09
Instruction debug fmt improvements

Dentosal:
Currently the `reason.instruction` sub-field of `PanicInstruction` is just a raw integer representation of the instruction bytes. This PR changes the `Debug` implementation to decode the instruction when possible. It will also show the bytes in hexadecimal format, which is more useful for debugging as it can be used to spot the instruction in a hexdump, and it shows the opcode byte separately.

This PR also changes the debug format of register (non-immediate) operands to be hexadecimal, as it is the established convention (e.g. the specification lists registers using hex values, and do does most of the test code). 

### Before

```rust
reason: PanicInstruction {
    reason: OutOfGas,
    instruction: 1920991264,
},
```

```rust
reason: PanicInstruction {
    reason: ErrorFlag,
    instruction: 0,
},
```

### After

```rust
reason: PanicInstruction {
    reason: OutOfGas,
    instruction: MOVI { dst: 0x20, val: 32 } (bytes 72 80 00 20),
}
```

```rust
reason: PanicInstruction {
    reason: ErrorFlag,
    instruction: Unknown (bytes 00 00 00 00),
},
```

### DIFF
--- a/fuel-asm/src/panic_instruction.rs
+++ b/fuel-asm/src/panic_instruction.rs
@@ -1,3 +1,5 @@
+use core::fmt;
+
 use crate::{
     Instruction,
     PanicReason,
@@ -5,7 +7,7 @@ use crate::{
     Word,
 };
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "typescript", wasm_bindgen::prelude::wasm_bindgen)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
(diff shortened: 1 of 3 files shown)


fuel-asm/src/panic_instruction.rs
```diff
@@ -35,6 +37,35 @@ impl PanicInstruction {
...
+/// Helper struct to debug-format a `RawInstruction` in `PanicInstruction::fmt`.
+struct InstructionDbg(RawInstruction);
+impl fmt::Debug for InstructionDbg {
+    /// Formats like this: `MOVI { dst: 32, val: 32 } (bytes 72 80 00 20)`}`
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match Instruction::try_from(self.0) {
+            Ok(instr) => write!(f, "{:?}", instr)?,
+            Err(_) => write!(f, "Unknown")?,
+        };
+        write!(f, " (bytes ")?;
```
MitchTurner, line 49: Nit: I think I'd prefer `bytes: `.

fuel-asm/src/panic_instruction.rs
```diff
@@ -35,6 +37,35 @@ impl PanicInstruction {
...
 }
 
+/// Helper struct to debug-format a `RawInstruction` in `PanicInstruction::fmt`.
+struct InstructionDbg(RawInstruction);
+impl fmt::Debug for InstructionDbg {
+    /// Formats like this: `MOVI { dst: 32, val: 32 } (bytes 72 80 00 20)`}`
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match Instruction::try_from(self.0) {
+            Ok(instr) => write!(f, "{:?}", instr)?,
+            Err(_) => write!(f, "Unknown")?,
```
MitchTurner, line 47: Nit: Maybe something like `Unknown Instruction` or `Unknown Operation' might be slightly more clear when the caller doesn't know what to expect in this field. I don't feel strongly about it though.

MitchTurner (commented):
LGTM. I just left a couple nits. 
I hadn't thought of just adding it to the `Debug`. This seems like a good way to get what we wanted without touching much code.

xgreenx (approved):
I like both variants=D

fuel-asm/src/panic_instruction.rs
Dentosal: It already says `instruction: Unknown`, so I think it's good enough

fuel-asm/src/panic_instruction.rs
Dentosal: Ok, changed.

xgreenx (approved):

PR #547, opened 2023-08-14
Bump dalek version because of RustSec Advisory

MitchTurner:


fuel-vm/Cargo.toml
```diff
@@ -36,7 +36,7 @@ tai64 = "4.0"
 thiserror = "1.0"
 
 [dev-dependencies]
-ed25519-dalek = "1.0" # TODO: upgrade to 2.0 when it's released, and remove rand below
+ed25519-dalek = "2.0.0"
 ed25519-dalek_old_rand = { package = "rand", version = "0.7.3" }
```
xgreenx, line 40: Do we still need to use this one?=)

fuel-crypto/Cargo.toml
```diff
@@ -15,7 +15,7 @@ borrown = "0.1"
 coins-bip32 = { version = "0.8", default-features = false, optional = true }
 coins-bip39 = { version = "0.8", default-features = false, features = ["english"], optional = true }
 ecdsa = { version = "0.16", default-features = false }
-ed25519-dalek = { version = "1.0", default-features = false, features = ["u64_backend"] }
+ed25519-dalek = { version = "2.0.0", default-features = false }
```
xgreenx, line 18: Is `u64_backend` a default behavior right now?

MitchTurner: I couldn't see what happened to it, so I figured I'd rely on the the compiler to tell me.

fuel-vm/Cargo.toml
xgreenx: image

It has `rand_core` feature, maybe we can use it isntead

MitchTurner: Yep. That seems to work.

xgreenx (approved):

PR #548, opened 2023-08-14
Bump dalek version to deal with security vulnerability

### LINKED ISSUE #549: RUSTSEC-2022-0093: Double Public Key Signing Function Oracle Attack on `ed25519-dalek`
> Double Public Key Signing Function Oracle Attack on `ed25519-dalek`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `ed25519-dalek`                      |
| Version             | `1.0.1`                   |
| URL                 | [https://github.com/MystenLabs/ed25519-unsafe-libs](https://github.com/MystenLabs/ed25519-unsafe-libs) |
| Date                | 2022-06-11                         |
| Patched versions    | `>=2`                  |

Versions of `ed25519-dalek` prior to v2.0 model private and public keys as
separate types which can be assembled into a `Keypair`, and also provide APIs
for serializing and deserializing 64-byte private/public keypairs.

Such APIs and serializations are inherently unsafe as the public key is one of
the inputs used in the deterministic computation of the `S` part of the signature,
but not in the `R` value. An adversary could somehow use the signing function as
an oracle that allows arbitrary public keys as input can obtain two signatures
for the same message sharing the same `R` and only differ on the `S` part.

Unfortunately, when this happens, one can easily extract the private key.

Revised public APIs in v2.0 of `ed25519-dalek` do NOT allow a decoupled
private/public keypair as signing input, except as part of specially labeled
&quot;hazmat&quot; APIs which are clearly labeled as being dangerous if misused.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2022-0093.html) for additional details.

MitchTurner:
Closes #549

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Description of the upcoming release here.
 
+## [Hotfix 0.35.4]
+
+### Fixed
+
+- [#548](https://github.com/FuelLabs/fuel-vm/pull/548): Bumped dalek to `2.0.0` to fix security vulnerability.
+
 ## [Version 0.35.3]
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Description of the upcoming release here.
 
-## [Hotfix 0.35.4]
+## [Version 0.35.4]
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ edition = "2021"
 homepage = "https://fuel.network/"
 license = "BUSL-1.1"
 repository = "https://github.com/FuelLabs/fuel-vm"
-version = "0.35.3"
+version = "0.35.4"
 
 [workspace.dependencies]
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Description of the upcoming release here.
 
+## [Hotfix 0.35.4]
```
MitchTurner, line 11: I just realized I made this up without thinking about it. Should this be `Version` still?

xgreenx (approved):

xgreenx (commented):
You also need to bump the version to `0.35.4`=)

Dentosal:
This also needs fixing in fuel core, issue https://github.com/FuelLabs/fuel-core/issues/1298. Should be a fairly simple fix as well.

Dentosal (dismissed):

Cargo.toml
```diff
@@ -18,7 +18,7 @@ edition = "2021"
 homepage = "https://fuel.network/"
 license = "BUSL-1.1"
 repository = "https://github.com/FuelLabs/fuel-vm"
-version = "0.35.3"
+version = "0.35.4"
 
 [workspace.dependencies]
 fuel-asm = { version = "0.35.3", path = "fuel-asm", default-features = false }
```
xgreenx, line 24: You also need to update it in `workspace.dependencies` section

PR #554, opened 2023-08-26
Remove `debug` feature from `fuel-vm` and use debugger by default.

xgreenx:


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Breaking
 
+- [#554](https://github.com/FuelLabs/fuel-vm/pull/554): Removed `debug` feature from teh `fuel-vm`. The debugger is always available and becomes active after calling any `set_*` method.
```
bvrooman, line 13:
```suggestion
- [#554](https://github.com/FuelLabs/fuel-vm/pull/554): Removed `debug` feature from the `fuel-vm`. The debugger is always available and becomes active after calling any `set_*` method.
```

fuel-vm/src/state/debugger.rs
```diff
@@ -14,31 +14,37 @@ use std::collections::{
...
+    is_active: bool,
     /// Single-stepping mode triggers a breakpoint after each instruction
     single_stepping: bool,
     breakpoints: HashMap<ContractId, HashSet<Word>>,
     last_state: Option<ProgramState>,
 }
 
 impl Debugger {
+    /// Returns `true` if the `Debugger` is active and used.
+    pub const fn is_active(&self) -> bool {
```
bvrooman, line 30: Would it be useful to expose an API to activate the debugger explicitly, without calling a `set_*` function?

bvrooman: Is there any way to turn off debugging? In case you need to switch between debugging and performance modes

bvrooman (commented):
What is the motivation behind this change? On one hand, I can see that it could simplify our builds if we don't need to distinguish between a debug and release build. On the other hand, it seems a bit unorthodox to mix debug features in as an always-compiled runtime check.

xgreenx:
> What is the motivation behind this change? On one hand, I can see that it could simplify our builds if we don't need to distinguish between a debug and release build. On the other hand, it seems a bit unorthodox to mix debug features in as an always-compiled runtime check.

If we [don't want to produce two images](https://github.com/FuelLabs/fuel-core/pull/1322/files/cde339ef7b5c8f563a9188f337ff6adbfa02f525..dfd1be9265be7507ca3b2d805c14b3cfc569b8ab)(`debug` and `release`), then we need to either enable the debugger as part of the actual VM, or make the `Interpreter` generic over the `Debugger`.

While the second solution doesn't impact performance, it causes a lot of modifications in the code, propagating generic everywhere. 

I decided to select the approach with the checking of one variable instead of a generic one to make the change fast. It shouldn't affect performance too much. If you prefer the way with generic, I can re-work it=) But I tried to avoid potential conflicts with https://github.com/FuelLabs/fuel-vm/pull/533

bvrooman (approved):

bvrooman (approved):

Issue #536, opened 2023-07-28
Make the fuel-vm no-std or WASM/RiscV compatible


SilentCicero:
## Abstract
We need to make the entire `fuel-vm` implementation **RiscV compatible** for fraud proving.

This would also allow for better browser, debugging and educational tools for the FuelVM.

### Currently Compatible
-  `fuel-asm`
- `fuel-types`
- `fuel-merkle`
- `fuel-storage`
- `fuel-crypto`

### Not Compatible
- `fuel-tx` (uses `fuel-crypto`)
- `fuel-vm` (uses `fuel-tx`)

Related:
https://github.com/FuelLabs/fuel-vm/issues/443
https://github.com/FuelLabs/fuel-vm/pull/533

Dentosal:
Related: `getrandom` doesn't support freestanding riscv64, so we'll need [a fallback rng](https://docs.rs/getrandom/latest/getrandom/#unsupported-targets) or some other way to work around that.

DefiCake:
Just reporting the state of compatibility with risc0 / riscV, fuel-crypto is compiling correctly now.

Currently, the build process complains about `fuel-tx`. 

```
error[E0277]: the trait bound `Transaction: fuel_tx::FormatValidityChecks` is not satisfied
   --> /Users/mad/fuel/vm/fuel-vm/src/checked_transaction.rs:577:22
    |
577 | impl IntoChecked for Transaction {
    |                      ^^^^^^^^^^^ the trait `fuel_tx::FormatValidityChecks` is not implemented for `Transaction`
    |
    = help: the following other types implement trait `fuel_tx::FormatValidityChecks`:
              fuel_tx::Create
              fuel_tx::Mint
              fuel_tx::Script
... 
```

Not very experienced in rs, but it looks to me as the `FormatValidityChecks` is only available under `std`:
```rs
#[cfg(feature = "std")]
impl FormatValidityChecks for Transaction {
...
```

Dentosal:
This is almost achieved in https://github.com/FuelLabs/fuel-vm/issues/536, which makes the VM itself runnable under `no_std` conditions but still lacks some of the RiscV-specific stuff.

xgreenx:
After https://danielmangum.com/posts/risc-v-bytes-rust-cross-compilation/, we are able to compile `fuel-vm` and `fuel-tx` into WASM and RiscV.

``` shell
[...]

PR #578, opened 2023-09-01
no_std fuel-crypto

Dentosal:
Work towards #536.

This PR removes `secp256k1` dependency from `fuel-crypto` when not depending on stdlib, and uses currently **un-audited** `k256` crate, which is a pure-rust implementation. [There's an audit underway for the it, though](https://github.com/RustCrypto/elliptic-curves/pull/907).

The API doesn't change much, but generating new private secp256k1 keys now requires the rng to implement [the `rand::CryptoRng` trait](https://docs.rs/rand/0.8.5/rand/trait.CryptoRng.html), meaning they cannot be generated using `rng.gen()` but require `SecretKey::random(&mut rng)` instead.


.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           - command: check
             args: --all-targets --all-features
           - command: check
-            args: --target thumbv6m-none-eabi -p fuel-asm -p fuel-storage -p fuel-merkle --no-default-features
+            args: --target thumbv6m-none-eabi -p fuel-crypto --no-default-features
```
Voxelot, line 62: why only fuel-crypto now instead of all the other crates?

fuel-crypto/src/secp256/backend/k1/k256.rs
```diff
@@ -0,0 +1,147 @@
...
+    let sk: ecdsa::SigningKey<k256::Secp256k1> = sk.into();
+    let vk = sk.verifying_key();
+    vk.into()
+}
+
+/// Sign a given message and compress the `v` to the signature
+///
+/// The compression scheme is described in
+/// <https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/cryptographic-primitives.md>
+pub fn sign<SK: Into<k256::SecretKey>>(secret: SK, message: &Message) -> [u8; 64] {
```
Voxelot, line 49: maybe we should use an actual trait interface in front of these backends so that it's less likely users of this crate get unexpected errors when switching between std and nostd?

fuel-crypto/benches/signature.rs
```diff
@@ -251,10 +247,14 @@ fn signatures(c: &mut Criterion) {
 
     group_recover.bench_with_input(
         "k256",
-        &(k2_recoverable, k2_digest),
-        |b, (recoverable, digest)| {
+        &(k2_signature, k2_recovery_id, k2_digest),
+        |b, (signature, recovery_id, digest)| {
             b.iter(|| {
-                recoverable.recover_verifying_key_from_digest(black_box(digest.clone()))
+                k256::ecdsa::VerifyingKey::recover_from_digest(
```
Voxelot, line 253: Shouldn't we be benching off of the library we use during regular block execution in std?

Voxelot: Also why remove the black box?

fuel-crypto/src/secp256/backend/k1/k256.rs
Dentosal: I considered it, but I really don't see what extra value using a trait here would bring. Any errors would already be caught by the type system, or at least the tests. I wont mind wrapping them in a trait here if you feel like it's better like that.

fuel-crypto/benches/signature.rs
Dentosal: We bench from both `k256` and `secp256k1` libraries, since we want both of the benchmarks.

fuel-crypto/src/secp256/backend/k1/k256.rs
Dentosal: Also, the types used by `public_key` and `sign` functions are not the same for both modules, as they use `Into<...>` with a type from the corresponding backing crate, although that could be changed as well.

fuel-vm/src/checked_transaction.rs
```diff
@@ -712,9 +712,9 @@ mod tests {
         let tx = TransactionBuilder::script(vec![], vec![])
             .gas_price(gas_price)
             .gas_limit(gas_limit)
-            .add_unsigned_message_input(rng.gen(), rng.gen(), rng.gen(), input_amount, vec![0xff; 10])
```
bvrooman, line 715: Instead of replacing all the instances of `rng.gen()`, is it possible to do something like: 
```impl Distribution<SecretKey> for Standard { ... }```
instead? I believe this should allow you to keep the existing syntax in these places.

**Review by bvrooman — COMMENTED** [review]
I am not very familiar with the cryptography concepts in this PR, but nothing sticks out as an obvious issue. 

Just one small suggestion related to implementing `Distribution` for `SecretKey`.

File: fuel-vm/src/checked_transaction.rs (hunk 4 above)
Comment (MitchTurner) [reply fuel-vm/src/checked_transaction.rs:715]: I don't hate this syntax. It's explicit.

**Review by MitchTurner — COMMENTED** [review]
LGTM. IDK if I feel comfortable approving because a bit of the crypto stuff is too unfamiliar for me to scrutinize closely.

File: fuel-vm/src/checked_transaction.rs (hunk 4 above)
Comment (Dentosal) [reply fuel-vm/src/checked_transaction.rs:715]: No. We had that before, but sadly the `elliptic-curve` crate (used by `k256`) [requires](https://docs.rs/elliptic-curve/0.13.4/elliptic_curve/struct.SecretKey.html#method.random) use to accept only rngs with [`CryptoRngCore` trait](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRngCore.html). At least I couldn't figure out how to make `impl Distribution` work with this.

File: .github/workflows/ci.yml (hunk 1 above)
Comment (Dentosal) [reply .github/workflows/ci.yml]: Good catch. I didn't intend to remove those. Restored in 0a180b40

File: fuel-crypto/benches/signature.rs (hunk 3 above)
Comment (Dentosal) [reply fuel-crypto/benches/signature.rs:253]: The inputs are automatically passed through a `black_box` [docs](https://bheisler.github.io/criterion.rs/book/user_guide/benchmarking_with_inputs.html#benchmarking-with-one-input). `black_box`ing the result of `.clone()` adds overhead that's not part of what we're benchmarking here.

File: fuel-crypto/src/secp256/backend/k1/k256.rs (hunk 2 above)
Comment (Voxelot) [reply fuel-crypto/src/secp256/backend/k1/k256.rs:49]: It would probably be better if we had our own types on the public interfaces, and then implemented conversion to various backends as needed based on what's compiled in. Otherwise if one of these crypto libs changes their types or interfaces it could cause a lot of breaking changes in other places.

Comment (Dentosal) [reply fuel-crypto/src/secp256/backend/k1/k256.rs:49]: Agreed, although I find it unlikely to actually cause breakage, but it's true that upgrading an underlying crypto library is a breaking change in the current implementation. I'll fix that.

Comment (Dentosal) [reply fuel-crypto/src/secp256/backend/k1/k256.rs:49]: Ok, I've fixed this now for `secp256k1`. For `secp256r1` we currently have only a single implementation, which still exposes some types from the underlying crypto lib under `#[cfg(feature = "test-helpers")]`, which I think is fine.

**Review by bvrooman — DISMISSED** [review]

File: fuel-crypto/src/secp256/backend/k1/k256.rs (hunk 2 above)
Comment (Voxelot) [reply fuel-crypto/src/secp256/backend/k1/k256.rs:49]: Which commit has this fix? I'm not seeing it

Comment (Dentosal) [reply fuel-crypto/src/secp256/backend/k1/k256.rs]: f52fe821ce585b43a43cd2e8c03734f083ef0985 and 053fa238f122214ad79670de54029116434bd8c7. Forgot to push those, sorry.

**Review by bvrooman — APPROVED** [review]

**Comment by bvrooman** [comment]
No additional questions from me. Again, I'm not familiar with all the cryptography concepts, so I will have to go by the tests and your judgement here.

File: fuel-crypto/src/tests/signature.rs [hunk 5]
```diff
@@ -5,11 +5,14 @@ use crate::{
...
     Signature,
 };
+
+#[cfg(feature = "std")]
 use rand::{
     rngs::StdRng,
     SeedableRng,
 };
 
+#[cfg(feature = "std")]
```
Voxelot, line 15: I get that the secret key generation doesn't work in no_std, but do we have tests in other places to ensure that the no_std backends work the same as std?

fuel-vm/Cargo.toml
```diff
@@ -19,7 +19,7 @@ derivative = "2.2"
 dyn-clone = { version = "1.0", optional = true }
 ethnum = "1.3"
 fuel-asm = { workspace = true }
-fuel-crypto = { workspace = true }
+fuel-crypto = { workspace = true, default-features = false }
```
Voxelot, line 22: Would this make the VM use the no-std backend by default in all cases? Maybe we should also enable the std feature on fuel-crypto if it's enabled on fuel-vm?

Voxelot: Using the no-std version by default could also impact our gas pricing if the performance between the native secp256k1 lib is very different from the pure-rust impl.

fuel-crypto/src/tests/signature.rs
Dentosal: The test for backend equivalency is in `fuel-crypto/src/secp256/backend.rs`. Only the secp256k1 has different implementations currently.

I agree that it makes sense to add some tests to the backends themselves.

fuel-vm/Cargo.toml
Dentosal: I'll check that the features are correct, as that seems like a potential issue. It's somewhat hard to verify that the correct crypto backend is getting used, as they have the same external API.

> Using the no-std version by default could also impact our gas pricing if the performance between the native secp256k1 lib is very different from the pure-rust impl.

It certainly does, the `no_std` version is about 5x slower.  The gas pricing shouldn't care about `no_std` performance, as the standard option would be to run the nodes with native secp256k1.

Dentosal: Added feature propagation in dc6962a9]

fuel-crypto/src/tests/signature.rs
Dentosal: I added standalone test cases for k256 module that also run in `no_std` mode, in e3943dee

Voxelot (approved):

PR #581, opened 2023-09-06
fix: Organize `ConsensusParameters` gas costs in alphabetical order

bvrooman:
Currently some gas costs are not presented in alphabetical order, creating confusion around the way the chainspec gas costs should be ordered. By putting them in alphabetical order, there is a deterministic way to define the gas costs in a chainspec. 

Putting the chainspec gas costs in alphabetical order alone is insufficient, as it creates a discrepancy in the serialized versions of the chainspec. This also causes tests to fail.


MitchTurner (approved):
LGTM

Issue #230, opened 2022-10-11
Error propagation


vlopes11:
In order to deliver more & fast, `std::io::Error` was abused due to its broad compatibility with different crates + downstream + upstream usage, ranging from serde serialization to rocksdb storage operations

We need to unify the error scheme of the stack, keeping in mind that we have two classes of errors:

- recoverable: anything that derives from a script that cannot cause an execution halt and should end charging the consumed gas from the user
- unrecoverable: anything that goes beyond the static transactions such as OS errors (out of memory, etc) or storage errors. In these cases, gas can't be charged from the user, for valid transactions

Dentosal:
This will be likely fixed in https://github.com/FuelLabs/fuel-vm/pull/582, as we cannot use `std::io::Error` anymore.

PR #582, opened 2023-09-07
no_std support for `fuel-tx` and `fuel-vm`

Dentosal:
Makes fuel-vm crate compile without stdlib, but still requireing alloc. This is a breaking change, and the corresponding updates to fuel-core are done in https://github.com/FuelLabs/fuel-core/pull/1367. Most of the  breakage comes from removing `std::io::Error`, especially from storage traits which now use concrete error types instead and only allow `std::io::Error` on some variants available with `std` feature enabled.

Currently passes with the following command for RiscV:

```bash
cargo build --target riscv32imac-unknown-none-elf -p fuel-vm --no-default-features --features alloc
```

and check passes for freestanding x86-64, but building doesn't, since it requires soft-float configuration:

```bash
cargo check --target x86_64-unknown-none -p fuel-vm --no-default-features --features alloc
```

Closes https://github.com/FuelLabs/fuel-vm/issues/230 https://github.com/FuelLabs/fuel-vm/issues/536


fuel-merkle/src/storage.rs
```diff
@@ -24,7 +23,7 @@ pub trait StorageMutateInfallible<Type: Mappable> {
 
 impl<S, Type> StorageInspectInfallible<Type> for S
 where
-    S: StorageInspect<Type, Error = Infallible>,
+    S: StorageInspect<Type>,
     Type: Mappable,
```
xgreenx, line 27: The idea of this blanket implementation is that we can unwrap the result of the call because we are sure it is impossible to have an error. Is that true now?

fuel-storage/src/error.rs
```diff
@@ -0,0 +1,55 @@
+/// Wraps around possible errors that can occur during storage operations.
+#[derive(Debug)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+pub enum StorageError {
```
xgreenx, line 4: Why do we want to move this error from `fuel-core` to `fuel-vm`?

fuel-tx/src/lib.rs
```diff
@@ -98,8 +98,9 @@ pub use transaction::{
     Witness,
 };
 
-#[cfg(feature = "std")]
+#[cfg(feature = "alloc")]
 pub use transaction::Signable;
+#[cfg(feature = "alloc")]
 pub use transaction::UniqueIdentifier;
```
xgreenx, line 104: Hmm, I think it should work without `alloc` feature

fuel-tx/src/transaction/validity/error.rs
```diff
@@ -6,14 +6,10 @@ use fuel_types::{
...
 };
-#[cfg(feature = "std")]
-use std::{
-    error,
-    io,
-};
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
```
xgreenx, line 12: Do we really need to support this kind of conversion? Because on the `fuel-core` side we also should be able to compile executor in `no_std` mode

fuel-vm/src/error.rs
```diff
@@ -5,45 +5,50 @@ use fuel_asm::{
     PanicReason,
     RawInstruction,
 };
+use fuel_storage::StorageError;
 use fuel_tx::CheckError;
+
+#[cfg(feature = "std")]
 use thiserror::Error;
```
xgreenx, line 12: Maybe we want to use https://docs.rs/thiserror-no-std/latest/thiserror_no_std/

fuel-vm/src/interpreter/constructors.rs
```diff
@@ -57,6 +59,8 @@ where
     where
         P: ProfileReceiver + Send + Sync + 'static,
     {
+        #[cfg(not(feature = "std"))]
+        use alloc::boxed::Box;
         self.profiler.set_receiver(Box::new(receiver));
```
xgreenx, line 64:
```suggestion
        self.profiler.set_receiver(alloc::boxed::Box::new(receiver));
```

fuel-vm/src/interpreter/debug.rs
```diff
@@ -119,6 +125,7 @@ fn breakpoint_script() {
 
 #[test]
 fn single_stepping() {
+    use alloc::vec;
```
xgreenx, line 128: It seems you already should have this import from above.

fuel-vm/src/lib.rs
```diff
@@ -2,8 +2,15 @@
 
 #![warn(missing_docs)]
 #![deny(unsafe_code)]
-#![deny(unused_crate_dependencies)]
+// #![deny(unused_crate_dependencies)]
```
xgreenx, line 5: Can we uncomment it?=)

fuel-vm/src/lib.rs
```diff
@@ -2,8 +2,15 @@
...
 #![deny(clippy::string_slice)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[doc(hidden)] // Needed by some of the exported macros
+pub extern crate alloc;
+
+#[cfg(feature = "std")]
+extern crate libm as _; // Not needed with stdlib
```
xgreenx, line 13: Based on the comment we need to use `cfg(not(feature = "std"))`

fuel-vm/src/storage/memory.rs
```diff
@@ -204,30 +203,28 @@ impl StorageRead<ContractsRawCode> for MemoryStorage {
...
-            .memory
-            .contracts
-            .get(key)
-            .and_then(|c| c.as_ref().read(buf).ok()))
+    ) -> Result<Option<usize>, StorageError> {
+        Ok(self.memory.contracts.get(key).map(|c| {
+            let len = buf.len().min(c.as_ref().len());
+            buf.copy_from_slice(&c.as_ref()[..len]);
+            buf.len()
+        }))
```
xgreenx, line 211: Hmm, it seems not super related to this change=)

Do we really want to copy into the buffer silently if the sizes are different?

fuel-vm/src/lib.rs
```diff
@@ -26,6 +33,7 @@ pub mod util;
 pub mod profiler;
 
 #[cfg(test)]
+#[cfg(feature = "alloc")]
```
xgreenx, line 36: We can enable `alloc` in `dev-dependencies` and use it by default. Running tests without this feature doesn't make a lot of sense.

xgreenx (commented):
It is excellent that we can compile and run `fuel-vm` in `no-std` mode. 

The main question is related to `StorageError`. It simplifies the code in many places, and we can do that because we removed dependency on `io::Error`. I liked the combination of associated types and infallible errors that allowed us to "unwrap" the result in some places before.

I'm not too fond of this decision because we move the definition of errors from `fuel-core`(the place that really knows how to work with the storage) to `fuel-storage` level. It forces us either define all possible business errors on this level(when it partially breaks the idea of having a separate `fuel-storage` crate) or work with strings-like errors. Plus, we still don't support `StoageError::Database` properly for `no_std` mode.

Maybe, if we moved the error from the associated type to `StoageError<DatabaseError>` itself it would be better in terms of implementation and support. But I'm not sure what is the best approach here.

fuel-storage/src/error.rs
Dentosal: The main issue here is that if the error type lives in `fuel-core`, the `fuel-vm` cannot understand this type without depending on `fuel-core-storage` which would be circular. This error is used in `InterpreterError`. We could instead use `anyhow::Error` or a custom error trait, but those have their own issues.

fuel-tx/src/lib.rs
Dentosal: Yep, fixed in 6d9229ef

fuel-vm/src/lib.rs
Dentosal: This line exists to satisfy `#![deny(unused_crate_dependencies)]`. Since we only depend on `libm` on `no_std` targets, we need to ignore the dependency like this when `stdlib` is available. It's not possible to specify a dependency that's disabled when a feature is enabled.

fuel-vm/src/lib.rs
Dentosal: The lint is broken when compiling without `std` feature, and we'd have to work around it quite a bit. Currently it's since some `dev-dependencies` are only used when `feature = "std"`.

The lint itself is quite problematic, see https://github.com/rust-lang/rust/issues/57274 for discussion on that.

Our options here are:
* Removing the lint and moving back to  `cargo-udeps`
* Removing the lint and not automatically checking for unused deps
* Feature-gating the lint: `#[cfg_attr(feature = "std", deny(unused_crate_dependencies))]`
* Or doing this:

```rust
#[cfg(all(test, not(feature = "std")))] use futures as _;
#[cfg(all(test, not(feature = "std")))] use tokio as _;
#[cfg(all(test, not(feature = "std")))] use tokio_rayon as _;
```

fuel-vm/src/lib.rs
Dentosal: Done in 90d00506

fuel-vm/src/interpreter/debug.rs
Dentosal: Removed in be98b794. Weird that `clippy` didn't catch this.

fuel-vm/src/error.rs
Dentosal: That crate looks incredibly suspicious. It contains no explanation on differences with `thiserror` crate itself, it's owned by a different user than `thiserror`, and all links (docs, repo, ...) point to those of the original `thiserror`.

We don't really need `thiserror` here at all, but I decided to keep it to avoid unnecessary breaking changes.

fuel-tx/src/transaction/validity/error.rs
Dentosal: No, just avoiding breaking changes whenever possible. Do you think it whould be removed?

fuel-storage/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+            (Self::NotFound(a, b), Self::NotFound(c, d)) => a == c && b == d,
+            #[cfg(feature = "std")]
+            (Self::Database(a), Self::Database(b)) => a.to_string() == b.to_string(),
+            _ => core::mem::discriminant(self) == core::mem::discriminant(other),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl From<StorageError> for std::io::Error {
```
bvrooman, line 51: Are there still instances where we use `std::io::Error`? Or can we get rid of this conversion now?

fuel-vm/src/arith.rs
```diff
@@ -30,55 +28,3 @@ pub fn checked_add_usize(a: usize, b: usize) -> Result<usize, PanicReason> {
...
-
-/// Add two checked words. Might wrap if overflow
-///
-/// The function is inlined so it will be optimized to `a + b` if `optimized` feature is
-/// enabled so the operation is unchecked.
-///
-/// This should be used in contexts that are checked and guaranteed by the protocol to
-/// never overflow, but then they might due to some bug in the code.
-#[inline(always)]
-pub fn add_word(a: Word, b: Word) -> Result<Word, RuntimeError> {
```
bvrooman, line 42: Is this code change necessary to support the `no_std` work, or is it simply tangential?

Dentosal:
Migrating to a single concrete `StorageError` type seems like an disliked approach, and while I still like it more myself, I can see that it has downsides.

I've reversed the direction on this, and instead parametrized `InterpreterError` with a generic parameter carrying the storage error type, so that we don't have to use a custom error trait. Now the resulting error type is directly available to the users of `fuel-vm`, and isn't hidden behind `std::io::Error`. This means some additional work on `fuel-core` side, but it should be too hard, especially since we only used type-erased errors before. I'm planning to do those changes tomorrow.

Let me know your opinions on the new approach.

fuel-merkle/src/storage.rs
Dentosal: I've changed the approach, see https://github.com/FuelLabs/fuel-vm/pull/582#issuecomment-1728570111

fuel-storage/src/error.rs
Dentosal: I've changed the approach, see https://github.com/FuelLabs/fuel-vm/pull/582#issuecomment-1728570111

fuel-crypto/src/secp256/backend/r1/p256.rs
```diff
@@ -57,6 +57,7 @@ pub fn sign_prehashed(
 
 /// Convert the public key point to its uncompressed non-prefixed representation,
 /// i.e. 32 bytes of x coordinate and 32 bytes of y coordinate.
+#[cfg(feature = "test-helpers")]
```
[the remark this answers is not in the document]
Dentosal: I don't like it, since it's likely that these will be un-feature-gated at some point

fuel-merkle/src/binary/merkle_tree.rs
```diff
@@ -23,10 +23,11 @@ use crate::{
 
 use alloc::vec::Vec;
 use core::marker::PhantomData;
+use fuel_storage::StorageError;
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 #[cfg_attr(feature = "std", derive(thiserror::Error))]
-pub enum MerkleTreeError<StorageError> {
```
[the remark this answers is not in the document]
Dentosal: Agreed. I was a bit overeager to un-genericize the types since I needed to do changes in `fuel-vm` removing some abstraction. It was unnecessary on `fuel-merkle`. I've changed the approach, see https://github.com/FuelLabs/fuel-vm/pull/582#issuecomment-1728570111

fuel-merkle/src/common/storage_map.rs
```diff
@@ -44,15 +45,16 @@ where
...
+        &self,
+        key: &Type::Key,
+    ) -> Result<Option<Cow<Type::OwnedValue>>, StorageError> {
         let result = self.map.get(key);
         let value = result.map(Cow::Borrowed);
         Ok(value)
     }
 
-    fn contains_key(&self, key: &Type::Key) -> Result<bool, Self::Error> {
+    fn contains_key(&self, key: &Type::Key) -> Result<bool, StorageError> {
```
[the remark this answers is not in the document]
Dentosal: Agreed

fuel-vm/src/arith.rs
Dentosal: Not striclty necessary, but the error/bug type rework is needed. It would be more work to keep it like this, and it needs fixing anyways.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -567,15 +571,20 @@ where
             .checked_sub(a)
             .ok_or(PanicReason::NotEnoughBalance)?;
 
-        self.storage
-            .merkle_contract_asset_id_balance_insert(contract_id, &asset_id, balance)
-            .map_err(RuntimeError::from_io)?;
+        match self.storage.merkle_contract_asset_id_balance_insert(
```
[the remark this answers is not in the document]
Dentosal: Yep, 94d1050b

fuel-vm/src/arith.rs
bvrooman: Yeah, I definitely agree that it is helpful work towards the main goal - I simply ask because sometimes we can separate discrete chunks of work into separate PRs. But that comes down to personal preference

fuel-vm/src/error.rs
bvrooman: Removing `thiserror` and implementing `Display` would be a nice simplification. 

I am okay with the current if we would like to remove `thiserror` in a separate PR.

bvrooman (approved):
LGTM. Personally, I like the approach to parameterize the `InterpreterError` because it allows us to maintain separation between discrete levels of logic.

re: `thiserror` - If we want to remove `thiserror` in favour of writing custom `Display` traits, I think that we can do that in a separate PR.

xgreenx (approved):
It would be nice to have it in this PR because it can remove some `std` features from some crates, but I'm also okay with doing that in a separate PR=)

.github/workflows/ci.yml
```diff
@@ -95,7 +97,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_VERSION }}
-          targets: "thumbv6m-none-eabi,wasm32-unknown-unknown"
+          targets: "thumbv6m-none-eabi,riscv32imac-unknown-none-elf,wasm32-unknown-unknown"
```
bvrooman, line 100: Just curious - did you chose the IMAC extensions for any particular reason, or is that just a safe default?

Dentosal: That's closes to the internal target used by risc-zero

Dentosal:
I'm leaving the thiserror -> display impl change for a follow-up PR, as I feel like it's nicer to review separately.

PR #585, opened 2023-09-18
Update AssetId to follow new specs

SwayStar123:
Current AssetId type still assumes that the ContractId and AssetId are the same. This PR adds new `new` and `default_for_contract` methods that match the `new` and `default` methods from the Sway std lib type.


fuel-types/src/array_types.rs
```diff
@@ -367,6 +390,42 @@ impl ContractId {
...
+    /// * `contract_id` - The ContractId of the contract that created the asset.
+    pub fn default_for_contract<T: Into<ContractId>>(contract_id: T) -> Self {
+        Self::new(contract_id, SubId::zeroed())
+    }
+
+    /// The base_asset_id represents the base asset of a chain.
+    ///
+    /// ### Additional Information
+    ///
+    /// On the Fuel network, the base asset is Ether. It is hardcoded as the 0x00..00
```
xgreenx, line 424: It is not true after we made the base asset id configurable. We plan to hide `AssetId::BASE` under `test-helpers` feature in the future, so it is better to remove this method

fuel-types/src/array_types.rs
```diff
@@ -80,16 +88,30 @@ macro_rules! key_with_big_array {
...
             /// Bytes constructor.
             pub const fn new(bytes: [u8; $s]) -> Self {
                 Self(bytes)
             }
+        }
+    };
+    ($i:ident, $s:expr,false) => {
+        impl $i {
+            /// Bytes constructor
+            pub const fn new_from_bytes(bytes: [u8; $s]) -> Self {
```
xgreenx, line 103: It is better to have only one the same interface/constructor for all types. We can use `new_from_bytes` for each type because it has the same meaning for all of them. But is is a huge breaking change. Instead, we can continue to use `new` but describe that it creates the type from the bytes without any business logic.

fuel-types/Cargo.toml
```diff
@@ -15,6 +15,7 @@ fuel-derive = { workspace = true }
 hex = { version = "0.4", default-features = false, optional = true }
 rand = { version = "0.8", default-features = false, optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive", "alloc"], optional = true }
+sha2 = { version = "0.10", default-features = false }
```
xgreenx, line 18: One of the reasons why we didn't do that during the initial implementation is that we didn't want to increase the number of dependencies. Plus, we want to use the default hasher from `fuel-crypto`.

xgreenx (commented):
We don't have `From` implementation for `ContractId` and `AssetId`

fuel-types/Cargo.toml
SwayStar123: I tried using fuel-crypto, but that creates a circular dependency, not sure what to do

xgreenx: Even if you can, it still brings a lot of dependencies=) I meant it is why we decided to implement `asset_id` method in `fuel-tx`, because it is the business logic defined on `fuel-tx` level rather than on `fuel-types`.

SwayStar123: Hmm i think ill just close this pr and make a simpler one which just adds the default asset method in the ContractIdExt trait then

PR #586, opened 2023-09-19
Add default_asset method

SwayStar123:
Adds a convenience method to create the "default" asset. Mirroring the `default` function in Sway


fuel-tx/src/lib.rs
```diff
@@ -113,6 +113,8 @@ pub use contract::Contract;
 pub trait ContractIdExt {
     /// Creates an `AssetId` from the `ContractId` and `sub_id`.
     fn asset_id(&self, sub_id: &Bytes32) -> AssetId;
+    /// Creates an `AssetId` from the `ContractId` and the default 0x00..000 `sub_id`.
```
xgreenx, line 116:
```suggestion
    fn asset_id(&self, sub_id: &Bytes32) -> AssetId;

    /// Creates an `AssetId` from the `ContractId` and the default 0x00..000 `sub_id`.
```

xgreenx (approved):

PR #587, opened 2023-09-22
Replace thiserror with derive_more::Display

Dentosal:
Follow-up to https://github.com/FuelLabs/fuel-vm/pull/582

Replaces thiserror dependency of no_std crates with a `core::fmt::Display` implementations generated by `derive_more` crate. This is a breaking change, since we lose `std::io::Error` trait impls when `std` feature is enabled.


xgreenx (approved):
Thank you! <3

PR #588, opened 2023-09-22
Follow-up for canonical se/de after an additional round of review

xgreenx:
- Removed all enum attributes. The manual implementation of the traits can do the same.
- Removed `clone` for each `Input` required to call `into_full`. It is possible because of a new `Empty<Type>` type that implements canonical se/de with minimal implementation.
- Removed `SIZE_NO_DYNAMIC` because it is not used in the codebase.
- Removed `SerializedSizeFixed` because we don't have any specific logic that uses it.
- Moved all methods from `SerializedSize` to `Serialize`. It looks unnatural that `#[derive(Serialize)]` implements 3 traits. Instead, it implements on trait now.
- Removed `SIZE_STATIC` because it is impossible to calculate in the case of enums(or if the `struct` has an enum field). Supporting something like this requires another way of doing that, plus it doesn't provide a lot of wins for us right now because `#[inline(always]` handle that well enough.
- Removed panics during size calculation. Instead we use `saturating_add`. If the size is incorrect we will fail in another place.
- Removed usage of `num_enum` and replaced with canonical se/de.
- Removed calculation of the discriminant on macro level. Replaced with native rust calculation that covers all cases.


CHANGELOG.md
```diff
@@ -6,13 +6,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+
+- [#586](https://github.com/FuelLabs/fuel-vm/pull/586): Added `default_asset` method to the `ContractIdExt` trait implementation, to mirror the `default` method on AssetId in the Sway std lib.
+
+### Changed
 #### Breaking
 
 - [#578](https://github.com/FuelLabs/fuel-vm/pull/578): Support `no_std` environments for `fuel-crypto`, falling back to a pure-Rust crypto implementation.
 - [#582](https://github.com/FuelLabs/fuel-vm/pull/582): Make `fuel-vm` and `fuel-tx` crates compatible with `no_std` + `alloc`. This includes reworking all error handling that used `std::io::Error`, replacing some `std::collection::{HashMap, HashSet}` with `hashbrown::{HashMap, HashSet}` and many changes to feature-gating of APIs.
-### Added
+- [#588](https://github.com/FuelLabs/fuel-vm/pull/588): Removed `SerialziedSize` and `SerialziedFixedSize` traits. Removed support for `SIZE_NO_DYNAMIC` and `SIZE_STATIC`. Removed enum attributes from derive macro for `Serialize` and `Deserialize` traits.
```
Voxelot, line 18:
```suggestion
- [#588](https://github.com/FuelLabs/fuel-vm/pull/588): Removed `SerializedSize` and `SerializedFixedSize` traits. Removed support for `SIZE_NO_DYNAMIC` and `SIZE_STATIC`. Removed enum attributes from derive macro for `Serialize` and `Deserialize` traits.
```

fuel-derive/src/deserialize.rs
```diff
@@ -92,36 +87,30 @@ fn deserialize_enum(s: &synstructure::Structure) -> TokenStream2 {
                 }
             });
 
+            if variant.ast().discriminant.is_some() {
```
Voxelot, line 90: is this native discriminant stable across rust versions?

fuel-tx/src/transaction.rs
```diff
@@ -451,6 +453,81 @@ impl From<Mint> for Transaction {
     }
 }
 
+impl Serialize for Transaction {
```
Voxelot, line 456: What's the benefit of implementing this kind of boilerplate manually? Does it just add unnecessary complexity to the macros by trying to automate this?

fuel-tx/src/transaction/types/input.rs
```diff
@@ -55,20 +56,65 @@ pub trait AsField<Type>: AsFieldFmt {
...
-    fn as_mut_field(&mut self) -> Option<&mut Type> {
-        None
+    #[inline(always)]
+    fn size_dynamic(&self) -> usize {
+        0
+    }
+
+    #[inline(always)]
+    fn encode_static<O: Output + ?Sized>(&self, buffer: &mut O) -> Result<(), Error> {
+        Type::default().encode_static(buffer)
```
Voxelot, line 90: If this is empty, then it probably shouldn't even write anything to the buffer at all right?

fuel-vm/src/tests/encoding.rs
```diff
@@ -24,15 +24,17 @@ use fuel_types::{
 
 pub fn assert_encoding_correct<T>(data: &[T])
 where
-    T: Serialize + Deserialize + fmt::Debug + Clone + PartialEq,
+    T: Serialize + Deserialize + Serialize + fmt::Debug + Clone + PartialEq,
```
Voxelot, line 27: Why do we have serialize here twice?

> Serialize + Deserialize + Serialize

CHANGELOG.md
```diff
@@ -6,13 +6,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 
-- [#586](https://github.com/FuelLabs/fuel-vm/pull/586): Added `default_asset` method to the `ContractIdExt` trait implementation, to mirror the `default` method on AssetId in the Sway std lib.
+#### Removed
+- [#588](https://github.com/FuelLabs/fuel-vm/pull/588): .
```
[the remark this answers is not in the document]
xgreenx: I accidentally moved in to `Changed` section=D

fuel-derive/src/deserialize.rs
xgreenx: Yep, I used `x if x == { Self::Mint as u64 }`. But I decided to replace it with const. Now it looks like:

```rust
const V1: u64 = { Transaction::Mint as u64 };

match value {
    V1 => { ... }
    _ => { ... }
}
```

fuel-tx/src/transaction.rs
xgreenx: We can use the `peek` function, which removes the `Clone` requirement from the `trait Input: Clone`. Skipping of the discriminant is a very "magic" thing and I prefer to do that manually because we need it only in one place for `Transaction`.

fuel-tx/src/transaction/types/input.rs
xgreenx: In the case of the `Vec`, it writes 8 bytes - the length of the vector with the value `0`. The compiler will optimize it into writing zero into the array.

The `Empty` field doesn't occupy memory in runtime, but during serialization and deserialization it should participate and we need to include the static part.

fuel-vm/src/tests/encoding.rs
xgreenx: Replacement in all places=D

fuel-tx/src/transaction/types/input.rs
```diff
@@ -55,20 +56,65 @@ pub trait AsField<Type>: AsFieldFmt {
     fn as_mut_field(&mut self) -> Option<&mut Type>;
 }
 
-impl<Type> AsField<Type> for () {
+/// The empty field used by sub-types of the specification.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Empty<Type>(::core::marker::PhantomData<Type>);
```
Dentosal, line 62: This is a much nicer way to handle the empty types

Dentosal:
> Removed all enum attributes. The manual implementation of the traits can do the same.

I dislike the added copy-paste, as that makes it possible to accidentally apply a change to only a part of the branches.

> Removed clone for each Input required to call into_full. It is possible because of a new Empty<Type> type that implements canonical se/de with minimal implementation.

The empty type changes I really like, and this is a nice additional benefit.

> Removed SIZE_NO_DYNAMIC because it is not used in the codebase.
> Removed SerializedSizeFixed because we don't have any specific logic that uses it.
> Moved all methods from SerializedSize to Serialize. It looks unnatural that #[derive(Serialize)] implements 3 traits. Instead, it implements on trait now.

Sure.
 
> Removed SIZE_STATIC because it is impossible to calculate in the case of enums(or if the struct has an enum field). Supporting something like this requires another way of doing that, plus it doesn't provide a lot of wins for us right now because #[inline(always] handle that well enough.

I though there was an attempt to get `const`-time size calculations. If this is no longer a goal, then removing the const-time size access is a good thing.

> Removed panics during size calculation. Instead we use saturating_add. If the size is incorrect we will fail in another place.

I dislike this, as it has a possibility to silently fail, and could hide implementation bugs. Shouldn't matter much, but if we change the serialization itself at some point, this makes it a bit harder.

> Removed usage of num_enum and replaced with canonical se/de.

Sure.

> Removed calculation of the discriminant on macro level. Replaced with native rust calculation that covers all cases.

The new code seems good enough.

Dentosal (approved):
I disagree with some design decisions here, as mentioned above.
Still, I don't think it matters enough to keep discussing this, so lets just go ahead and merge.

PR #593, opened 2023-09-28
Rework `Mint` transaction to work with the contract

xgreenx:


fuel-vm/src/interpreter/contract.rs
```diff
@@ -370,8 +372,8 @@ where
         .unwrap_or_default())
 }
 
-/// Increase the asset balance for a contract
-pub(crate) fn balance_increase<S>(
+/// Increase the asset balance for a contract.
+pub fn balance_increase<S>(
```
xgreenx, line 376: I made it public because we can easily reuse it on the `fuel-core` side to update the contract balance.

fuel-tx/src/transaction/types/output/contract.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// State root of contract after transaction execution.
+    pub state_root: Bytes32,
+}
+
+impl Contract {
+    /// The "Note" section from the specification:
+    /// <https://github.com/FuelLabs/fuel-specs/blob/master/src/tx-format/output.md#outputcontract>.
+    pub fn prepare_sign(&mut self) {
+        core::mem::take(&mut self.balance_root);
+        core::mem::take(&mut self.state_root);
```
Dentosal, line 25: I would find more readable to use just an assignment here. Both of these styles should work:

```rust
*self.balance_root = Bytes32::zeroed();
*self.state_root = Default::default();
```

fuel-tx/src/transaction/types/mint.rs
```diff
@@ -94,6 +73,9 @@ impl crate::UniqueIdentifier for Mint {
         }
 
         let mut clone = self.clone();
+        clone.input_contract.prepare_sign();
+        clone.output_contract.prepare_sign();
+
         crate::transaction::compute_transaction_id(chain_id, &mut clone)
```
Dentosal, line 79: This is starting to look quite ineffiience, but maybe there's no nice way to not clone here?

fuel-tx/src/transaction.rs
```diff
@@ -589,6 +597,30 @@ pub mod field {
...
+    pub trait MintAmount {
+        fn mint_amount(&self) -> &Word;
+        fn mint_amount_mut(&mut self) -> &mut Word;
+        fn mint_amount_offset(&self) -> usize;
+    }
+
+    pub trait MintAssetId {
+        fn mint_asset_id(&self) -> &AssetId;
+        fn mint_asset_id_mut(&mut self) -> &mut AssetId;
+        fn mint_asset_id_offset(&self) -> usize;
```
bvrooman, line 621: What are the offsets for? It looks like you're calculating the positions of data in a buffer.

fuel-tx/src/transaction/types/mint.rs
Voxelot: Agreed, although probably not worth optimizing until we have real benchmarks.

xgreenx: We do it for each `Transaction` variant inside the `id` function. We could optimize by directly zeroing bytes in the final buffer, but it could add complexity and bugs to our code, so cloning is okay for now. Plus, we cache the `TxId` in most cases, so it is a one-time operation.

fuel-tx/src/transaction.rs
xgreenx: Yes, it is the index of where the field starts in the canonical byte representation of the transaction.

Dentosal (approved):
Looks ok to me, although maybe we should still wait for https://github.com/FuelLabs/fuel-specs/pull/518 to be accepted?

bvrooman (dismissed):

fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -834,39 +834,53 @@ fn mint() {
...
+        rng.gen(),
+        rng.gen(),
+        rng.gen(),
+    )
+    .finalize()
+    .check(block_height, &test_params())
+    .expect_err("Expected erroneous transaction");
 
-    assert_eq!(err, CheckError::TransactionMintOutputIsNotCoin);
+    assert_eq!(err, CheckError::TransactionMintIncorrectOutputIndex);
```
bvrooman, line 849: nit: This has the (very small) chance of failing intermittently, since the `input_index` is random and could land on 0. I think it could be nice to set up the test in a way that shows why this is error is expected.

bvrooman (approved):

Dentosal (approved):

fuel-tx/src/transaction/validity.rs
```diff
@@ -227,11 +227,12 @@ impl Output {
...
-            },
+            Self::Contract(output::contract::Contract { input_index, .. }) => {
+                match inputs.get(*input_index as usize) {
+                    Some(Input::Contract { .. }) => Ok(()),
+                    _ => Err(CheckError::OutputContractInputIndex { index }),
+                }
+            }
 
             _ => Ok(()),
         }
```
bvrooman, line 238: Maybe something like:

```rust
        match self {
            Self::Contract(output::contract::Contract { input_index, .. }) => inputs
                .get(*input_index as usize)
                .ok_or(CheckError::OutputContractInputIndex { index }),
            _ => Ok(()),
        }?;
        Ok(())
```

to reduce the reading complexity a bit.

bvrooman (approved):

fuel-tx/src/tests/offset.rs
```diff
@@ -429,7 +434,6 @@ fn tx_offset_script() {
 
 #[test]
 fn tx_offset_mint() {
```
MitchTurner, line 436: Can you explain to me what this is testing? I'm not familiar with the concept of "offsets" in this context.

xgreenx: The test verifies that [offset getters](https://github.com/FuelLabs/fuel-vm/pull/593/files#diff-e0db6cbfd4a6b1d090f2fd45858593b8c51e6adecb94b4ffbddad14065855df9) for each field are correct.

Later, we can use those offsets to extract values directly from the canonical byte representation of the transaction instead of deserializing the whole object(it can be quite expensive).

For example, we already use it for `Script` transactions. During the execution of the script, its fields are serialized into the VM memory, and the programmer of the Sway contract has access to them.

Issue #181, opened 2022-08-02
Add message predicate verification tests


vlopes11:
lgtm, although we should probably have a test to verify MessagePredicate inputs are actually verified

_Originally posted by @Voxelot in https://github.com/FuelLabs/fuel-vm/pull/180#pullrequestreview-1059099267_

xgreenx:
I think that it is already solved in the latest codebase. But it would be nice it someone will double check that @Dentosal <3

PR #594, opened 2023-10-02
Predicate input validation tests

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/181

Also improves error propagation so that predicate error message better reflects the reason for invalidation. For instance, it now includes the disallowed contract opcode which was not accepted in predicate exection context.

### DIFF
--- a/fuel-vm/src/predicate.rs
+++ b/fuel-vm/src/predicate.rs
@@ -189,17 +189,38 @@ mod tests {
             ($predicate:expr) => {{
                 let predicate: Vec<u8> = $predicate.into_iter().collect();
                 let owner = Input::predicate_owner(&predicate, &ChainId::default());
-                Input::coin_predicate(
-                    rng.gen(),
-                    owner,
-                    rng.gen(),
-                    rng.gen(),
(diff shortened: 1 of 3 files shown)


fuel-vm/src/predicate.rs
```diff
@@ -59,108 +61,205 @@ fn from_tx_works() {
...
+
+        let height = 1.into();
+        let predicate_data =
+            b"If you think it's simple, then you have misunderstood the problem."
+                .to_vec();
+
+        macro_rules! predicate_input {
+            ($predicate:expr) => {{
+                let predicate: Vec<u8> = $predicate.into_iter().collect();
+                let owner = Input::predicate_owner(&predicate, &ChainId::default());
```
xgreenx, line 191: I thought that this issue was related to `message_data_predicate` and `message_coin_predicate` inputs=D

Dentosal: Oops fixed the wrong thing =D. Moving back to a draft so I'll add tests fro those as well.

xgreenx (approved):

PR #595, opened 2023-10-02
Remove wee_alloc dependency, use builtin allocator instead

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/518

Everything seems to be working wihtout it as well.


xgreenx:
We started to use the `wee` allocator because it reduced the size of WASM binary in for 30-40%

xgreenx (approved):

Dentosal:
I wonder what's the impact now? How did we measure it before, i.e. what's the crate that was used to generate the wasm blob?

xgreenx (approved):

PR #598, opened 2023-10-03
Include size lookup of contract for computing LDC opcode cost

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1271


fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -759,7 +759,8 @@ where
 
             Instruction::LDC(ldc) => {
                 let (a, b, c) = ldc.unpack();
-                self.dependent_gas_charge(self.gas_costs().ldc, r!(c))?;
+                let contract_size = self.load_contract_size(r!(a))?;
+                self.dependent_gas_charge(self.gas_costs().ldc, r!(contract_size))?;
                 self.load_contract_code(r!(a), r!(b), r!(c))?;
```
bvrooman, line 764: If I understand correctly, the new approach is:
- Retrieve the contract size (in bytes) of the given contract (`$rA`) from storage
- Instead of charging for the user-requested size (potentially 0), we now charge for the amount of contract bytes retrieved from storage
- Finally, as before, load the contract code

Essentially, we now charge for the full size of the contract, regardless of how much is requested.

It seems like this approach makes sense given that we load contracts in their entirety:
```rust
pub(crate) fn contract<'s, S>(
    storage: &'s S,
    contract: &ContractId,
) -> ...
where
    S: InterpreterStorage,
{
    storage
        .storage_contract(contract)
    ...
}

``` 
rather than the requested size, anyway.

Have I understood this correctly?

fuel-vm/src/tests/blockchain.rs
```diff
@@ -453,78 +447,19 @@ fn ldc_reason_helper(
...
-        // Generate code for pushing contract id to heap
-        for (i, byte) in contract_id.as_ref().iter().enumerate() {
-            let index = i as Immediate12;
-            let value = *byte as Immediate12;
-            load_contract.extend([
-                op::xor(reg_a, reg_a, reg_a),    // r[a] := 0
-                op::ori(reg_a, reg_a, value),    // r[a] := r[a] | value
-                op::sb(RegId::HP, reg_a, index), // m[$hp+index] := r[a] (=value)
-            ]);
-        }
```
bvrooman, line 492: I'm not sure I understand what any of this was doing, or why we no longer need it. Can you provide any context here?

MitchTurner: This code was all in the `true` branch of the if/else. But we only ever used `false` in any of our tests. It's dead code.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -543,6 +554,26 @@ where
     }
 }
 
+struct ContractSizeCodeCtx<'vm, S> {
```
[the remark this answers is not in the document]
MitchTurner: It looks like we might be removing this anyway. Been talking to Green offline.

fuel-vm/src/interpreter/executors/instruction.rs
MitchTurner: That sounds all correct. I'm simplifying some of this code right now to get the last test to pass.

fuel-vm/src/tests/blockchain.rs
bvrooman: Gotcha 👍

MitchTurner:
It looks like the storage allows getting the contract bytes for a specified buffer, so we only copy what is specified with `$rC` now, not the entire contract :).

I've added a test that copies a contract with a bunch of `noop`s and checks the costs based on that.

Still might clean up these tests a bit. There is some DRYing I can do before merging this.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -96,6 +96,7 @@ where
     /// contract_code = contracts[contract_id]
     /// mem[$ssp, $rC] = contract_code[$rB, $rC]
     /// ```
+    /// Returns size of contract
```
bvrooman, line 99: Is this comment accurate? The return type appears to be `IoResult<(), S::DataError>`.

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -759,8 +759,9 @@ where
 
             Instruction::LDC(ldc) => {
                 let (a, b, c) = ldc.unpack();
-                self.dependent_gas_charge(self.gas_costs().ldc, r!(c))?;
-                self.load_contract_code(r!(a), r!(b), r!(c))?;
+                let len = r!(c);
+                self.load_contract_code(r!(a), r!(b), len)?;
+                self.dependent_gas_charge(self.gas_costs().ldc, len)?;
```
bvrooman, line 764: If this isn't a material change, I might suggest reverting it and keeping it as is.

fuel-vm/src/interpreter/contract.rs
```diff
@@ -154,6 +154,26 @@ where
         .ok_or_else(|| PanicReason::ContractNotFound.into())
 }
 
+/// Gets the bytes of the contract code from storage. If the contract is longer that
+/// specified `size` the contract will be shortened to `size` bytes. If the contract is
+/// shorter than `size` bytes, then the contract will stay it's original length.
```
bvrooman, line 159:
```suggestion
/// Gets the bytes of the contract code from storage. If the contract is longer than
/// `size` bytes the contract will be shortened to `size` bytes. If the contract is
/// shorter than `size` bytes, then the contract will stay its original length.
```

fuel-vm/src/interpreter/contract.rs
```diff
@@ -154,6 +154,26 @@ where
...
+pub(crate) fn contract_bytes<S>(
```
bvrooman, line 160: I might suggest keeping the name `contract`, since the return type is still a `Contract`. To me, `contract_bytes` implies a vector of bytes or something along those lines.

fuel-vm/src/tests/blockchain.rs
```diff
@@ -398,11 +398,165 @@ fn load_external_contract_code() {
...
+    }
+
+    // when
+    load_contract.extend([
+        op::move_(reg_a, RegId::HP),  // r[a] := $hp
+        op::xor(reg_b, reg_b, reg_b), // r[b] := 0
+        op::ori(reg_b, reg_b, len),   /* r[b] += 12 (will be
+                                       * padded to 16) */
+        op::ldc(reg_a, RegId::ZERO, reg_b), // Load first two words from the contract
+        op::move_(reg_a, RegId::SSP),       // r[b] := $ssp
```
bvrooman, line 505: Since this is `reg_a`, should the comment be:

```suggestion
        op::move_(reg_a, RegId::SSP),       // r[a] := $ssp
```

fuel-vm/src/tests/blockchain.rs
```diff
@@ -398,11 +398,165 @@ fn load_external_contract_code() {
...
+        op::subi(reg_a, reg_a, 8 * 2),      // r[a] -= 16 (start of the loaded code)
```
bvrooman, line 506:
```suggestion
        op::subi(reg_a, reg_a, 16),      // r[a] -= 16 (start of the loaded code)
```

fuel-vm/src/interpreter/contract.rs
MitchTurner: There is a separate function called `contract`. I had it named `contract_partial` before, but that's less clear IMO.

MitchTurner: What about something like `contract_subsection`?

fuel-vm/src/interpreter/executors/instruction.rs
MitchTurner: It's a pattern we use on other opcodes. I don't think it will make a difference in performance, but it's a little clearer IMO.

fuel-vm/src/interpreter/blockchain.rs
MitchTurner: Thanks. Left over from earlier iteration.

fuel-vm/src/interpreter/contract.rs
bvrooman: Cool, sounds good to me.

fuel-vm/src/tests/blockchain.rs
```diff
@@ -552,29 +560,37 @@ fn ldc_reason_helper(
...
-        op::ldc(RegId::ZERO, RegId::ZERO, reg_a), /* Load first two words from the
-                                      * contract */
-    ];
+    let (load_contract, _) = script_with_data_offset!(
+        data_offset,
+        vec![
+            op::movi(0x10, data_offset as Immediate18),
+            op::move_(reg_a, RegId::HP), // r[a] := $hp
+            op::ldc(0x10, RegId::ZERO, reg_a), /* Load first two words from the
+                                          * contract */
```
Dentosal, line 588: The comment is incorrect here. This loads first `$hp` bytes from the contract. The code should likely read like this:

```suggestion
            op::ldc(0x10, RegId::ZERO, RegId::HP),
```

fuel-vm/src/tests/blockchain.rs
```diff
@@ -552,29 +560,37 @@ fn ldc_reason_helper(
...
-        op::ldc(RegId::ZERO, RegId::ZERO, RegId::ZERO), /* Load first two words from
-                        * the contract */
-    ];
+    let (load_contract, _) = script_with_data_offset!(
+        data_offset,
+        vec![
+            op::movi(0x10, data_offset as Immediate18),
+            op::cfei(0x1), // sp += 1
+            op::ldc(0x10, RegId::ZERO, RegId::ZERO), /* Load first two words from *
+                            * the contract */
```
Dentosal, line 569: This loads zero words from the contract, so the comment is incorrect. I suggest loading a single byte anyway, since this test shouldn't be hitting the zero-length-load corner case.

```suggestion
            op::ldc(0x10, RegId::ZERO, RegId::ONE),
```

fuel-vm/src/tests/blockchain.rs
```diff
@@ -334,18 +420,19 @@ fn load_external_contract_code() {
...
+        op::move_(reg_a, RegId::HP),        // r[a] := $hp
+        op::xor(reg_b, reg_b, reg_b),       // r[b] := 0
+        op::ori(reg_b, reg_b, len),         // r[b] += len
         op::ldc(reg_a, RegId::ZERO, reg_b), // Load first two words from the contract
-        op::move_(reg_a, RegId::SSP),       // r[b] := $ssp
-        op::subi(reg_a, reg_a, 8 * 2),      // r[a] -= 16 (start of the loaded code)
+        op::move_(reg_a, RegId::SSP),       // r[a] := $ssp
+        op::subi(reg_a, reg_a, 16),         // r[a] -= 16 (start of the loaded code)
         op::xor(reg_b, reg_b, reg_b),       // r[b] := 0
         op::addi(reg_b, reg_b, 16),         // r[b] += 16 (length of the loaded code)
```
Dentosal, line 432:
```suggestion
        op::subi(reg_a, RegId::SSP, 16),    // r[a] := $ssp - 16 (start of the loaded code)
        op::movi(reg_b, 16),                // r[b] = 16 (length of the loaded code)
```

fuel-vm/src/tests/blockchain.rs
MitchTurner: Thanks. I think a lot of these are wrong, but I don't really understand them enough to fix them. I'm just copying what was already there.

Might be worth fixing all of them, maybe in another PR.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -499,8 +499,13 @@ where
...
 
         // Fetch the storage contract
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract = contract.as_ref().as_ref();
+        // If $rC is greater than the code size, zero bytes are filled in.
+        let contract = super::contract::contract_subsection(
+            self.storage,
+            &contract_id,
+            length_unpadded as usize,
+        )?;
```
xgreenx, line 507: Our specification says:

> Copy `$rC` bytes of code starting at `$rB` for contract with ID equal to the 32 bytes in memory starting at `$rA` into memory starting at `$ssp`.

You are missing `$rb` offset. We need to copy the subset starting from `[$rb, $rb + $rc]`.

xgreenx (changes requested):
We still load the whole contract into the memory before we write it into the buffer(we need to charge for it). Plus, if the buffer is less than the size of the value, you will receive an error.

image

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- [#598](https://github.com/FuelLabs/fuel-vm/pull/598): Update cost model for `ldc` opcode to take into account contract size.
```
xgreenx, line 11: It is a breaking change, because it ma break some unit tests/code that uses fixed gas.

MitchTurner:
> We still load the whole contract into the memory before we write it into the buffer(we need to charge for it). Plus, if the buffer is less than the size of the value, you will receive an error.

That's disappointing. I guess the trait definition doesn't exclude that possibility. I'm going to go back to explicitly loading the whole contract and getting the sub-bytes I guess :( It's much more expensive unfortunately. 

Looking at rocksdb docs, I don't see anything in opts that would give us what we want, but since this behind an interface, it allows us to upgrade this eventually if possible.

fuel-vm/src/interpreter/blockchain.rs
MitchTurner: I fixed this, but I still need to add test to check it is working.

MitchTurner: K. Added test.

fuel-vm/src/tests/blockchain.rs
```diff
@@ -268,27 +270,188 @@ fn state_read_write() {
...
+
+    let contract_size = 1000;
+    let offset = 0;
+    let len = 100;
+    let starting_gas_amount =
+        ldc__gas_cost_for_len(&mut client, rng, salt, contract_size, offset, len);
+
+    for i in 1..10 {
+        let offset = i * 4;
+        let expected_gas_used = starting_gas_amount - (i * noop_cost);
```
MitchTurner, line 365: Hmm. I just realized that this test shouldn't be passing--or at least I don't know why it's passing.

Since, the `len` is staying the same, each of the copied codes should be identical. Indeed, if I debug the code that's what I'm seeing. But for some reason execution of that code isn't identical.

In fact, locally, I've changed `contract_size` to be `25` so that the remaining bytes are empty. In that case I'd expect the test to pass, but it fails because each is 2 less than original, rather than just 1.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -498,9 +498,18 @@ where
...
+
         // Fetch the storage contract
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract = contract.as_ref().as_ref();
+        let contract_bytes: Vec<u8> =
+            super::contract::contract(self.storage, &contract_id)?
+                .into_owned()
+                .into();
+        let contract_len = contract_bytes.len();
+        let contract_sub_bytes =
```
xgreenx, line 511: I thought it was already handled by `copy_from_slice_zero_fill_noownerchecks`.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -498,9 +498,18 @@ where
...
+            .map_err(|_| PanicReason::MemoryOverflow)?;
+
         // Fetch the storage contract
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract = contract.as_ref().as_ref();
+        let contract_bytes: Vec<u8> =
+            super::contract::contract(self.storage, &contract_id)?
+                .into_owned()
+                .into();
+        let contract_len = contract_bytes.len();
```
xgreenx, line 510: It would be nice to charge the user here before we do the main work.

fuel-vm/src/tests/blockchain.rs
```diff
@@ -299,21 +462,23 @@ fn load_external_contract_code() {
...
+            .into_checked(height, &consensus_params)
+            .expect("failed to check tx");
 
     client.deploy(tx_create_target);
 
     // Then deploy another contract that attempts to read the first one
     let reg_a = 0x20;
-    let reg_b = 0x21;
+    let reg_c = 0x21;
+    let reg_b = 0x22;
```
xgreenx, line 481:
```suggestion
    let reg_b = 0x21;
    let reg_c = 0x22;
```

Maybe you want to have something like this=D

fuel-vm/src/tests/blockchain.rs
```diff
@@ -268,27 +270,188 @@ fn state_read_write() {
...
+    let contract_size = 0;
+    let offset = 0;
+    let len = 0;
+    let starting_gas_amount =
+        ldc__gas_cost_for_len(&mut client, rng, salt, contract_size, offset, len);
+
+    let bytes_per_op = 4;
+
+    for i in 1..10 {
+        let contract_size = contract_size + (i * ldc_dep_len / bytes_per_op) as usize;
```
xgreenx, line 398: It is better to name the variable `opcodes_number` or `number_of_noop_ops`

fuel-vm/src/tests/blockchain.rs
xgreenx: Indeed the gas amount should be the same. What is the value of `noop_cost`? Maybe it is zero

fuel-vm/src/interpreter/blockchain.rs
MitchTurner: Yes. And reverting to using this fixes the test bug.

fuel-vm/src/tests/blockchain.rs
MitchTurner: The bug was in the interpreter code. I was calling `copy_from_slice_zero_fill_noownerchecks` with the wrong values.

fuel-vm/src/interpreter/blockchain.rs
MitchTurner: I'll look into it. This would be quite the architectural change, I think, but if it's how it needs to work, then I'll make it work.

MitchTurner: K. I split the function in half to return the bytes and then copy them separately. I don't love the solution becuase the functions are doing too much, but it's probably better :)

xgreenx (dismissed):
@MitchTurner I moved charging for the contract size into the `load_contract_code`. It works in the same way as for `code_size`.

If you are okay with it, then you can merge the PR=)

fuel-vm/src/tests/blockchain.rs
```diff
@@ -334,18 +500,20 @@ fn load_external_contract_code() {
...
+        op::move_(reg_a, RegId::HP),   // r[a] := $hp
+        op::xor(reg_b, reg_b, reg_b),  // r[b] := 0
+        op::ori(reg_b, reg_b, offset), // r[b] += offset
+        op::xor(reg_c, reg_c, reg_c),  // r[b] := 0
+        op::ori(reg_c, reg_c, len),    // r[b] += len
+        op::ldc(reg_a, reg_b, reg_c),  // Load first two words from the contract
+        op::subi(reg_a, RegId::SSP, 16), /* r[a] := $ssp - 16 (start of the loaded
+                                        * code) */
+        op::movi(reg_c, 16), // r[b] = 16 (length of the loaded code)
+        op::logd(RegId::ZERO, RegId::ZERO, reg_a, reg_c), /* Log digest of the
```
bvrooman, line 514: It looks like you're loading `len` bytes from the contract, given that `len` is stored in `reg_c`. In the tests, you pass `0` for `len`, but the following opcodes hardcode `16`. If I understand correctly, `ldc` still loads the whole contract regardless of `$rC`, but I think the `subi`, `movi`, and `logd` opcodes here become incorrect. Are they even necessary for these tests? Maybe we can remove them.

MitchTurner: It fails the `ldc__load_external_contract_code` test without it.

But you're completely right. I think it just needs to be parameterized for the `len`. It just so happens that the padded length for that test is 16 :clown_face:

fuel-vm/src/tests/blockchain.rs
```diff
@@ -334,20 +508,28 @@ fn load_external_contract_code() {
...
+        op::ldc(reg_a, reg_b, reg_c),  // Load first two words from the contract
     ]);
 
+    if include_log_d {
+        let padded_len = pad(len);
+        load_contract.extend([
+            op::subi(reg_a, RegId::SSP, padded_len), /* r[a] := $ssp - 16 (start of
+                                                      * the loaded
+                                                      * code) */
+            op::movi(reg_c, padded_len as u32), // r[b] = 16 (length of the loaded code)
```
xgreenx, line 525: Maybe you want to update a comment regarding `16`

xgreenx (dismissed):

xgreenx (approved):

PR #603, opened 2023-10-09
Improved Memory Efficiency in Merkle Root Calculation

Salka1988:
This PR addresses the memory consumption concern in the Merkle root calculation process. By employing a stack-based approach and optimizing node merging, we've achieved a notable reduction in memory usage.

- Created a `MerkleRootCalculator` struct with a stack-based algorithm.
- Added a `push` method to efficiently add nodes before handling node merging and removing the merged ones for optimized memory usage.
- Implemented a `root` method for the final Merkle root calculation.


fuel-merkle/src/binary/in_memory.rs
```diff
@@ -62,6 +64,80 @@ impl Default for MerkleTree {
...
+pub struct MerkleRootCalculator {
+    stack: Vec<Node>,
+    second_top_on_stack: Option<Node>,
+}
+
+impl MerkleRootCalculator {
+    pub fn new() -> Self {
+        Self {
+            stack: Vec::new(),
+            second_top_on_stack: None,
```
xgreenx, line 77: Internally I feel that we should be able to do that without an additional `second_top_on_stack` field. The stack structure itself and knowing each height should be enough to know when we need to merge.

@bvrooman  Could you validate the algorithm, please? Maybe you can see how we can simplify it

xgreenx (commented):
Thank you! I think we can reuse optimized version for header's transaction and message root calculation too, not only for regenesis=)

xgreenx:
@Salka1988 May I ask you to update all places where we use BMT with your version, please? It seems we only need a final root in most cases.

Salka1988:
> @Salka1988 May I ask you to update all places where we use BMT with your version, please? It seems we only need a final root in most cases.

Ok

fuel-merkle/src/binary/in_memory.rs
Salka1988: removed

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - [#595](https://github.com/FuelLabs/fuel-vm/pull/595): Removed `wee_alloc` dependency from `fuel-asm`. It now uses the builtin allocator on web targets as well.
+- [#603](https://github.com/FuelLabs/fuel-vm/pull/603): Added `MerkleRootCalculator`for efficient in-memory merkle root calculation.
```
xgreenx, line 12: It fits better for `Added` section=D

fuel-merkle/src/binary/in_memory.rs
```diff
@@ -62,6 +67,72 @@ impl Default for MerkleTree {
...
+    pub fn new() -> Self {
+        Self { stack: Vec::new() }
+    }
+
+    pub fn push(&mut self, data: &[u8]) {
+        let node = Node::create_leaf(0, data);
+        self.stack.push(node);
+
+        if let Some(top_node) = self.stack.last() {
+            if self.stack.len() > 1 {
```
xgreenx, line 85:
```suggestion\
            if self.stack.len() > 1 {
```

It already should be enough=)

fuel-merkle/src/binary/in_memory.rs
```diff
@@ -62,6 +67,72 @@ impl Default for MerkleTree {
...
+                                    if last_merged.height() == second_last_merged.height()
+                                    {
+                                        let merged_node = Node::create_node(
+                                            second_last_merged,
+                                            last_merged,
+                                        );
+                                        self.stack.pop();
+                                        self.stack.pop();
+                                        self.stack.push(merged_node);
+                                    } else {
```
xgreenx, line 106: You duplicate the same logic before this loop. I think you can have one loop that does these checks.

Salka1988: You're absolutely right! I overlooked this. 😅 At one point I needed both...

fuel-merkle/src/binary/in_memory.rs
```diff
@@ -62,6 +67,54 @@ impl Default for MerkleTree {
...
+                        break
+                    }
+                }
+            } else {
+                break
+            }
+        }
+    }
+
+    pub fn root(&mut self) -> Bytes32 {
```
bvrooman, line 102: This should probably consume `self` since you are modifying the internal stack:

```suggestion
    pub fn root(self) -> Bytes32 {
```

bvrooman (commented):
I think the algorithm looks fine and simple enough in its current incarnation.

I recommend:
- Moving the algorithm to its own file under the `binary` module (not under `in_memory`)
- Exposing it as a single function that takes in an iterator and returns the root
- Adding tests

fuel-merkle/src/binary/in_memory.rs
```diff
@@ -62,6 +67,54 @@ impl Default for MerkleTree {
...
+    stack: Vec<Node>,
+}
+
+impl MerkleRootCalculator {
+    pub fn new() -> Self {
+        Self { stack: Vec::new() }
+    }
+
+    pub fn push(&mut self, data: &[u8]) {
+        let node = Node::create_leaf(0, data);
```
bvrooman, line 81: I am a little surprised that this works without a specifying the correct leaf index. But if it works, it works!

CHANGELOG.md
```diff
@@ -6,11 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [#603](https://github.com/FuelLabs/fuel-vm/pull/603): Added `MerkleRootCalculator`for efficient in-memory merkle root calculation.
```
bvrooman, line 10:
```suggestion
- [#603](https://github.com/FuelLabs/fuel-vm/pull/603): Added `MerkleRootCalculator`for efficient in-memory Merkle root calculation.
```

fuel-merkle/src/binary/in_memory.rs
Salka1988:
```rust
pub fn root(mut self) -> Bytes32 {
```
has to be `mut` without `&`

Salka1988:
> I think the algorithm looks fine and simple enough in its current incarnation.
> 
> I recommend:
> 
> * Moving the algorithm to its own file under the `binary` module (not under `in_memory`)
> * Exposing it as a single function that takes in an iterator and returns the root
> * Adding tests

we need push() and root() exposed. I will add new function that will accept iterator

fuel-merkle/src/binary/mrc.rs
```diff
@@ -0,0 +1,140 @@
...
+        Self { stack: Vec::new() }
+    }
+
+    pub fn push(&mut self, data: &[u8]) {
+        let node = Node::create_leaf(0, data);
+        self.stack.push(node);
+
+        while let Some(top_node) = self.stack.last() {
+            if self.stack.len() > 1 {
+                if let Some(second_top_node) = self.stack.get(self.stack.len() - 2) {
```
xgreenx, line 28:
```suggestion
        while self.stack.len() > 1 {
            let top_node = self.stack[self.stack.len() - 1];
            let second_top_node = self.stack[self.stack.len() - 2];
```

It seems we can also remove too `if` statements=)

fuel-merkle/src/binary.rs
```diff
@@ -17,4 +17,6 @@ pub use merkle_tree::{
 };
 pub use primitive::Primitive;
 pub mod in_memory;
+pub mod mrc;
```
xgreenx, line 20:
```suggestion
pub mod root_calculator;
```

xgreenx (commented):
LGTM! Could you simplify it one more time and I think we can merge it=)

xgreenx (approved):

PR #605, opened 2023-10-11
feat: Prepare `no_std` compatibility in `fuel-core`

bvrooman:


fuel-vm/Cargo.toml
```diff
@@ -23,7 +23,7 @@ fuel-asm = { workspace = true, default-features = false }
 fuel-crypto = { workspace = true, default-features = false }
 fuel-merkle = { workspace = true, default-features = false }
 fuel-storage = { workspace = true }
-fuel-tx = { workspace = true, default-features = false }
+fuel-tx = { workspace = true, default-features = false, features = ["alloc", "builder"] }
```
xgreenx, line 26: Why do you need to enable these features on this level?

PR #607, opened 2023-10-12
Add ECAL instruction

Dentosal:
First draft of https://github.com/FuelLabs/fuel-specs/pull/523

Currently uses a `EcalAccess` trait to work around the need to have same vm interface regardless of the generics. This is a bit ugly as it leads to code duplication, but might also be a nice way to definite the interface for external functions.

### DIFF
--- a/fuel-vm/src/interpreter/ecal.rs
+++ b/fuel-vm/src/interpreter/ecal.rs
@@ -4,10 +4,17 @@ use fuel_types::Word;
 
 use crate::{
     call::CallFrame,
+    constraints::reg_key::{
+        split_registers,
+        SystemRegisters,
+    },
     error::SimpleResult,
 };
 
-use super::Interpreter;
+use super::{
+    internal::inc_pc,
+    Interpreter,
+};
 
 pub trait EcalAccess {
     // Accessors
@@ -84,8 +91,7 @@ fn noop_ecall(
     _: RegId,
     _: RegId,
 ) -> SimpleResult<()> {
-    vm.gas_charge(vm.gas_costs().noop)?;
-    Ok(())
+    vm.gas_charge(vm.gas_costs().noop)
 }
 
 impl<S, Tx> Interpreter<S, Tx> {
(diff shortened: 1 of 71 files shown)


fuel-vm/src/interpreter/ecal.rs
```diff
@@ -0,0 +1,93 @@
...
+use fuel_types::Word;
+
+use crate::{
+    call::CallFrame,
+    error::SimpleResult,
+};
+
+use super::Interpreter;
+
+pub trait EcalAccess {
```
Dentosal, line 12: If anyone can figure out a better way than having this trait, I'd be delighted to hear about it

MitchTurner:
I'm trying to understand the point of this opcode. I've read through the parent issue, and it seems like it's meant to extend the VM with arbitrary functions? Could you give a concrete example of what that would look like?

My first impression is that this wouldn't be useful if the extension only existed on your local machine, and would be a giant attack vector if it was propagatable to other machines.

Voxelot:
> I'm trying to understand the point of this opcode. I've read through the parent issue, and it seems like it's meant to extend the VM with arbitrary functions? Could you give a concrete example of what that would look like?
> 
> My first impression is that this wouldn't be useful if the extension only existed on your local machine, and would be a giant attack vector if it was propagatable to other machines.

The use case is for being able to use the vm/sway in other contexts off-chain such as indexing, or for other projects that may want to customize the VM for their network in an extensible way without forking the VM.

Dentosal:
> I'm trying to understand the point of this opcode. I've read through the parent issue, and it seems like it's meant to extend the VM with arbitrary functions? Could you give a concrete example of what that would look like?

The imporatant part is that the arbitrary function is provided by the Rust code constructing the VM. It's simply an instruction that can be changed without having to fork FuelVM codebase.

fuel-vm/src/interpreter/ecal.rs
xgreenx: I would prefer more the approach with `Ecall`/`External` generic in `Interpreter`. It is more perform, plus we don't need to define the trait with a limited number of functions that should be implemented by the `Interpreter`. Instead, the developer can have access to all public functions of the `Interpreter`.

[Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1a86cb5176f88c1f46ad53bac271ec3f) is an example of how it can work.

Dentosal: Done in 51150738. I feel like it's super ugly as it pollutes all `impl` singatures of `Interpreter` and `Transactor`. Default value doesn't help us much.

xgreenx (approved):

xgreenx (dismissed):

xgreenx (approved):

Issue #599, opened 2023-10-05
Charge transaction for signatures verification as part of `min_fee`


xgreenx:
## Problem overview

ECR is the most expensive opcode in the VM (~1700 noops). However, we charge significantly less than that currently for verifying signatures used by coin or message inputs. This is partly because witness data is not chargeable. If we were to charge for signatures in witness data and also increase the GAS_PER_BYTE factor until the signature cost is adequately captured, this would disproportionately increase the cost of other bytes which should otherwise be cheap. We should have an independent charge which is based on the number of signatures a transaction has to verify, instead of warping the cost of bytes to cover it.

## Implementation details

The `check_signatures` method uses `is_predicate_owner_valid` and `recover_witness` under the hood(based on the type of inputs). They are the most expensive operations during the verification of the signatures.

For `recover_witness` we can use the the gas cost of the `ecr1` opcode. 

For `is_predicate_owner_valid`, it is more tricky because we use BMT under the hood + hashing function. For BMT we can either come up with a logarithmic formula or use hardcoded gas based on the worst possible scenario(the maximum size of the contract is 17 MB, while the predicate is less).

Both should affect the final `min_fee` in `TransactionFee::checked_from_values`. It may require some architectural modification into the `TransactionFee` structure because it requires inputs to calculate `min_fee`.

bvrooman:
Closed by https://github.com/FuelLabs/fuel-vm/pull/613 and  https://github.com/FuelLabs/fuel-vm/pull/618

PR #613, opened 2023-10-22
feat: Charge for input signature verification (address recovery and predicate roots)

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-vm/issues/599

This PR adds support for charging users for recovering addresses from signatures of signed inputs. For each signed input (non-predicate in effect), we charge the cost of an EC recovery. For each predicate input, we calculate the gas cost based on the length of the bytecode. The calculation is based on the gas cost of calculating the Merkle root of contract bytecode, which is dependent on contract length.  

This PR uses the gas price generated by the benchmark in [this PR](https://github.com/FuelLabs/fuel-core/pull/1452). This benchmark calculates a gas cost by timing contract root calculation using the maximum size for contracts at 17mb.

### DIFF
--- a/fuel-tx/src/transaction.rs
+++ b/fuel-tx/src/transaction.rs
@@ -726,6 +726,7 @@ pub mod field {
     pub trait Inputs {
         fn inputs(&self) -> &Vec<Input>;
         fn inputs_mut(&mut self) -> &mut Vec<Input>;
+        fn num_inputs(&self) -> usize;
         fn inputs_offset(&self) -> usize;
 
         /// Returns the offset to the `Input` at `idx` index, if any.
--- a/fuel-tx/src/transaction.rs
+++ b/fuel-tx/src/transaction.rs
@@ -726,7 +726,6 @@ pub mod field {
     pub trait Inputs {
         fn inputs(&self) -> &Vec<Input>;
         fn inputs_mut(&mut self) -> &mut Vec<Input>;
-        fn num_inputs(&self) -> usize;
         fn inputs_offset(&self) -> usize;
 
         /// Returns the offset to the `Input` at `idx` index, if any.
--- a/fuel-tx/src/transaction.rs
+++ b/fuel-tx/src/transaction.rs
@@ -556,6 +556,7 @@ pub mod field {
         Deref,
         DerefMut,
     };
+    use itertools::Itertools;
 
     pub trait GasPrice {
         fn gas_price(&self) -> &Word;
@@ -733,6 +734,25 @@ pub mod field {
 
         /// Returns predicate's offset and length of the `Input` at `idx`, if any.
         fn inputs_predicate_offset_at(&self, idx: usize) -> Option<(usize, usize)>;
+
+        /// Returns all signed inputs. This excludes predicate inputs.
+        fn signed_inputs(&self) -> Vec<&Input> {
+            self.inputs()
+                .iter()
+                .filter_map(|input| input.witness_index().is_some().then_some(input))
+                .collect_vec()
+        }
+
+        /// Returns all signed inputs filtered by witness uniqueness.
+        fn signed_inputs_with_unique_witnesses(&self) -> Vec<&Input> {
+            self.signed_inputs()
+                .into_iter()
+                .sorted_unstable_by(|a, b| {
+                    Ord::cmp(&a.witness_index(), &b.witness_index())
+                })
+                .dedup()
+                .collect_vec()
+        }
     }
 
     pub trait Outputs {
--- a/fuel-tx/src/transaction.rs
+++ b/fuel-tx/src/transaction.rs
@@ -556,7 +556,6 @@ pub mod field {
         Deref,
         DerefMut,
     };
-    use itertools::Itertools;
 
     pub trait GasPrice {
         fn gas_price(&self) -> &Word;
@@ -735,24 +734,24 @@ pub mod field {
         /// Returns predicate's offset and length of the `Input` at `idx`, if any.
         fn inputs_predicate_offset_at(&self, idx: usize) -> Option<(usize, usize)>;
 
-        /// Returns all signed inputs. This excludes predicate inputs.
-        fn signed_inputs(&self) -> Vec<&Input> {
-            self.inputs()
-                .iter()
-                .filter_map(|input| input.witness_index().is_some().then_some(input))
-                .collect_vec()
-        }
-
-        /// Returns all signed inputs filtered by witness uniqueness.
-        fn signed_inputs_with_unique_witnesses(&self) -> Vec<&Input> {
-            self.signed_inputs()
-                .into_iter()
-                .sorted_unstable_by(|a, b| {
-                    Ord::cmp(&a.witness_index(), &b.witness_index())
-                })
-                .dedup()
-                .collect_vec()
-        }
+        // /// Returns all signed inputs. This excludes predicate inputs.
+        // fn signed_inputs(&self) -> Vec<&Input> {
+        //     self.inputs()
+        //         .iter()
+        //         .filter_map(|input| input.witness_index().is_some().then_some(input))
+        //         .collect_vec()
+        // }
+        //
+        // /// Returns all signed inputs filtered by witness uniqueness.
+        // fn signed_inputs_with_unique_witnesses(&self) -> Vec<&Input> {
+        //     self.signed_inputs()
+        //         .into_iter()
+        //         .sorted_unstable_by(|a, b| {
+        //             Ord::cmp(&a.witness_index(), &b.witness_index())
+        //         })
+        //         .dedup()
+        //         .collect_vec()
+        // }
     }
 
     pub trait Outputs {
--- a/fuel-tx/src/transaction.rs
+++ b/fuel-tx/src/transaction.rs
@@ -733,25 +733,6 @@ pub mod field {
 
         /// Returns predicate's offset and length of the `Input` at `idx`, if any.
         fn inputs_predicate_offset_at(&self, idx: usize) -> Option<(usize, usize)>;
-
-        // /// Returns all signed inputs. This excludes predicate inputs.
-        // fn signed_inputs(&self) -> Vec<&Input> {
-        //     self.inputs()
-        //         .iter()
-        //         .filter_map(|input| input.witness_index().is_some().then_some(input))
-        //         .collect_vec()
-        // }
-        //
-        // /// Returns all signed inputs filtered by witness uniqueness.
-        // fn signed_inputs_with_unique_witnesses(&self) -> Vec<&Input> {
-        //     self.signed_inputs()
-        //         .into_iter()
-        //         .sorted_unstable_by(|a, b| {
-        //             Ord::cmp(&a.witness_index(), &b.witness_index())
-        //         })
-        //         .dedup()
-        //         .collect_vec()
-        // }
     }
 
     pub trait Outputs {
--- a/fuel-tx/src/transaction/consensus_parameters/gas.rs
+++ b/fuel-tx/src/transaction/consensus_parameters/gas.rs
@@ -339,6 +339,8 @@ pub struct DependentCost {
 
 impl DependentCost {
     pub fn resolve(&self, units: Word) -> Word {
+        // Apply the linear transformation from units to cost:
+        // f(x) = base + mx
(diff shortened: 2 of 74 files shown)


fuel-tx/src/transaction.rs
```diff
@@ -733,6 +734,25 @@ pub mod field {
...
+        /// Returns all signed inputs filtered by witness uniqueness.
+        fn signed_inputs_with_unique_witnesses(&self) -> Vec<&Input> {
+            self.signed_inputs()
+                .into_iter()
+                .sorted_unstable_by(|a, b| {
+                    Ord::cmp(&a.witness_index(), &b.witness_index())
+                })
+                .dedup()
+                .collect_vec()
+        }
```
xgreenx, line 755: It is better to not add one-time functions into the trait. We only need them in one place, plus we need to iterate over all inputs because you need to charge for `Signed` and for `Predicate` inputs

fuel-tx/src/transaction/fee.rs
```diff
@@ -113,18 +120,23 @@ impl TransactionFee {
...
     ///
     /// Will return `None` if arithmetic overflow occurs.
-    pub fn checked_from_tx<T: Chargeable>(
+    pub fn checked_from_tx<T: Chargeable + Inputs>(
+        gas_costs: &GasCosts,
         params: &FeeParameters,
         tx: &T,
     ) -> Option<Self> {
         let metered_bytes = tx.metered_bytes_size() as Word;
+        let unique_witnesses = tx.signed_inputs_with_unique_witnesses().len() as Word;
```
xgreenx, line 129: Because you need to find `Signed` and `Predicate` inputs, it is better to iterate over them here and calculate the number of unique witnesses, plus charge for predicates.

You can use one loop to cover all inputs. Plus, you can already implement charging for the predicates too, but with some dummy values that you will update later.

fuel-tx/src/transaction/fee.rs
```diff
@@ -113,18 +120,23 @@ impl TransactionFee {
     /// Attempt to create a transaction fee from parameters and transaction internals
     ///
     /// Will return `None` if arithmetic overflow occurs.
-    pub fn checked_from_tx<T: Chargeable>(
+    pub fn checked_from_tx<T: Chargeable + Inputs>(
+        gas_costs: &GasCosts,
```
xgreenx, line 124: Regarding predicates, you can already add a new cost into `GasCosts` and add a new structure that we can use later to charge for the predicates. 

Maybe you want to extend `FeeParameters`, for example, the price of the `ecr1` can be duplicated there. The constructor can accept `gas_costs`, like `FeeParameters::new(gas_costs)`.

bvrooman: We could add new fields for both the cost of signed input address recovery and the cost of predicate ownership check. The advantage of this is that we don't have to pass `GasCosts` around, which is a little bit simpler. 

On the other hand, it is not clear to me why that should be the responsibility of `FeeParameters` when `GasCosts` already contains these values. `FeeParameters` is responsible for providing information about how to convert gas to fees, rather than storing any concrete gas costs themselves.

fuel-tx/src/transaction.rs
bvrooman: I understand your suggestion. Instead specializing the trait here, I can just iterate over the inputs at the time of calculating the fee, as you suggest in another comment. 

Currently, this method is also used in tests to verify the fee for a transaction is correct. But maybe I can refactor the tests so that we can test the fees without duplicating the calculation logic, and we can keep the calculation isolated to the one place.

bvrooman:
I will put this PR back into draft as I apply the following changes:
- refactor calculation to use a single loop over the inputs, checking for both signed inputs and predicates at this time, rather than separately
- adding dummy value for predicate costs
- (maybe) refactoring `FeeParameters`

fuel-tx/src/transaction/fee.rs
MitchTurner: I will point out that the calculation for predicate gas is done by the `tx`:
```
tx.gas_used_by_predicates()
```
whereas the signature cost is done here.

Would it be wrong to have a
```
tx.gas_used_for_sign_verification()
```
method on the `tx`? And if yes, would it make sense to change the predicate calculation?

I don't really have an opinion yet, but seems strange that they are so different.

fuel-tx/src/transaction/fee.rs
```diff
@@ -159,68 +171,87 @@ mod tests {
...
 
+    fn gas_to_fee(gas: u64, gas_price: Word) -> f64 {
+        let fee = gas * gas_price;
+        fee as f64 / PARAMS.gas_price_factor as f64
+    }
+
     #[test]
     fn base_fee_is_calculated_correctly() {
         let metered_bytes = 5;
+        let gas_used_by_signature_recovery = 13;
```
MitchTurner, line 182: Why does this one cost more than the other tests?

fuel-tx/src/transaction/fee.rs
```diff
@@ -159,68 +171,87 @@ mod tests {
         .with_gas_per_byte(2)
         .with_gas_price_factor(3);
 
+    fn gas_to_fee(gas: u64, gas_price: Word) -> f64 {
+        let fee = gas * gas_price;
+        fee as f64 / PARAMS.gas_price_factor as f64
+    }
+
     #[test]
     fn base_fee_is_calculated_correctly() {
```
MitchTurner, line 180: Might be worth considering adding multi-sig tests?

fuel-tx/src/transaction/fee.rs
bvrooman: I think it may be beneficial to move this to the `Chargeable` trait, and allow the method to take in a reference to `GasCosts`. We calculate the gas cost when checking the transaction, but also in a number of tests to verify the fee. Ideally, we want to test the calculation used in production, rather than testing a recreation of the calculation that may become out of sync. This means we need to wrap it in a method that we can call on the tx.

fuel-tx/src/transaction/fee.rs
bvrooman: I added additional [tests for multiple signed inputs](https://github.com/FuelLabs/fuel-vm/pull/613/files#diff-a9aa3c1302258ae48cda7b9cb2998209f3b240a4bc9f64e58d221d97dd2488d1R908). Is that what you mean by multi-sig tests?

fuel-tx/src/transaction/types/create.rs
```diff
@@ -183,6 +193,43 @@ impl Chargeable for Create {
         }
         cumulative_predicate_gas
     }
+
+    fn gas_used_by_signature_checks(&self, gas_costs: &GasCosts) -> Word {
+        let mut witness_cache: HashMap<u8, bool> =
```
xgreenx: You cam use `HasSet` instead

fuel-vm/src/checked_transaction.rs
```diff
@@ -1146,20 +1363,21 @@ mod tests {
...
-            balances::initial_free_balances(tx, fee_params, base_asset_id)?;
+            balances::initial_free_balances(tx, gas_costs, fee_params, base_asset_id)?;
         // cant overflow as metered bytes * gas_per_byte < u64::MAX
-        let bytes = (tx.metered_bytes_size() as u128)
+        let fee = tx.metered_bytes_size() as u128
             * fee_params.gas_per_byte as u128
             * tx.price() as u128;
         let gas = tx.limit() as u128 * tx.price() as u128;
-        let total = bytes + gas;
+        let total = fee + gas;
```
xgreenx, line 1380: `max_fee` it is `min_fee` + (`GasLimit` converted into the price). So you need to update it in the same way as for `is_valid_min_fee`=)

fuel-tx/src/transaction/consensus_parameters/gas/default_gas_costs.rs
```diff
@@ -161,5 +161,6 @@ pub fn default_gas_costs() -> GasCostsValues {
             base: 44,
             dep_per_unit: 5,
         },
+        contract_root: 3024932,
```
xgreenx, line 164: Hmm, it is a very huge value. Maybe it is better if we use `DependentCost` instead. Could you check how expensive it is in this case, please?

fuel-vm/src/checked_transaction.rs
bvrooman: If I understand correctly:
```
max_fee = (min_fee + gas_limit) * gas_price
```

That would mean that `max_fee` is always greater than or equal to `min_fee`. 

After making that change, some tests fail, such as:
```rust
    #[test]
    fn base_fee_gas_limit_less_than_gas_used_by_predicates() {
        let metered_bytes = 5;
        let gas_used_by_signature_checks = 12;
        let gas_used_by_predicates = 8;
        let gas_limit = 7;
        let gas_price = 11;

        let fee = TransactionFee::checked_from_values(
            &PARAMS,
            metered_bytes,
            gas_used_by_signature_checks,
            gas_used_by_predicates,
            gas_limit,
            gas_price,
        )
        .expect("failed to calculate fee");

        assert!(fee.min_fee > fee.max_fee);
    }
 ``` 
where we test that the `min_fee` is greater than `max_fee`. 

Is there a specification for min/max fee?

File: fuel-tx/src/transaction/fee.rs [hunk 9]
```diff
@@ -74,8 +93,10 @@ impl TransactionFee {
 
         // TODO: use native div_ceil once stabilized out from nightly
         let bytes_gas = params.gas_per_byte.checked_mul(metered_bytes)?;
-        let min_gas = bytes_gas.checked_add(gas_used_by_predicates)?;
-        let max_gas = bytes_gas.checked_add(gas_limit)?;
+        let min_gas = bytes_gas
+            .checked_add(gas_used_by_signature_checks)?
+            .checked_add(gas_used_by_predicates)?;
+        let max_gas = min_gas.checked_add(gas_limit)?;
```
bvrooman, line 99: Fee is effectively calculated using [the formula defined in the spec](https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/tx-validity.md#script-execution). Min fee and max fee are separated by the gas limit. 

In effect, `max_fee = min_fee + gas_limit * gas_price`.

fuel-tx/src/transaction/fee.rs
```diff
@@ -308,23 +426,4 @@ mod tests {
...
-            &PARAMS,
-            metered_bytes,
-            gas_used_by_predicates,
-            gas_limit,
-            gas_price,
-        )
-        .expect("failed to calculate fee");
-
-        assert!(fee.min_fee > fee.max_fee);
-    }
```
bvrooman, line 329: I am removing this test because it is now no longer possible for `min_fee` to be greater than `max_fee`.

fuel-tx/src/transaction/consensus_parameters/gas.rs
```diff
@@ -334,6 +337,14 @@ pub struct DependentCost {
     pub dep_per_unit: Word,
 }
 
+impl DependentCost {
+    pub fn resolve(&self, units: Word) -> Word {
+        // Apply the linear transformation from units to cost:
+        // f(x) = base + mx
+        self.base + self.dep_per_unit.saturating_mul(units)
```
xgreenx, line 344: The formula is `base + units/dep_pet_unit`.


```suggestion
        self.base + units.saturating_div(self.dep_per_unit)
```

image

bvrooman: How do we handle the case where `dep_per_unit` is 0, as with `DependentCost::unit()`?

xgreenx: In production, we don't have cases like this=) But in the tests it overflows

fuel-tx/src/transaction/consensus_parameters/gas/default_gas_costs.rs
```diff
@@ -161,5 +161,9 @@ pub fn default_gas_costs() -> GasCostsValues {
             base: 44,
             dep_per_unit: 5,
         },
+        contract_root: DependentCost {
+            base: 75,
+            dep_per_unit: 1,
+        },
```
xgreenx, line 167: I got other results from benchmarks=) But it's okay. We can finalize them later before the final gas prices update.

image

xgreenx (approved):

Issue #616, opened 2023-10-27
Limit `$pc` to always stay within `$ssp`..`$sp`


xgreenx:
It should resolve the https://github.com/FuelLabs/fuel-vm/issues/564

dmihal:
How does this affect upgradability or other use cases where it's desirable to execute arbitrary code?

Dentosal:
This doesn't work at all. Firstly, VM is initialized `$ssp=$sp` and `$pc < $ssp`. Secondly, you can still write to stack.

xgreenx:
It seems the rule like `$pc < $ssp` should work?=D

Dentosal:
Likely. I'm trying to think it through.

Dentosal:
Feels like `$is <= $pc < $ssp` would be the bounds we want

Voxelot:
This makes more sense and works great for contract calls. My only concern is for script execution, since the script bytecode isn't the only data in-between `$is..$ssp`. Granted, this isn't really much of a security risk, but it could lead to weird things like the vm trying to parse random transaction bits as instructions and cause unexpected behavior.

Issue #564, opened 2023-08-28
TOB-FUEL-26: Memory is executable


xgreenx:
## Description
The interpreter’s memory is executable. The figure 26.1 demonstrates this by writing the instruction RET 0x1 to the heap and then jumping to it. This is problematic as it enables compilers to produce code which utilize this behavior. Programs might become vulnerable to classical security vulnerabilities like heap or stack buffer-overflows which are able to change the programs control flow.

#### Figure 26.1: Proof of concept for executing code on the heap.

```assembler
// allocate
op::movi(0x13, 8),
op::aloc(0x13),
// write RET op
op::movi(0x14, 0x24),
op::sb(RegId::HP, 0x14, 0),
// write 0x1 register for return code
op::movi(0x14, 0x1 << 2),
op::sb(RegId::HP, 0x14, 1),
// Set jump target to beginning of heap
op::move_(0x11, RegId::HP),
// Subtract $is
op::sub(0x11, 0x11, RegId::IS),
op::divi(0x11, 0x11, 4),
// Jump
op::jmp(0x11),
```

A further demonstration of this behavior can be seen in the following example (see figure 26.2). The script allocates some stack memory, which is larger than the call frame of the internal subcontract. Note, that the subcontract is vulnerable here, because it does not finish with a RET instruction. Then the contract writes executable code just after the call frame and code of the subcontract which will be called later in the script. After freeing the stack, the subcontract is called. Code which has been written by the script is executed in the context of the internal contract.

Note that the code size of the calling script must be padded to align to words.

#### Figure 26.2: Unit test which demonstrates executing script controlled code in the internal contract.

```rust
let gas_limit = 10_000_000;
let asset_id: AssetId = rng.gen();
let call_amount = 10u64;
let mut test_context = TestBuilder::new(2322u64);
let subcontract = vec![
    // op::ret(RegId::BAL) // no return code
];
let contract_id = test_context.setup_contract(subcontract.clone(), None, None)
    .contract_id;
let (script_ops, offset) = script_with_data_offset!(
    data_offset,
vec![
// pad
op::noop(),
        // over allocate
        op::cfei(1024),
        // attack address
        op::movi(0x15, 11648), // fp of called contract + frame size + code size
        // ret op code
        op::movi(0x14, 0x24),
        op::sb(0x15, 0x14, 0),
         // ret value (reg 1)
        op::movi(0x14, 1 << 2),
        op::sb(0x15, 0x14, 1),
        // deallocate
[...]

PR #617, opened 2023-10-27
Memory write xor execute

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/616. Closes https://github.com/FuelLabs/fuel-vm/issues/564.

This PR makes attempting to execute opcodes outside `$is..$ssp` range panic with `MemoryNotExecutable` error. It also separates `ErrorFlag` into `InvalidFlags` and `InvalidInstruction`, which make more sense. Then finally it fixes some tests that incorrectly assumed that any of these were equivalent, and some that had even weirder misconceptions.

### DIFF
--- a/fuel-vm/src/interpreter/executors/instruction.rs
+++ b/fuel-vm/src/interpreter/executors/instruction.rs
@@ -96,7 +96,7 @@ where
         raw: RawInstruction,
     ) -> IoResult<ExecuteState, S::DataError> {
         let instruction = Instruction::try_from(raw)
-            .map_err(|_| RuntimeError::from(PanicReason::ErrorFlag))?;
+            .map_err(|_| RuntimeError::from(PanicReason::InvalidInstruction))?;
 
         // TODO additional branch that might be optimized after
         // https://github.com/FuelLabs/fuel-asm/issues/68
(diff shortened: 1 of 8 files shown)


fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -38,24 +39,42 @@ where
...
+        &self,
+    ) -> Result<RawInstruction, InterpreterError<S::DataError>> {
+        let pc = self.registers[RegId::PC];
+        let instruction = self.fetch_instruction_bytes(pc).ok_or(
+            InterpreterError::PanicInstruction(PanicInstruction::error(
+                PanicReason::MemoryOverflow,
+                0, // The value is meaningless since fetch was out-of-bounds
+            )),
+        )?;
+        if pc < self.registers[RegId::IS] || pc >= self.registers[RegId::SSP] {
```
xgreenx, line 69: I thought about the solution to update all places where we modify `$pc`. But it also works=)

I don't know how it affects the performance, but we can try to optimize it later=)

xgreenx (approved):

xgreenx:
Hmm, it is a huge breaking change for any bytecode that doesn't return from the execution=)

xgreenx (approved):

fuel-vm/src/interpreter/executors/instruction.rs
Voxelot: all jmp opcodes use the same inner fn, so it would be an easy place to centralize that logic.

Issue #600, opened 2023-10-05
Charge `Create` transaction for `ContractId` calculate and validation


xgreenx:
## Overview

The validation of the `Create` transaction requires a `ContractId` calculation. It involves 2 BMT to calculate the code root hash and state root hash, plus a final hashing to produce `ContractId`.

These operations are not fast, and we should charge users for them.

```rust
        let contract_root = contract.root();
        let state_root = Contract::initial_state_root(storage_slots.iter());
        let contract_id = contract.id(salt, &contract_root, &state_root);
```

## Implementation details

The charging for predicate verification is similar to charging for a contract. So if the https://github.com/FuelLabs/fuel-vm/issues/599 is done, we can reuse the code from there. The only unsolved part is how to charge for the storage slots' BMT root calculation. We can come up with a formula or use the worst case with the `max_storage_slots` constant.

bvrooman:
> It involves 2 BMT to calculate the code root hash and state root hash, plus a final hashing to produce ContractId.

Looking at the code, we have 1 BMT root for the code root, and 1 SMT root for the state root. State root is not a BMT so we will need to add another benchmark for this to calculate the dependent cost.

bvrooman:
Completed by https://github.com/FuelLabs/fuel-vm/pull/618

PR #618, opened 2023-10-28
feat: Charge for `Create` metadata

bvrooman:
Related Issues:
- https://github.com/FuelLabs/fuel-vm/issues/600

This PR extends the gas costs to include costs for contract state root and contract id.

The gas costs are not final and will need to be calculated by running benchmarks on a dedicated instance.


fuel-tx/src/transaction/types/create.rs
```diff
@@ -172,6 +173,32 @@ impl Chargeable for Create {
...
+        let contract_root_gas = contract
+            .map(|contract| {
+                let contract_len = contract.len() as Word;
+                gas_costs.contract_root.resolve(contract_len)
+            })
+            .unwrap_or(0);
+        let state_root_length = storage_slots.len() as Word;
+        let state_root_gas = gas_costs.state_root.resolve(state_root_length);
+
+        let contract_id_gas = gas_costs.contract_id;
```
xgreenx, line 198: `contract_id` calculation is simply the calling of the hash functions. You can use dependent cost for hash function instead

xgreenx (approved):

fuel-tx/src/transaction/types/create.rs
```diff
@@ -172,6 +174,37 @@ impl Chargeable for Create {
...
+        let Create {
+            bytecode_witness_index,
+            witnesses,
+            storage_slots,
+            ..
+        } = self;
+
+        let contract: Option<Contract> = witnesses
+            .get(*bytecode_witness_index as usize)
+            .map(|c| c.as_ref().into());
```
Voxelot, line 188: Should we just get the len here? Based on the type annotation above is seems like we're creating an owned copy of the contract which is going to add a lot more overhead to the gas calculation.

bvrooman: Yep, I can update that in follow up PR.

Issue #570, opened 2023-08-28
TOB-FUEL-36: Consensus between 32bit and 64bit systems can fail


xgreenx:
## Description
Consensus between 32bit and 64bit systems might fail due to a truncating cast. In the following figure, b + c is cast to an usize, which can either be 32bit or 64 bit depending on the architecture. Because of the cast the value of bc can differ between systems, which would hinder the fuel network of forming a consensus if 32bit systems are deployed.

#### Figure 36.1: Cast which leads to different results on 32bit vs 64bit systems. (fuel-vm/fuel-vm/src/interpreter/memory.rs#290–306)

```rust
pub(crate) fn load_byte(
    memory: &[u8; MEM_SIZE],
    pc: RegMut<PC>,
    result: &mut Word,
b: Word,
    c: Word,
) -> Result<(), RuntimeError> {
    let bc = b.saturating_add(c) as usize;
    if bc >= VM_MAX_RAM as RegisterId {
        Err(PanicReason::MemoryOverflow.into())
    } else {
        *result = memory[bc] as Word;
inc_pc(pc) }
}
```

The following unit test panics on a 64bit system but passes on a 32bit system.

#### Figure 36.2: Unit test which succeeds on 32bit fails on 64bit systems.

```rust
let mut memory: Memory<MEM_SIZE> = vec![1u8; MEM_SIZE].try_into().unwrap();
let mut pc = 4;
let mut result = 0;
load_byte(&memory, RegMut::new(&mut pc), &mut result, 0xFFFFFFFF00000000,
0x0).expect("should fail");
```

Other occurrences of this bug can be found in read_bytes. The next figure shows the test two test vectors.

#### Figure 36.3: Unit test which shows two test vectors to triggere divergent behavior depending on the system architecture.

```rust
// Test vector 1: for read_bytes
load_word(&memory, RegMut::new(&mut pc), &mut result, 0xFFFFFFFF00000000,
0x0).expect("should fail");
// Test vector 2: for get_transaction_field
let input = GTFInput {
    tx: &Script::default(),
    tx_offset: 0,
    pc: RegMut::new(&mut pc),
};
let mut result = 1;
input.get_transaction_field(&mut result, 0xFFFFFFFF00000000, 0x201).expect("should
fail");
```

Other casts like in memeq are protected by an if-clause which voids such problems because VM_MAX_RAM is lower than 232-1.

## Exploit ~Attack~ Scenario
[...]

PR #619, opened 2023-10-30
Avoid possible truncation of higher bits

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/570

### DIFF
--- a/fuel-merkle/src/common/msb.rs
+++ b/fuel-merkle/src/common/msb.rs
@@ -53,6 +53,7 @@ impl<const N: usize> Msb for [u8; N] {
     }
 }
 
+#[allow(clippy::cast_possible_truncation)]
 #[cfg(test)]
 mod test {
     use crate::common::{
@@ -62,10 +63,11 @@ mod test {
         Bytes8,
         Msb,
     };
+    use core::mem::size_of;
 
     #[test]
     fn test_msb_for_bytes_1() {
-        const NUM_BITS: u32 = 8;
+        const NUM_BITS: u32 = size_of::<Bytes1>() as u32 * 8;
 
         let bytes: Bytes1 = [0b10101010];
         let expected_n = u8::from_be_bytes(bytes);
@@ -82,7 +84,7 @@ mod test {
 
     #[test]
     fn test_msb_for_bytes_2() {
-        const NUM_BITS: u32 = 2 * 8;
+        const NUM_BITS: u32 = size_of::<Bytes2>() as u32 * 8;
 
         let bytes: Bytes2 = [0b10101010, 0b10101010];
         let expected_n = u16::from_be_bytes(bytes);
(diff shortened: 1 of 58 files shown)


fuel-tx/src/transaction/types/output.rs
```diff
@@ -181,7 +181,11 @@ impl Output {
     }
 
     pub fn message_nonce(txid: &Bytes32, idx: Word) -> Nonce {
-        (*Hasher::default().chain(txid).chain([idx as u8]).finalize()).into()
+        (*Hasher::default()
+            .chain(txid)
+            .chain(idx.to_bytes())
```
xgreenx, line 186: If we are okay with supporting more than `255` receipts, then we also need to update the [specification](https://github.com/FuelLabs/fuel-specs/blob/master/src/identifiers/utxo-id.md#message-nonce)

fuel-merkle/src/common/msb.rs
```diff
@@ -62,11 +62,10 @@ mod test {
         Bytes8,
         Msb,
     };
-    use core::mem::size_of;
 
     #[test]
     fn test_msb_for_bytes_1() {
-        const NUM_BITS: usize = size_of::<Bytes1>() * 8;
```
Voxelot, line 69: I like the `size_of` approach as it's more obvious to the code reader how this value is constructed. Maybe we should split  this into two constants like `NUM_BYTES` vs `NUM_BITS`?

fuel-tx/src/transaction/types/output.rs
Dentosal: I think this was settled yesterday and we do want to support U16::MAX receipts. Would a PR changing the linked paragraph be enough? I recall there was some discussion about serialization as well.

xgreenx: We need to update the paragraph and mention that we need to use canonical serialization for integers. 

We have a separate issue to add a `u16::max` limit https://github.com/FuelLabs/fuel-vm/issues/620

Dentosal (approved):

fuel-merkle/src/common/msb.rs
```diff
@@ -5,11 +5,11 @@ pub enum Bit {
 }
 
 trait GetBit {
-    fn get_bit(&self, bit_index: usize) -> Option<Bit>;
+    fn get_bit(&self, bit_index: u32) -> Option<Bit>;
```
bvrooman, line 8: Are these changes to `fuel-merkle` necessary? It feels like we're changing library logic to fit business logic. A library should promise correctness under all targets it supports, including all 32 bit and 64 bit targets. If it has internal problems with truncation, that should be fixed in the library. But this change appears to be changing the library interface to accommodate changes to external libraries.

xgreenx: It is a private trait, so it is an internal detail of how Merkle tree works=)

Issue #576, opened 2023-08-29
Support `Transaction` Policies


xgreenx:
## Overview

Related issue https://github.com/FuelLabs/fuel-core/issues/1404 that should define `gas_per_byte`.

Related specification change https://github.com/FuelLabs/fuel-specs/pull/514

Add policies to transactions with all limits managed by policies (in preparation for multi-dimensional pricing in the future):
- Move the gas limit to be a policy
- Add witness data limit as a policy

Min fee also should include `gas_per_byte * WL(Witness limit)`. And WL should affect the validity rules of the transaction.
The gas limit should exclude the predicate gas used, and the predicate gas used should be part of the min fee.

xgreenx:
This change is better to do on top of https://github.com/FuelLabs/fuel-vm/pull/533

PR #623, opened 2023-11-04
Support transaction policies

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/576 Closes https://github.com/FuelLabs/fuel-core/issues/1398

Added support for transaction policies. The `Script` and `Create` 
    transactions received a new field, `policies`. Policies allow the addition 
    of some limits to the transaction to protect the user or specify some details regarding execution.

This change makes the `GasPrice` and `Maturity` fields optional, allowing to save space in the future. Also, this will enable us to support multidimensional prices later.
    
Along with this change, we introduced two new policies:
- `WitnessLimit` - allows the limitation of the maximum size of witnesses in bytes for the contract. Because of the changes in the gas calculation model(the blockchain also charges the user for the witness data), the user should protect himself from the block producer or third parties blowing up witness data and draining the user's funds.
- `MaxFee` - allows the upper bound for the maximum fee that users agree to pay for the transaction.

This change brings the following modification to the gas model:
- The `GasLimit` only limits the script execution. Previously, the `GasLimit` also limited the predicate execution time, but it is not valid anymore. So, it is not possible to use the `GasLimit` for transaction cost limitations. A new `MaxFee` policy is a way to do that. The `GasLimit` field was removed from the `Create` transaction because it only relates to the script execution(which the `Create` transaction doesn't have).
- The blockchain changes the user for the size of witness data(before it was free). There is no separate price for the storage, so it uses gas to charge the user. It affects the `min_gas` and `min_fee` calculation.
- A new `WhitnessLimit` also impacts the `max_gas` and `max_fee` calculation along with the `GasLimit`(in the case of `Create` transaction only `WitnessLimit` affects the `max_gas` and `max_fee`).
- The minimal gas also charges the user for transaction ID calculation.
      
The change has the following modification to the layout transaction:
- The `Create` transaction doesn't have the `GasLimit` field anymore. Because the `Create` transaction doesn't have any script to execute
- The `Create` and `Script` transactions don't have explicit `maturity` and `gas_price` fields. Instead, these fields can be set via a new `policies` field.
- The `Create` and `Script` transactions have a new `policies` field with a unique canonical serialization and deserialization for optimal space consumption.
    
Other breaking changes caused by the change:
- Each transaction requires setting the `GasPrice` policy.
- Previously, `GasLimit` should be less than the `MAX_GAS_PER_TX` constant. After removing this field from the `Create` transaction, it is impossible to require it. Instead, it requires that `max_gas <= MAX_GAS_PER_TX` for any transaction. Consequently, any `Script` transaction that uses `MAX_GAS_PER_TX` as a `GasLimit` will always fail because of a new rule. Setting the estimated gas usage instead solves the problem.
- If the `max_fee > policies.max_fee`, then transaction will be rejected.
- If the `witnessses_size > policies.witness_limit`, then transaction will be rejected.
- GTF opcode changed its hardcoded constants for fields. It should be updated according to the values from the specification on the Sway side.


fuel-asm/src/args.rs
```diff
@@ -50,218 +50,221 @@ crate::enum_try_from! {
...
-
-        /// Set `$rA` to `tx.gasPrice`
-        CreateGasPrice = 0x010,
-
-        /// Set `$rA` to `tx.gasLimit`
-        CreateGasLimit = 0x011,
-
-        /// Set `$rA` to `tx.maturity`
-        CreateMaturity = 0x012,
+        ScriptWitnessAtIndex = 0x00D,
```
MitchTurner, line 87: Why does it skip from `0x00D` to `0x100`?

MitchTurner: It looks like jumps on a few of them.

fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -136,6 +116,104 @@ fn maturity() {
     assert_eq!(CheckError::TransactionMaturity, err);
 }
 
+#[test]
+fn witness_limit() {
```
MitchTurner, line 120: Could we give this name more context? Like, its testing an expected failure mode, it would help of the test name communicated why this is failing.
e.g.
`check__fails_when_too_many_witnesses` or whatever. 

"Given, when, then" would also help. That might mean these should be separate tests.

fuel-tx/src/transaction.rs
```diff
@@ -551,20 +547,48 @@ pub mod field {
...
     pub trait GasPrice {
-        fn gas_price(&self) -> &Word;
-        fn gas_price_mut(&mut self) -> &mut Word;
-        fn gas_price_offset(&self) -> usize {
-            Self::gas_price_offset_static()
+        fn gas_price(&self) -> Word;
+        fn set_gas_price(&mut self, value: Word);
+    }
+
+    impl<T: Policies + ?Sized> GasPrice for T {
```
MitchTurner, line 562: Curious why you include `?Sized` in all the blanket impls.

fuel-asm/src/args.rs
xgreenx: The values are from the [specification](https://github.com/FuelLabs/fuel-specs/pull/514/files#diff-ee3e0bbd57928f651eca1142f65942e1592743ec5fd20e20b8a44c9998679b06)

fuel-tx/src/transaction.rs
xgreenx: By default in Rust, the implementation requires `Sized`, while the trait definition - no. Because implementation of the `Policies` trait is enough to have the other four traits implementations (`GasPrice`, `WitnessLimit`, `Maturity` `MaxFeeLimit`), I don't want to require `Sized` on the `Policies` or in any other bounds which try to utilize this blanket implementation.

You can try to remove `?Sized` and you will see that I'm talking about=)

fuel-tx/src/transaction/policies.rs
```diff
@@ -0,0 +1,324 @@
...
+    #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+    pub struct PoliciesBits: u32 {
+        /// If set, the gas price is present in the policies.
+        const GasPrice = 1;
+        /// If set, the witness limit is present in the policies.
+        const WitnessLimit = 2;
+        /// If set, the maturity is present in the policies.
+        const Maturity = 4;
+        /// If set, the max fee is present in the policies.
+        const MaxFee = 8;
```
Voxelot, line 34: Might look better to use bitshifts or binary representation

```suggestion
        const WitnessLimit = 1<<2;
        /// If set, the maturity is present in the policies.
        const Maturity = 1<<3;
        /// If set, the max fee is present in the policies.
        const MaxFee = 1<<4;
```

fuel-tx/src/transaction.rs
MitchTurner: Hmm. Okay. At a high level I understand what it's saying.

> You can try to remove ?Sized and you will see that I'm talking about=)

It still compiles without it. So I guess I was wondering if it's a "best practice" sort of thing or if it is required by something depending on the trait.

I'm not worried about it, just trying to learn.

fuel-tx/src/transaction/validity.rs
```diff
@@ -317,23 +312,49 @@ where
...
+        let witness_size = tx.witnesses().size_dynamic();
+        if witness_size as u64 > witness_limit {
+            Err(CheckError::TransactionWitnessLimitExceeded)?
+        }
+    }
+
+    let max_gas = tx.max_gas(gas_costs, fee_params);
+    if max_gas > tx_params.max_gas_per_tx {
+        Err(CheckError::TransactionMaxGasExceeded)?
+    }
```
xgreenx, line 349: The specification says `GasLimit > MAX_GAS_PER_TX`, while in real life should limit the total transaction cost.

I've created a [specification PR](https://github.com/FuelLabs/fuel-specs/pull/533) to actualize it.

But it is a huge breaking change because SDK right now uses `MAX_GAS_PER_TX ` as a `GasLimit` -> the transaction always fails because of this check.

fuel-vm/src/interpreter/internal/message_tests.rs
```diff
@@ -13,12 +13,12 @@ use fuel_tx::Create;
...
-#[test_case(0, 1, 2 => Some(184))]
-#[test_case(0, 2, 3 => Some(256))]
-#[test_case(0, 1, 3 => Some(184))]
-#[test_case(44, 2, 3 => Some(300))]
+#[test_case(0, 0, 1 => Some(96))]
+#[test_case(88, 0, 1 => Some(184))]
+#[test_case(0, 1, 2 => Some(168))]
+#[test_case(0, 2, 3 => Some(240))]
+#[test_case(0, 1, 3 => Some(168))]
+#[test_case(44, 2, 3 => Some(284))]
```
xgreenx, line 21: Because we removed two fields, we need to subtract `16`.

fuel-tx/src/transaction/fee.rs
```diff
@@ -142,57 +92,122 @@ impl TransactionFee {
...
 
-    /// Returns the gas limit.
-    fn limit(&self) -> Word;
+    /// Returns the maximum possible gas after the end of transaction execution.
+    ///
+    /// The function guarantees that the value is not less than [Self::min_gas].
+    fn max_gas(&self, gas_costs: &GasCosts, fee: &FeeParameters) -> Word {
+        let remaining_allowed_witness = self
+            .witness_limit()
+            .saturating_sub(self.witnesses().size_dynamic() as u64)
```
MitchTurner, line 142: I was thinking `witness_limit()` returned the _count_ of witnesses. You are subtracting the _size_ of all the witnesses, so maybe I was mistaken. Just want to make sure you aren't subtracting size from count.

xgreenx: `WitnessLimit` limits the total size of witnesses in bytes=) Because each witness can be arbitrary in size, limiting by the count will not help

image

fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -136,6 +116,142 @@ fn maturity() {
...
+    let bytecode = vec![];
+    TransactionBuilder::create(bytecode.clone().into(), rng.gen(), vec![])
+        .add_random_fee_input()
+        .finalize()
+        .check(block_height, &test_params())
+        .expect("Failed to validate tx create");
+}
+
+#[test]
+fn witness_limit_script_limit_works() {
```
MitchTurner, line 146: This should probably be 2 tests. Then you'd have something like:

```
// given
let witness_size = Signature::LEN + vec![0u8; 0].size_static();

// when
let limit = witness_size;
let res = TransactionBuilder::script(generate_bytes(rng), generate_bytes(rng))
        .add_random_fee_input()
        .witness_limit(limit as u64)
        .finalize()
        .check(block_height, &test_params());
        
// then
assert!(res.is_ok());
```
And the other can just do the same but with:
```
let limit = witness_size - 1;
```
with the same `assert_eq!`

The same goes for all the following tests.

fuel-tx/src/transaction/policies.rs
```diff
@@ -0,0 +1,329 @@
...
+            let maturity: u32 = rng.gen();
+            policies.set(PolicyType::Maturity, Some(maturity as u64));
+        }
+
+        policies
+    }
+}
+
+#[test]
+fn values_for_bitmask_works() {
```
MitchTurner, line 274: I think these could both have better names. `works` implies that the reader knows what the desired behavior is, but one of the points of the tests is to communicate the desired behaviour:
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests

Both could benefit from some given-when-then love as well:
https://martinfowler.com/bliki/GivenWhenThen.html

fuel-tx/src/transaction/policies.rs
```diff
@@ -0,0 +1,329 @@
...
+    for bitmask in 0..MAX_BITMASK {
+        let bits =
+            PoliciesBits::from_bits(bitmask).expect("Should construct a valid bits");
+        set.insert(Policies::values_for_bitmask(bits, VALUES));
+    }
+    assert_eq!(set.len(), MAX_BITMASK as usize);
+}
+
+#[test]
+fn canonical() {
```
MitchTurner, line 289: I think it's generally suggested to only have one `assert` per test. I'm kinda flexible on that when the entire test is the same except for the `then` section. That's assuming that it's communicated well why you're checking multiple things.

fuel-tx/src/transaction/types/create.rs
```diff
@@ -183,16 +168,12 @@ impl Chargeable for Create {
             ..
         } = self;
 
-        let contract: Option<Contract> = witnesses
+        let contract_len = witnesses
             .get(*bytecode_witness_index as usize)
-            .map(|c| c.as_ref().into());
+            .map(|c| c.as_ref().len())
+            .unwrap_or_default();
```
MitchTurner, line 174:
```suggestion
            .unwrap_or(0);
```
Helps with readability.

fuel-tx/src/transaction/consensus_parameters/gas.rs
```diff
@@ -582,11 +582,11 @@ impl GasCostsValues {
 }
 
 impl DependentCost {
-    /// Create costs that are all set to zero.
+    /// Create costs that make operations free.
     pub fn free() -> Self {
         Self {
             base: 0,
-            dep_per_unit: 0,
+            dep_per_unit: u64::MAX,
```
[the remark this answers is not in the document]
Dentosal: Adding a comment explaining that would be nice. Although, I'm also considering using `0` to make them free anyway in https://github.com/FuelLabs/fuel-vm/pull/625

fuel-tx/src/transaction/policies.rs
```diff
@@ -0,0 +1,329 @@
...
+        Standard,
+    },
+    Rng,
+};
+
+bitflags::bitflags! {
+    /// See https://github.com/FuelLabs/fuel-specs/blob/master/src/tx-format/policy.md#policy
+    #[derive(Clone, Copy, Default, Debug, PartialEq, Eq, Hash)]
+    #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+    pub struct PoliciesBits: u32 {
```
Dentosal, line 26: Is there a particular reason to use `u32`?

fuel-tx/src/transaction/types/script.rs
```diff
@@ -81,11 +81,13 @@ impl Default for Script {
         let script = fuel_asm::op::ret(0x10).to_bytes().to_vec();
 
         Self {
-            gas_price: Default::default(),
-            gas_limit: TxParameters::DEFAULT.max_gas_per_tx,
-            maturity: Default::default(),
+            gas_limit: TxParameters::DEFAULT.max_gas_per_tx >> 1,
```
Dentosal, line 84: Where does this value come from? Also using `/ 2` instead of a bit shift would be cleaner anyway.

fuel-vm/src/interpreter/memory/tests.rs
```diff
@@ -26,7 +26,7 @@ fn memcopy() {
         InterpreterParams::from(&consensus_params),
     );
     let tx = TransactionBuilder::script(op::ret(0x10).to_bytes().to_vec(), vec![])
-        .gas_limit(tx_params.max_gas_per_tx)
+        .gas_limit(tx_params.max_gas_per_tx >> 1)
```
Dentosal, line 29: Needs explanation.

fuel-vm/src/interpreter/metadata.rs
```diff
@@ -154,16 +156,39 @@ impl<Tx> GTFInput<'_, Tx> {
...
+            GTFArgs::PolicyCount => tx.policies().len() as Word,
+            GTFArgs::PolicyType => {
+                tx.policies()
+                    .get_type_by_index(b)
+                    .ok_or(PanicReason::PolicyNotFound)? as Word
+            }
+            GTFArgs::PolicyGasPrice => tx
+                .policies()
+                .get(PolicyType::GasPrice)
+                .ok_or(PanicReason::PolicyIsNotSet)?,
```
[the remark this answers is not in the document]
Dentosal: Just throw an error. Returning an invalid value is not backward compability, it's just wrong.

fuel-vm/src/tests/predicate.rs
```diff
@@ -142,7 +140,11 @@ where
             assert_eq!(p_gas_used, s_gas_used);
             true
         }
-        (Err(_), Err(_)) => false,
+        (Err(p_err), Err(s_err)) => {
+            assert_eq!(p_err, s_err);
+            println!("{:?}", p_err);
```
Dentosal, line 145: Do we need this print here?

fuel-vm/src/interpreter/memory/tests.rs
MitchTurner, line 29: What's going on here? Isn't this just halving the `max_gas_per_tx`?

fuel-vm/src/util.rs
```diff
@@ -592,7 +591,7 @@ pub mod test_helpers {
     ) {
         let gas_price = 0;
         let tx_params = TxParameters::default().with_max_gas_per_tx(Word::MAX / 2);
-        let gas_limit = tx_params.max_gas_per_tx;
+        let gas_limit = tx_params.max_gas_per_tx >> 1;
```
MitchTurner, line 594: Here too.

fuel-tx/src/builder.rs
```diff
@@ -129,11 +116,10 @@ pub struct TransactionBuilder<Tx> {
 impl TransactionBuilder<Script> {
     pub fn script(script: Vec<u8>, script_data: Vec<u8>) -> Self {
         let tx = Script {
-            gas_price: Default::default(),
             gas_limit: Default::default(),
-            maturity: Default::default(),
             script,
             script_data,
+            policies: Policies::new().with_gas_price(0),
```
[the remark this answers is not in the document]
xgreenx: Because of this check:
image

We use builders mostly for tests and it is better to set this policy for all tests instead of updating all places.

fuel-tx/src/transaction/policies.rs
xgreenx: https://github.com/FuelLabs/fuel-specs/pull/514#discussion_r1366359260

CHANGELOG.md
```diff
@@ -6,6 +6,38 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+    - `WitnessLimit` - allows the limitation of the maximum size of witnesses in bytes for the contract. Because of the changes in the gas calculation model(the blockchain also charges the user for the witness data), the user should protect himself from the block producer or third parties blowing up witness data and draining the user's funds.
+    - `MaxFee` - allows the upper bound for the maximum fee that users agree to pay for the transaction.
+
+    This change brings the following modification to the gas model:
+    - The `GasLimit` only limits the script execution. Previously, the `GasLimit` also limited the predicate execution time, but it is not valid anymore. So, it is not possible to use the `GasLimit` for transaction cost limitations. A new `MaxFee` policy is a way to do that. The `GasLimit` field was removed from the `Create` transaction because it only relates to the script execution(which the `Create` transaction doesn't have).
+    - The blockchain changes the user for the size of witness data(before it was free). There is no separate price for the storage, so it uses gas to charge the user. It affects the `min_gas` and `min_fee` calculation.
```
Voxelot, line 25:
```suggestion
    - The blockchain charges the user for the size of witness data(before it was free). There is no separate price for the storage, so it uses gas to charge the user. It affects the `min_gas` and `min_fee` calculation.
```

CHANGELOG.md
```diff
@@ -6,6 +6,38 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+    - A new `WhitnessLimit` also impacts the `max_gas` and `max_fee` calculation along with the `GasLimit`(in the case of `Create` transaction only `WitnessLimit` affects the `max_gas` and `max_fee`).
```
Voxelot, line 26:
```suggestion
    - A new policy called `WitnessLimit` also impacts the `max_gas` and `max_fee` calculation in addition to `GasLimit`(in the case of `Create` transaction only `WitnessLimit` affects the `max_gas` and `max_fee`).
```

CHANGELOG.md
```diff
@@ -6,6 +6,38 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+    - The minimal gas also charges the user for transaction ID calculation.
+      
+    The change has the following modification to the layout transaction:
```
Voxelot, line 29:
```suggestion
    The change has the following modification to the transaction layout:
```

Dentosal (dismissed):
LGTM; I suggest you wait for another review as well.

CHANGELOG.md
```diff
@@ -6,6 +6,40 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+
+#### Breaking
+
+- [#623](https://github.com/FuelLabs/fuel-vm/pull/623): 
+    Added support for transaction policies. The `Script` and `Create` 
+    transactions received a new field, `policies`. Policies allow the addition 
+    of some limits to the transaction to protect the user or specify some details regarding execution.
+    This change makes the `GasPrice` and `Maturity` fields optional, allowing to save space in the future. 
+    Also, this will enable us to support multidimensional prices later.
+    Renamed the `GasLimit` was renamed to `ScriptGasLimit`.
```
Voxelot, line 19:
```suggestion
    `GasLimit` was renamed to `ScriptGasLimit`.
```

CHANGELOG.md
```diff
@@ -6,6 +6,40 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+    
+    Along with this change, we introduced two new policies:
+    - `WitnessLimit` - allows the limitation of the maximum size of witnesses in bytes for the contract. Because of the changes in the gas calculation model(the blockchain also charges the user for the witness data), the user should protect himself from the block producer or third parties blowing up witness data and draining the user's funds.
+    - `MaxFee` - allows the upper bound for the maximum fee that users agree to pay for the transaction.
+
+    This change brings the following modification to the gas model:
+    - The `ScriptGasLimit` only limits the script execution. Previously, the `ScriptGasLimit` also limited the predicate execution time, but it is not valid anymore. So, it is not possible to use the `ScriptGasLimit` for transaction cost limitations. A new `MaxFee` policy is a way to do that. The `ScriptGasLimit` field was removed from the `Create` transaction because it only relates to the script execution(which the `Create` transaction doesn't have).
```
Voxelot, line 26:
```suggestion
    - The `ScriptGasLimit` only limits script execution. Previously, the `ScriptGasLimit` also limited the predicate execution time, instead predicate gas is now directly included into `min_fee`. So, it is not possible to use the `ScriptGasLimit` for transaction cost limitations. A new `MaxFee` policy is a way to do that. The `GasLimit` field was removed from the `Create` transaction because it only relates to the script execution (which the `Create` transaction doesn't have).
```

fuel-asm/src/args.rs
```diff
@@ -50,218 +50,218 @@ crate::enum_try_from! {
...
 
         /// Set `$rA` to `tx.witnesses[$rB].dataLength`
-        WitnessDataLength = 0x301,
+        WitnessDataLength = 0x400,
 
         /// Set `$rA` to `Memory address of tx.witnesses[$rB].data`
-        WitnessData = 0x302,
+        WitnessData = 0x401,
+
+        /// Set `$rA` to `tx.policies[$rB].policyTypes`
```
Voxelot, line 251:
```suggestion
        /// Set `$rA` to `tx.policyTypes`
```

fuel-asm/src/args.rs
```diff
@@ -50,218 +50,218 @@ crate::enum_try_from! {
...
+        PolicyTypes = 0x501,
```
Voxelot, line 252: We should bump these up since we aren't using `GTF_POLICY_COUNT` anymore

```suggestion
        PolicyTypes = 0x500,
```

fuel-tx/src/builder.rs
```diff
@@ -313,8 +297,11 @@ impl<Tx: Buildable> TransactionBuilder<Tx> {
         self
     }
 
-    pub fn gas_limit(&mut self, gas_limit: Word) -> &mut Self {
-        self.tx.set_gas_limit(gas_limit);
+    pub fn gas_limit(&mut self, gas_limit: Word) -> &mut Self
```
Voxelot, line 300:
```suggestion
    pub fn script_gas_limit(&mut self, gas_limit: Word) -> &mut Self
```

fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -136,6 +116,195 @@ fn maturity() {
...
+        .witness_limit(limit as u64)
+        .finalize()
+        .check(block_height, &test_params());
+
+    // Then
+    assert!(result.is_ok());
+}
+
+#[test]
+fn script_set_witness_limit_for_non_empty_witness_fails() {
```
Voxelot, line 172:
```suggestion
fn script_set_witness_limit_less_than_witness_data_size_fails() {
```

fuel-tx/src/transaction.rs
```diff
@@ -17,6 +39,7 @@ use itertools::Itertools;
 
 mod fee;
 mod metadata;
+pub mod policies;
```
Voxelot, line 42: Should pub mods be sorted into a separate section from private mods?

fuel-vm/src/tests/encoding.rs
```diff
@@ -212,115 +213,184 @@ fn transaction() {
 
     assert_encoding_correct(&[
         Transaction::script(
-            Word::MAX >> 1,
             Word::MAX >> 2,
-            (u32::MAX >> 3).into(),
             vec![0xfa],
             vec![0xfb, 0xfc],
+            Policies::new()
+                .with_gas_price(Word::MAX >> 1)
```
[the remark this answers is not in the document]
MitchTurner: Gonna bump this. 

I don't think we should be using `>>` and I think it would be best to move the values to variables.

fuel-tx/src/transaction/policies.rs
```diff
@@ -0,0 +1,333 @@
...
+    MaxFee,
+}
+
+impl PolicyType {
+    pub const fn index(&self) -> usize {
+        match self {
+            PolicyType::GasPrice => 0,
+            PolicyType::WitnessLimit => 1,
+            PolicyType::Maturity => 2,
+            PolicyType::MaxFee => 3,
```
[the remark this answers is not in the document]
xgreenx: The idea was to write the implementation manually to control values because, in the future, we may remove `GasPrice` or any other policy.

MitchTurner (dismissed):

fuel-tx/src/transaction/policies.rs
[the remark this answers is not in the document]
bvrooman: As long as they don't get out of sync 👍

bvrooman:
I proposed some changes in https://github.com/FuelLabs/fuel-vm/pull/630 - mostly for the purpose of allowing us to encapsulate policy definitions and checks better. It would be nice to have this logic centralized rather than spread out, because it means a reduction in overall complexity: Having one place to define this logic would mean less duplication and easier extensibility. But I am simply suggesting one approach - if you have an idea on how you would like to do this, we can review that too now or at a later date after this PR is in.

bvrooman (approved):

PR #627, opened 2023-11-08
Remove SMT nodes along the path during `update`

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/480


MitchTurner (dismissed):

CHANGELOG.md
```diff
@@ -40,6 +40,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     - If the `witnessses_size > policies.witness_limit`, then transaction will be rejected.
     - GTF opcode changed its hardcoded constants for fields. It should be updated according to the values from the specification on the Sway side.
 
+### Fixed
+
+- [#627](https://github.com/FuelLabs/fuel-vm/pull/627): Added removal of SMT nodes along the path during `update` operation.
```
bvrooman, line 45:
```suggestion
- [#627](https://github.com/FuelLabs/fuel-vm/pull/627): Added removal of obsolete SMT nodes along the path during `update` and `delete` operations.
```

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -924,6 +934,56 @@ mod test {
...
+        // Given
+        tree.update(tenth_key, b"DATA").unwrap();
+        let size_after_tenth = tree.storage().len();
+        assert_ne!(size_after_tenth, size_before_tenth);
+
+        // When
+        tree.update(tenth_key, b"ANOTHER_DATA").unwrap();
+
+        // Then
+        assert_eq!(tree.storage().len(), size_after_tenth);
```
bvrooman, line 959: I don't believe that it's guaranteed that the number of entries in the DB will necessarily be the same. Depending on the case, you will end up with more or less placeholder nodes in your path (which will not be in the database) and more or less real nodes. That means that the number of entries is not constant.
Because of the sparseness of the SMT, you're statistically likely to get a path of length 1 or 2 at a low quantity of leaves, and I think that is what is happening here. I will investigate.

bvrooman: Discussed with @xgreenx offline re: number of expected nodes during update. We concluded that we can calculate deterministically the number of nodes before and after a consecutive update using the same key. In this way, we can test for a constant value.

bvrooman (approved):

Dentosal (approved):

PR #629, opened 2023-11-08
Charge the user for VM initialization

xgreenx:
Related to https://github.com/FuelLabs/fuel-core/issues/1442


Dentosal:
This looks suspcious. How does this not break more tests?

fuel-vm/src/tests/predicate.rs
```diff
@@ -204,7 +204,7 @@ async fn get_verifying_predicate() {
 async fn execute_gas_metered_predicates(
     predicates: Vec<Vec<Instruction>>,
 ) -> Result<u64, ()> {
-    const GAS_LIMIT: Word = 10000;
+    const GAS_LIMIT: Word = 20000;
```
MitchTurner, line 207: Were some tests running out of gas or something?

xgreenx: Yes, because now `min_gas` is higher than before, it fails out of gas. But I decided to revert this change and fix the test configuration instead.

xgreenx:
> This looks suspcious. How does this not break more tests?

Because mostly in all tests we are using default `max_gas_per_tx`=)

MitchTurner (dismissed):

Dentosal (approved):

bvrooman (commented):
Are we certain that it is only 1 VM initialization per transaction? I remember discussing that there could be many VM initializations for things like predicates.

xgreenx:
We charge the user for each predicate too=) 
image

Voxelot (approved):

Issue #602, opened 2023-10-05
Increase the cost of storage opcodes by the gas per byte fee in addition to the execution cost


xgreenx:
## Problem overview

Some opcodes store data in storage, allowing the user to blow up the footprint.

## Solution

The solution is to charge the user for the storage in addition to the execution. Because multi-dimensional resources require much adoption work, we will charge the user in terms of additional gas for now.

While we work on https://github.com/FuelLabs/fuel-core/issues/1404 to find what value we need to charge, we also need to update opcodes and benchmarks to use `gas_per_byte`.

## Implementation details

We need to find all opcodes that modify the storage first. For that, we can find all places where we use methods from the `InterpreterStorage` trait.
It's okay to charge the user for `deploy_contract_with_id`, because we store the contract's bytecode in two tables(transactions and contracts).
We need to charge the user only for adding new information. If the user overrides the already existing value, we only need to charge for newly added bytes(so if the size of the final value is the same or less, we don't charge).

### Note
Inserting a new K/V into a state or asset into balances also has Sparse Merkle Tree storage overhead. You don't need to worry about that right now. Maybe later, we will have a separate functionality that charges for this overhead.

As a temporary solution, we may try to include this into `gas_per_byte`.

PR #634, opened 2023-11-10
Charge for new storage bytes

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/602

This is still a rough draft.

Breaking changes:
* Write opcodes now return the number of new storage slots instead of just a boolean on whether the value existed. [The return value is currently ignored by the Sway stdlib](https://github.com/FuelLabs/sway/blob/28d3500e702337fde0d7e6975d6e0a40aa49ad1d/sway-lib-std/src/storage/storage_api.sw#L56C13-L56C31), so it shouldn't be too bad

### DIFF
--- a/fuel-tx/src/transaction/consensus_parameters/gas.rs
+++ b/fuel-tx/src/transaction/consensus_parameters/gas.rs
@@ -321,6 +321,7 @@ pub struct GasCostsValues {
     // Non-opcode costs
     pub contract_root: DependentCost,
     pub state_root: DependentCost,
+    pub new_storage_per_byte: Word, // TODO: this is currently per slot
 }
 
 /// Dependent cost is a cost that depends on the number of units.
@@ -474,6 +475,7 @@ impl GasCostsValues {
             swwq: DependentCost::free(),
             contract_root: DependentCost::free(),
             state_root: DependentCost::free(),
+            new_storage_per_byte: 0,
         }
     }
 
@@ -588,6 +590,7 @@ impl GasCostsValues {
             swwq: DependentCost::unit(),
             contract_root: DependentCost::unit(),
             state_root: DependentCost::unit(),
+            new_storage_per_byte: 1,
         }
     }
 }
--- a/fuel-vm/src/interpreter/blockchain.rs
+++ b/fuel-vm/src/interpreter/blockchain.rs
@@ -6,6 +6,7 @@ use super::{
     },
     gas::{
         dependent_gas_charge,
+        gas_charge,
         ProfileGas,
     },
     internal::{
@@ -318,8 +319,18 @@ where
         rb: RegisterId,
         c: Word,
     ) -> IoResult<(), S::DataError> {
-        let (SystemRegisters { fp, pc, .. }, mut w) =
-            split_registers(&mut self.registers);
+        let new_storage_gas_per_byte = self.gas_costs().new_storage_per_byte;
+        let (
+            SystemRegisters {
+                cgas,
+                ggas,
+                is,
+                fp,
+                pc,
+                ..
+            },
+            mut w,
+        ) = split_registers(&mut self.registers);
         let (result, got_result) = w
             .get_mut_two(WriteRegKey::try_from(ra)?, WriteRegKey::try_from(rb)?)
             .ok_or(RuntimeError::Recoverable(
@@ -336,6 +347,12 @@ where
                 storage,
                 memory,
                 context,
+                profiler: &mut self.profiler,
+                new_storage_gas_per_byte,
+                current_contract: self.frames.last().map(|frame| frame.to()).copied(),
+                cgas,
+                ggas,
+                is: is.as_ref(),
                 fp: fp.as_ref(),
                 pc,
             },
@@ -373,8 +390,18 @@ where
         rb: RegisterId,
         c: Word,
     ) -> IoResult<(), S::DataError> {
-        let (SystemRegisters { fp, pc, .. }, mut w) =
-            split_registers(&mut self.registers);
+        let new_storage_gas_per_byte = self.gas_costs().new_storage_per_byte;
+        let (
+            SystemRegisters {
+                cgas,
+                ggas,
+                is,
+                fp,
+                pc,
+                ..
+            },
+            mut w,
+        ) = split_registers(&mut self.registers);
         let exists = &mut w[WriteRegKey::try_from(rb)?];
         let Self {
             ref mut storage,
@@ -387,6 +414,12 @@ where
                 storage,
                 memory,
                 context,
+                profiler: &mut self.profiler,
+                new_storage_gas_per_byte,
+                current_contract: self.frames.last().map(|frame| frame.to()).copied(),
+                cgas,
+                ggas,
+                is: is.as_ref(),
                 fp: fp.as_ref(),
                 pc,
             },
@@ -403,8 +436,14 @@ where
         c: Word,
         d: Word,
     ) -> IoResult<(), S::DataError> {
+        let new_storage_per_byte = self.gas_costs().new_storage_per_byte;
         let contract_id = self.internal_contract().copied();
-        let (SystemRegisters { pc, .. }, mut w) = split_registers(&mut self.registers);
+        let (
+            SystemRegisters {
+                is, cgas, ggas, pc, ..
+            },
+            mut w,
+        ) = split_registers(&mut self.registers);
         let result = &mut w[WriteRegKey::try_from(rb)?];
 
         let input = StateWriteQWord::new(a, c, d)?;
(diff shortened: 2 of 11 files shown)


fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -846,6 +898,12 @@ pub(crate) struct StateWordCtx<'vm, S> {
     pub storage: &'vm mut S,
     pub memory: &'vm [u8; MEM_SIZE],
     pub context: &'vm Context,
+    pub profiler: &'vm mut Profiler,
+    pub new_storage_gas_per_byte: Word,
+    pub current_contract: Option<ContractId>,
+    pub cgas: RegMut<'vm, CGAS>,
+    pub ggas: RegMut<'vm, GGAS>,
+    pub is: Reg<'vm, IS>,
```
Dentosal, line 906: These are not needed for state read opcodes, so maybe it makes sense to split this struct into two? On the other hand, we're handing over over half of the VM anyway, so just passing the VM directly might make sense.

fuel-tx/src/transaction/consensus_parameters/gas.rs
```diff
@@ -321,6 +321,7 @@ pub struct GasCostsValues {
     // Non-opcode costs
     pub contract_root: DependentCost,
     pub state_root: DependentCost,
+    pub new_storage_per_byte: Word,
```
xgreenx, line 324: I think it is better to use `DependentCost` here because we need to charge for a new storage slot, plus, in the future(when we support dynamic storage values), charge per byte.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -913,7 +978,23 @@ pub(crate) fn state_write_word<S: InterpreterStorage>(
...
+            pc: pc.as_ref(),
+            is,
+            current_contract,
+            profiler,
+        };
+        gas_charge(
+            cgas,
+            ggas,
+            profiler,
+            (Bytes32::LEN as u64) * new_storage_gas_per_byte,
```
xgreenx, line 995: I think we need to charge static cost and per byte=)

xgreenx (commented):
The issues says:
> We need to find all opcodes that modify the storage first. For that, we can find all places where we use methods from the InterpreterStorage trait.

It is similar to the issue https://github.com/FuelLabs/fuel-core/issues/1239

Voxelot:
Should we also charge for 32  bytes(asset id) + 8 bytes (amount) when inserting a new asset type into the balances tree?

xgreenx:
> Should we also charge for 32 bytes(asset id) + 8 bytes (amount) when inserting a new asset type into the balances tree?
Yep
> We need to find all opcodes that modify the storage first. For that, we can find all places where we use methods from the InterpreterStorage trait.

fuel-tx/src/transaction/consensus_parameters/gas.rs
Dentosal: I don't think it makes any sense to future-proof this for the dynamic storage value support. The whole API is going to change anyway.

fuel-vm/src/interpreter/blockchain.rs
Dentosal: What static cost? The opcode itself already costs for instertion, this is only for the new slot which has a fixed cost.

xgreenx: It is related to the comment about `DepenentCost`=) `base` cost can include inserting the new entry into the storage(it will represent the burden of the SMT), and `deps_per_unit` will be gas per byte that you will multiply with `32` or `64`, but in the later versions with dynamic storage by the size of the value.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -654,10 +727,28 @@ where
...
+                pc: self.pc.as_ref(),
+                is: self.is,
+                current_contract: Some(*contract_id),
+                profiler: self.profiler,
+            };
+            gas_charge(
+                self.cgas,
+                self.ggas,
+                profiler,
+                ((AssetId::LEN + size_of::<u64>()) as u64)
```
xgreenx, line 747:
```suggestion
                ((AssetId::LEN + WORD_SIZE) as u64)
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -913,7 +1017,23 @@ pub(crate) fn state_write_word<S: InterpreterStorage>(
...
+            pc: pc.as_ref(),
+            is,
+            current_contract,
+            profiler,
+        };
+        gas_charge(
+            cgas,
+            ggas,
+            profiler,
+            (Bytes32::LEN as u64) * new_storage_gas_per_byte,
```
xgreenx, line 1034:
```suggestion
            (2 * Bytes32::LEN as u64) * new_storage_gas_per_byte,
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -1139,11 +1266,27 @@ fn state_write_qword<S: InterpreterStorage>(
...
+            is,
+            current_contract,
+            profiler,
+        };
+        gas_charge(
+            cgas,
+            ggas,
+            profiler,
+            // Overflow safety: unset_count * 32 can be at most VM_MAX_RAM
+            (unset_count as u64) * (Bytes32::LEN as u64) * new_storage_gas_per_byte,
```
xgreenx, line 1287:
```suggestion
            (unset_count as u64) * ( 2 * Bytes32::LEN as u64) * new_storage_gas_per_byte,
```

xgreenx (commented):
Looks good=) Waiting for rest of opcodes

fuel-vm/src/interpreter/contract.rs
```diff
@@ -250,7 +288,24 @@ impl<'vm, S, Tx> TransferCtx<'vm, S, Tx> {
...
+                is: self.is,
+                current_contract: internal_context,
+                profiler: self.profiler,
+            };
+            gas_charge(
+                self.cgas,
+                self.ggas,
+                profiler,
+                // Overflow safety: unset_count * 32 can be at most VM_MAX_RAM
+                (2 * Bytes32::LEN as u64) * self.new_storage_gas_per_byte,
```
xgreenx, line 306:
```suggestion
                (Bytes32::LEN as u64 + WORD_SIZE) * self.new_storage_gas_per_byte,
```

fuel-vm/src/interpreter/flow.rs
```diff
@@ -546,13 +550,30 @@ where
...
+                is: self.registers.system_registers.is.as_ref(),
+                current_contract: self.current_contract,
+                profiler: self.profiler,
+            };
+            gas_charge(
+                self.registers.system_registers.cgas.as_mut(),
+                self.registers.system_registers.ggas.as_mut(),
+                profiler,
+                // Overflow safety: unset_count * 32 can be at most VM_MAX_RAM
+                (2 * Bytes32::LEN as u64) * self.new_storage_gas_per_byte,
```
xgreenx, line 573:
```suggestion
                (Bytes32::LEN as u64 * WORD_SIZE) * self.new_storage_gas_per_byte,
```

fuel-vm/src/interpreter/contract.rs
```diff
@@ -196,8 +230,12 @@ struct TransferCtx<'vm, S, Tx> {
     context: &'vm Context,
     balances: &'vm mut RuntimeBalances,
     receipts: &'vm mut ReceiptsCtx,
+    profiler: &'vm mut Profiler,
```
xgreenx, line 233: You forgot to update `transfer_output`

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -318,8 +333,18 @@ where
         rb: RegisterId,
         c: Word,
     ) -> IoResult<(), S::DataError> {
-        let (SystemRegisters { fp, pc, .. }, mut w) =
-            split_registers(&mut self.registers);
+        let new_storage_gas_per_byte = self.gas_costs().new_storage_per_byte;
```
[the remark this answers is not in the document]
Dentosal: We do not, but write and read shared the context struct. Fixed in 7863362c.

fuel-vm/src/interpreter/contract.rs
Dentosal: Fixed in c5931980

fuel-vm/src/interpreter/flow.rs
Dentosal: Fixed in c5931980.

fuel-vm/src/interpreter/contract.rs
Dentosal: I though TRO doesn't create markle tree entries? Am I missing something here?

xgreenx (approved):

PR #637, opened 2023-11-17
Charge for the actual size of the contract in `ccp` opcode

xgreenx:


fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -211,16 +211,37 @@ where
         c: Word,
         d: Word,
     ) -> IoResult<(), S::DataError> {
+        let mut gas_cost = self.gas_costs().ccp;
+        // Charge only for the `base` execution.
+        // We will charge for the contracts size in the `code_copy`.
```
bvrooman, line 216:
```suggestion
        // We will charge for the contract's size in the `code_copy`.
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -211,16 +211,37 @@ where
...
+        self.gas_charge(gas_cost.base())?;
+        gas_cost.set_base(0);
```
bvrooman, line 218: I know we already use this pattern where we:
1. Charge the base of the dependent cost
2. Set the base of the dependent cost to 0
3. Charge the variable amount of the dependent cost

In general, I do not think this pattern makes a lot of sense. The structure of `DependentCost` doesn't line up with how it's used: We basically split the one dependent cost into both a fixed cost and dependent cost so we can charge at different points. We break encapsulation by reading and charging `base` as a fixed cost, then we mutate `base` to 0, which means `CodeCopyCtx` (and other places we do this) requires a nonstandard mutated `DependentCost` input, but there's no type safety or documentation to let developers know. All of this breaks the principle of least surprise.

Maybe we could break up dependent costs into both a `DependentCost` and fixed cost, where dependent cost is just the variable amount and the fixed cost is the base. This way, we have type safety and no gymnastics to fit our two-stage charging pattern.

bvrooman (commented):
One small spelling fix and a general comment. Once you apply the spelling fix, it's good to go.

fuel-vm/src/interpreter/blockchain/code_tests.rs
```diff
@@ -80,20 +80,27 @@ fn test_load_contract() -> IoResult<(), Infallible> {
 fn test_code_copy() -> IoResult<(), Infallible> {
```
MitchTurner, line 80: Is there any coverage that needs to be added to the `blockchain` tests? Like we have all these for `ldc` [here](https://github.com/FuelLabs/fuel-vm/blob/master/fuel-vm/src/tests/blockchain.rs#L314). Specifically, it might be good to have a test that covers the case where contract size is larger than `$rD` and show the cost is proportional to the contract that is being copied rather than the bytes specified.

fuel-vm/src/interpreter/blockchain.rs
xgreenx: I tried to minimise the number of changes. Updated in the https://github.com/FuelLabs/fuel-vm/pull/637/commits/7d9cfdd6a69fb8efc8043f0f723df1acd37de9ce

Dentosal (approved):

Issue #639, opened 2023-11-20
Support SMT proofs generation


xgreenx:
The Fuel protocol uses sparse Merkle trees (SMTs) to generate "commitments" of data to a data set. Sparse Merkle trees are built upon sparse data - that is, data that is non-contiguous. This data can represented as a hash map, where data keys map to data values. As data is entered into the set, we build a tree of hashes (`sha256` hashes of the data and neighbour data) upwards.

Use cases for SMTs include contract storage slots.

A SMT proof is a way to prove that a specific chunk of data is included in a sparse set of data without having to share (for privacy, storage, or performance reasons) the entire set. This enables data-integrity checks in logarithmic time complexity.

By enabling SMT proofs, we can prove the existence or non-existence of data used in the Fuel protocol, including storage slots.

SMT proofs can be in a compressed or uncompressed format.

MitchTurner:
Can we add some context to this issue?

PR #648, opened 2023-12-20
feat: SMT Proofs (Inclusion and Exclusion)

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-vm/issues/639

This PR:
- Adds support for generating SMT proofs (inclusion and exclusion
- Adds support for verifying SMT proofs

Given an SMT, we can generate a proof for any key within the 32 byte key space. If the key is included in the tree, i.e., has been used in an update operation and not deleted, then we generate an inclusion proof. If the key has not been included in the tree, we generate an exclusion proof. Inclusion proofs verify that the given key-value is, in fact, in the tree. If the value undergoing verification does correspond to the given key, verification fails. Exclusion proofs verify that the key is a placeholder. If the given key is not a placeholder in the tree, verification fails.

## Example
Consider the following example:

We have a key-value data set, upon which we are building a sparse Merkle tree. 
The data set is structured as such:
| Key | Value |
|-----|-------|
| 000 | 42 |
| 010 | "Foo" |
| 110 | "Bar" |

The data set contains 3 key values. Because the key space supports up to 8 keys (000 - 111), we know that there are 5 unused slots in the key space. 

For each unused slot, we know that the value is unset. Therefore, the complete key-value data set looks like this:
| Key | Value |
|----|--------|
| 000 | 42 |
| 001 | |
| 010 | "Foo" |
| 011 | |
| 100 | |
| 101 | |
| 110 | "Bar" |
| 111 | |

The corresponding sparse Merkle tree is constructed on the key-value data set:

exmaple_smt

### Example Use Case

We can make statements about the data set. For example:
- **The key 000 has value 42 (inclusion).** This statement is _true_.
- **The key 111 is excluded from the table (exclusion).** This statement is _true_.
- **The key 100 is empty (exclusion).** This statement is _true_.
- **The key 110 has value "Hello" (inclusion)**. This statement is _false_ (the correct value is "Bar").
- **The key 110 is empty (exclusion).** This statement is _false_ (the key has value "Bar").
- **The key 101 has value "Foo" (inclusion).** This statement is _false_ (the key is empty).

We can categorize these statements into one of two categories:
1. I assert the key _K_ has the value _V_ (inclusion).
2. I assert the key _K_ is unset (exclusion).

Assertion 1 is an assertion of knowledge: I assert that key _K_ has value _V_. 1. For assertion 2, an equivalent assertion is: I assert that key _K_ is not in the table. 

For these assertions, we can generate cryptographic proofs. Respectively, these are proofs of inclusion and proofs of exclusion. Verification is the process of confirming a given statement using the proof. Verifying a statement and proof returns a boolean (true or false) output that can confirm the corresponding statement. If verification of the statement returns true, the statement has been proven. If verification of the statement returns false, the statement has not been proven. 

Verification failure _does not imply_ the opposite statement. For example, we _cannot_ assert that a key is in the table without providing a specific value, i.e. "key _K_ is included in the table (with some unknown value _V_)" using a failed exclusion assertion. If verification of the statement "key _K_ has a zero value" returns false, it is possible that _K_  has a non-zero value and is included in the tree; however, a failed proof of exclusion does not necessarily imply inclusion. This is because the proof provided to the verification can be somehow invalid or maliciously altered.

More concretely:

- $`V_{inclusion}(A_{key}, A_{value} = X, P) = true \implies A_{key} \in T, T[A_{key}] = X`$
- $`V_{exclusion}(A_{key}, P) = true \implies A_{key} \notin T`$
- $`V_{inclusion}(A_{key}, A_{value} = X, P) = false \centernot\implies A_{key} \notin T, T[A_{key}] \neq X`$ (Proof may be invalid!)
- $`V_{exclusion}(A_{key}, P) = false \centernot\implies A_{key} \in T`$

where:
- $`V`$ is the verifier.
- $`A`$ is the key-value pair in the statement undergoing verification.
- $`P`$ is a proof that proves the statement undergoing verification
- $`T`$ is the table of key-value data 

Proofs that fail to validate do not imply the opposite statement.


fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -560,6 +581,75 @@ where
...
+            // point where the paths of the requested placeholder and closest
+            // leaf diverge are saturated with the closest leaf's hash. In the
+            // case where the closest leaf is a placeholder, this hash is simply
+            // the zero sum. The hash of any placeholder under this point of
+            // divergence equates to this hash.
+            //
+            let exclusion_proof = ExclusionProof {
+                root,
+                proof_set,
+                path,
```
xgreenx, line 644: We don't need `path` since `key` is a path=)

fuel-merkle/src/sparse/verify.rs
```diff
@@ -0,0 +1,341 @@
...
+        },
+        Node,
+    },
+};
+
+pub fn verify<K: Into<Bytes32>, V: AsRef<[u8]>>(key: K, value: &V, proof: Proof) -> bool {
+    match proof {
+        Proof::InclusionProof(proof) => verify_inclusion(key, value, proof),
+        Proof::ExclusionProof(proof) => {
+            value.as_ref() == empty_sum() && verify_exclusion(key, proof)
```
xgreenx, line 25: The `value.as_ref() == empty_sum()` check truly is not required. It is more of a hack to have one function for both poofs.

fuel-merkle/src/sparse/verify.rs
```diff
@@ -0,0 +1,341 @@
...
+    let ExclusionProof {
+        root,
+        proof_set,
+        path,
+        hash,
+    } = proof;
+
+    if key.into() != path {
+        return false;
+    }
```
xgreenx, line 64: `key` already is a path, we don't need to store it inside of the `ExclusionProof`

fuel-merkle/src/sparse/verify.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    let mut current = hash;
```
xgreenx, line 66: Hmmm, the first question is why we can't use `Node::Placeholder.hash()` here?

I assume because of the optimization that we have(like replacing the whole subtree with one leaf node). But then the next question is: What if `hash` is a hash of the actual leaf that is present in the tree? Then `ExclusionProof` will return `true` while it should return `false`. In other words, the malicious prover creates proofs of inclusion but converts it into proofs of exclusion by using `node.hash()` as a `hash` in the `ExclusionProof`

fuel-merkle/src/sparse/verify.rs
bvrooman: The point of storing the path (which is originally a key) in the `ExclusionProof` is to make sure we verifying exclusion for the same key used to generate the exclusion proof. Because we can generate exclusion proofs for placeholders, proof generation may generate the same proof for multiple keys with placeholder values. Therefore, we may want a way to differentiate the proofs based on a unique property which, in this case, is the key.

fuel-merkle/src/sparse/verify.rs
bvrooman: Proof of exclusion is more than just proving that the key is a placeholder. If it was just proving that the key has a placeholder value, we could just check that the value was a placeholder, and we'd be done. And if it is a placeholder, we still need to know that the placeholder belongs to the tree.

You're correct: `hash` is not guaranteed to be the the placeholder value; it is defined by the value of the closest leaf to the requested key when generating the proof:
```
hash:  *actual_leaf.hash()
```
This is deliberate. The hash may or may not be a placeholder. While a placeholder is always an exclusion proof, a non-placeholder can be either an exclusion or inclusion proof. 

Why?

When generating a proof, we need a leaf that is part of the tree we for which we are generating the proof. Based on the current key-values of the tree, the 'actual leaf' (i.e., the closest leaf to the requested leaf during traversal) may be a placeholder or a non-placeholder. To later validate the proof, we need the hash of this leaf and the related proof set.

When validating an exclusion proof, we need to know that the 'actual leaf' (placeholder or non-placeholder) **belongs** to the tree. We are essentially doing a find operation on [disjoint sets](https://en.wikipedia.org/wiki/Disjoint-set_data_structure): We rebuild the root from the hash of the actual leaf and proof set, and compare roots to prove that this element belongs to this set. This also means that even placeholders can belong or not belong to the tree, which is why we need to be specific.

bvrooman: Regarding the ability to produce erroneous proofs: It's possible we need to add additional logic to prevent someone from modifying a proof that still validates. I can add more tests to see if it is possible to forge a false positive proof.

fuel-merkle/src/sparse/verify.rs
bvrooman: If the value is not the empty sum, then we already know that the exclusion proof should fail. 

I'm not sure I agree that this is a hack, but I can that check into `verify_exclusion` if you prefer. I can also make the `verify_*` functions into member functions on the `*Proof` structs themselves.

xgreenx (changes requested):
We need an additional call to double check it=)

bvrooman:
I've spent some time reworking the PR based on the above discussion. Part of that rework involves exclusion proofs, where we now attempt to rebuild the root starting with a placeholder (with the zero sum). We know that this is problematic when the tree has only 1 node, which is to say, only 1 leaf, because the root node is the leaf node, and we have no placeholders contributing to the root.

This actually extends to any case where the root is not composed of any placeholders, and this is theoretically possible with many configurations of  leaf nodes. Without a placeholder contributing to the root, we can never rebuild the root starting with a placeholder.

I'm investigating how other SMT implementations work around this, including Celestia and Libra, but I haven't reached any conclusions yet. However, this may include spec and implementation changes.

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,692 @@
...
+
+impl ExclusionProof {
+    pub fn verify<K: Into<Bytes32>>(&self, key: K) -> bool {
+        let Self {
+            root,
+            proof_set,
+            leaf,
+        } = self;
+        let key = key.into();
+        let mut current = *leaf.hash();
```
bvrooman, line 102: Based on how the proof was generated, the leaf can be either a real leaf or a placeholder. In the case of a placeholder, the hash will be the zero sum. In the case of a real leaf, this leaf corresponds to the "closest" leaf available at the time of generation, and the hash will correspond to that leaf.

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,692 @@
...
+            .field("Proof set", &proof_set)
+            .finish()
+    }
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct ExclusionProof {
+    pub root: Bytes32,
+    pub proof_set: ProofSet,
+    pub(crate) leaf: Node,
```
bvrooman, line 91: We use a `Node` here because `Node` contains the preimage of the hash, rather than the image. If we only stored the image, i.e. `hash: Bytes32` instead, a user could modify the exclusion proof to:
- remove the proof set
- use a `hash` that matches `root`
This would mean that `verify` necessarily returns true.

This may still be possible by replacing `leaf` with a placeholder node and `root` with the placeholder value.

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -557,7 +585,75 @@ where
     }
 }
 
+impl<TableType, StorageType, StorageError> MerkleTree<TableType, StorageType>
+where
+    TableType: Mappable<Key = Bytes32, Value = Primitive, OwnedValue = Primitive>,
+    StorageType: StorageInspect<TableType, Error = StorageError>,
+{
+    pub fn generate_proof<K: Into<Bytes32>>(
```
xgreenx, line 593: You need to use `MerkleTreeKey` as an input type or hash it under the hood.

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,692 @@
...
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct InclusionProof {
+    pub root: Bytes32,
+    pub proof_set: ProofSet,
+}
+
+impl InclusionProof {
+    pub fn verify<K: Into<Bytes32>, V: AsRef<[u8]>>(&self, key: K, value: &V) -> bool {
```
xgreenx, line 60: You need to use `MerkleTreeKey` as an input `key` type or hash it under the hood.

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,692 @@
...
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct ExclusionProof {
+    pub root: Bytes32,
+    pub proof_set: ProofSet,
+    pub(crate) leaf: Node,
+}
+
+impl ExclusionProof {
+    pub fn verify<K: Into<Bytes32>>(&self, key: K) -> bool {
```
xgreenx, line 95: You need to use `MerkleTreeKey` as an input `key` type or hash it under the hood.

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,692 @@
...
+        let Self {
+            root,
+            proof_set,
+            leaf,
+        } = self;
+        let key = key.into();
```
xgreenx, line 101: You are ignoring the `key`. You need to verify that `key` is somehow linked to the `leaf`. Otherwise, I can generate proof of inclusion and use it as proof of exclusion(I pushed a test to show this behavior).

xgreenx (changes requested):

fuel-merkle/src/sparse/proof.rs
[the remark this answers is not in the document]
bvrooman: `Node` is private to the crate, so we can't give `leaf` a higher level of exposure than `crate` either.

fuel-merkle/src/sparse/proof.rs
xgreenx: Almost fixed=D The problem is that getter `leaf_key` has an internal assert. So if someone provides a `node,` it will fail.

Added a test case with a panic=)

I think if the node is not a placeholder or a leaf, we need to return false. Overwise the proof provider may try to trick us=)

fuel-merkle/src/sparse/proof.rs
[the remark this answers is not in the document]
xgreenx: I think it should be public and `Node` type as well. Otherwise, it is not possible to transfer this object between applications. We need to also implement serialization and deserialization somehow=)

Also, it would be nice if we implement `serde` on top of that

xgreenx (commented):
I think it is bad that all these functions panics.

image


If these functions are for internal purposes only, they should be private. But since we are exposing the `Node` type, maybe we want to rework how they are used.

Or, instead of exposing `Node` types, you can create another type for the `ExclusionProof`.

bvrooman:
> Almost fixed=D The problem is that getter leaf_key has an internal assert. So if someone provides a node, it will fail.
>
> Added a test case with a panic=)
>
> I think if the node is not a placeholder or a leaf, we need to return false. Overwise the proof provider may try to trick us=)

Yep, that's a good point. 

The `assert`s should be `debug_assert`. I can fix that. 
I will also add a check that the node in the proof is a leaf node/placeholder, and if not, return `false`.

> I think it should be public and Node type as well. Otherwise, it is not possible to transfer this object between applications. We need to also implement serialization and deserialization somehow=)
>
>Also, it would be nice if we implement serde on top of that

Maybe it makes sense to store a subset of node data in the proof, rather than the whole node? E.g., just `leaf_key` and `leaf_data`, and we reconstruct the leaf from this using something like `Node::new(0 /*height*/, Prefix::Leaf, leaf_key, leaf_data)`. This ensures that there is no way to provide an internal node (`Prefix::Node`). This also means we don't need to expose the `Node` structure further either.

bvrooman:
Re: serde for proofs: https://github.com/FuelLabs/fuel-vm/issues/694

bvrooman:
This PR is ready for the next round of reviews.

I'm leaving `serde` for a [separate task](https://github.com/FuelLabs/fuel-vm/issues/694), because that will require changes that are not in scope of proofs.

fuel-merkle/src/sparse/node.rs
```diff
@@ -195,22 +195,22 @@ impl Node {
...
-        assert!(self.is_node());
+        debug_assert!(self.is_node());
         self.bytes_lo()
     }
 
     pub fn right_child_key(&self) -> &Bytes32 {
-        assert!(self.is_node());
+        debug_assert!(self.is_node());
         self.bytes_hi()
     }
```
xgreenx, line 215: I think public getters should return `Option` instead.

For internal usage, you can create a private version of this function. Maybe even it is better to mark them `unsafe`

fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -557,7 +612,71 @@ where
...
+            // However, in both cases, the path defined by the requested
+            // placeholder is sufficiently close: All branches stemming from the
+            // point where the paths of the requested placeholder and closest
+            // leaf diverge are saturated with the closest leaf's hash. In the
+            // case where the closest leaf is a placeholder, this hash is simply
+            // the zero sum. The hash of any placeholder under this point of
+            // divergence equates to this hash.
+            //
+            let leaf = actual_leaf.clone().into();
+            let exclusion_proof = ExclusionProof { proof_set, leaf };
```
xgreenx, line 671: I like your idea of returning the key and value in the case of the leaf. You can return an enum, which is either a placeholder or (key, value). And in the proof function, we can always create a leaf node.

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,795 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let proof_set = self.proof_set.iter().map(hex::encode).collect::<Vec<_>>();
+        f.debug_struct("InclusionProof")
+            .field("Proof set", &proof_set)
+            .finish()
+    }
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct ExclusionLeaf {
```
xgreenx, line 82: It should be an enum

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,795 @@
...
+    fn leaf_key(&self) -> &Bytes32 {
+        &self.leaf_key
+    }
+
+    fn leaf_value(&self) -> &Bytes32 {
+        &self.leaf_value
+    }
+
+    fn is_placeholder(&self) -> bool {
+        self.leaf_value() == zero_sum()
```
xgreenx, line 97: We discussed it before, user may use any value, even zero sum. We need to use a separate variant for the case of the placeholder

fuel-merkle/src/sparse/proof.rs
```diff
@@ -0,0 +1,795 @@
...
+    pub leaf: ExclusionLeaf,
+}
+
+impl ExclusionProof {
+    pub fn verify(&self, root: &Bytes32, key: &MerkleTreeKey) -> bool {
+        let Self { proof_set, leaf } = self;
+        if !leaf.is_placeholder() && *leaf.leaf_key() == key.as_ref() {
+            return false;
+        }
+        let mut current = leaf.hash();
```
xgreenx, line 139: We want to process enum here. If the `leaf` is a `(key, value)`, then we want to use `Node::create_leaf(key, value)`. Otherwise we want to use `Node::create_placeholder()`.

After you can call `node.hash()`.

fuel-merkle/src/sparse/node.rs
bvrooman: `Node` is public only to the `fuel-merkle` crate, i.e. `pub(crate) enum Node`. The only place `Node` should ever be used is in the context of the `MerkleTree`, where we already know what kind of node we are using. There is no reason for it to be used in a context where we do not know the kind of node, such as user space or another crate, which is why it is marked `pub(crate)`.  Therefore, I do not agree that these should return `Option` - we always know what data we are accessing, i.e, we could always `unwrap` the `Option`; we would never need to branch based on the variant. 

The `debug_assert`s are there for `fuel-merkle` developers who might modify the update/delete algorithms, not for consumers of the `fuel-merkle` crate. But if you prefer, I can mark these functions as `pub(crate)` as well to make that explicit. 

I also disagree that these functions are unsafe, because it is always "safe" to call them, even in the wrong context. You cannot create an unsafe/incoherent state using these functions.

fuel-merkle/src/sparse/proof.rs
bvrooman: Yes, but in this context, we are looking at the "exclusion leaf" (the nearest included leaf to the excluded key). The `value` of the exclusion leaf is the hash of the user data for this leaf, not the user data itself. This is how this version allows us to use the zero sum for user data.
But, either way, we can use an enum here to simplify this.

fuel-merkle/src/sparse/node.rs
xgreenx: If they are used only within the crate, let's mark them `pub(crate)` to avoid confusion in the future=) 

They are unsafe from a business perspective. I only brought attention to it because [this](https://github.com/FuelLabs/fuel-vm/pull/648/commits/0977a793f369a4a159978b40c5d8b24bc301ea94) commit shows that it is not always true. And from the API perspective, it would be right to have a separate bunch of functions only for internal `MerkleTree` work. And `pub` or `pub(crate)` functions that may fail outside of the `MerkleTree` implementation.

The `unsafe` for internal functions will show that it is dangerous to use. But the comment may describe why it is safe. Or we can always return an option and write `expect` and the reason why it is expected to be valid.

fuel-merkle/src/sparse/proof.rs
xgreenx: It is true that, at this point, we have a hash of the user's value. But if the hash is zero, we will treat it as a placeholder, while it is not true. Enum fixes that case.

fuel-merkle/src/sparse/node.rs
bvrooman: > If they are used only within the crate, let's mark them pub(crate) to avoid confusion in the future=)

We can also use `pub(in crate::sparse)` to restrict this scope further. We could even rearrange the file structure to put `Node` under the `merkle_tree` module and do pub(in crate::sparse::merkle_tree)` or really just `pub(super)` to make this very explicit. Any part of the interface marked `pub` is therefore also `pub(in crate::sparse::merkle_tree)` because of Rust's visibility rules.

 > They are unsafe from a business perspective. I only brought attention to it because [this](https://github.com/FuelLabs/fuel-vm/pull/648/commits/0977a793f369a4a159978b40c5d8b24bc301ea94) commit shows that it is not always true. And from the API perspective, it would be right to have a separate bunch of functions only for internal MerkleTree work. And pub or pub(crate) functions that may fail outside of the MerkleTree implementation.
 
I agree that we should differentiate between external and internal APIs, and that they can have different safety guarantees. At the moment, there is really only an internal API for `Node`. Putting `Node` in `Proof` violated that agreement by putting `Node` in user-space and mixing business logic with internal logic. That was bad! `Node` should only be concerned with internal logic, never business logic. Enforcing a more narrow visibility on `Node` will help reduce errors like that in the future.

To summarize: The interface was unsafe from a business perspective, but it was wrong to use it in business logic. I think isolating `Node` to internal logic is the first fix. If you still think `unsafe` should be used in this context, then I will add `unsafe` here.

bvrooman: Have a look at https://github.com/FuelLabs/fuel-vm/pull/696. I this is a bit nicer: Here, `Node` is completely restricted to the `merkle_tree` module using `pub(super)`, making it very explicit that the only consumer is `merkle_tree`, and that there are consumers external to `merkle_tree`. This means that the API is strictly internal, and we can always analyze it from an internal perspective.

xgreenx (approved):

fuel-merkle/src/sparse/proof.rs
```diff
@@ -58,9 +58,15 @@ pub struct InclusionProof {
 impl InclusionProof {
     pub fn verify(&self, root: &Bytes32, key: &MerkleTreeKey, value: &[u8]) -> bool {
         let Self { proof_set } = self;
+
+        if proof_set.len() > u32::MAX as usize {
```
bvrooman, line 62: Good catch! We can actually make this 
```
if proof_set.len() > 256 {
```
since a valid proof set cannot have more than log(n) side nodes.

xgreenx (approved):

Issue #652, opened 2024-01-04
Add user-data to ECAL


Dentosal:
The ECAL instruction handler should be able to store runtime variables (e.g. channel, database connection) somewhere. Change ECAL handler to be a struct implementing approximately the following API:

```rust
trait EcalHandler {
    fn ecal<S, Tx>(
        vm: &mut Interpreter<S, Tx, Self>,
        a: RegId,
        b: RegId,
        c: RegId,
        d: RegId,
    ) -> SimpleResult<()>;

  fn on_reset(vm: &mut Interpreter<S, Tx, Self>) {}
}
```

PR #653, opened 2024-01-04
Add state to ECAL handler

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/652

### DIFF
--- a/fuel-vm/src/interpreter/ecal.rs
+++ b/fuel-vm/src/interpreter/ecal.rs
@@ -88,4 +88,14 @@ where
             Ok(())
         }
     }
+
+    /// Read access to the ECAL state
+    pub fn ecal_state(&self) -> &Ecal {
+        &self.ecal_state
+    }
+
+    /// Write access to the ECAL state
+    pub fn ecal_state_mut(&mut self) -> &mut Ecal {
+        &mut self.ecal_state
+    }
 }
--- a/fuel-vm/src/interpreter/ecal.rs
+++ b/fuel-vm/src/interpreter/ecal.rs
@@ -20,7 +20,7 @@ use super::{
 };
 
 /// ECAL opcode handler
-pub trait EcalHandler: Default + Clone
+pub trait EcalHandler: Clone
 where
     Self: Sized,
 {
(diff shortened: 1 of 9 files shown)


fuel-vm/src/interpreter/ecal.rs
```diff
@@ -88,4 +88,38 @@ where
...
+    }
+
+    /// Write access to the ECAL state
+    pub fn ecal_state_mut(&mut self) -> &mut Ecal {
+        &mut self.ecal_state
+    }
+
+    /// Replace the ECAL state and handler.
+    /// Useful when the handler does not implement `Default`.
+    pub fn with_ecal<NewEcal: EcalHandler>(
```
xgreenx, line 104: It would be nice to have a constructor that allows you also to create an instance of the `Interpreter` with `Ecal` that doesn't support `Default`.

Because right now you need to create `Interpreter` with empty `Ecal` and replace it after, which is strange=)

xgreenx: If we have a new constructor, do we really need that method then?

MitchTurner:
ECAL still kinda confuses me. I understand it as adding custom operation to your code, but I haven't figured out how this would work in practice yet. I don't understand how you could include these operations despite them not existing on the block producer's machine, etc.

With that in mind, could you help me understand the context of this "internal state"?

fuel-vm/src/interpreter/ecal.rs
Dentosal: Changed to work like this in 6f32f3f0.

Dentosal:
> ECAL still kinda confuses me. I understand it as adding custom operation to your code, but I haven't figured out how this would work in practice yet. I don't understand how you could include these operations despite them not existing on the block producer's machine, etc.
> 
> With that in mind, could you help me understand the context of this "internal state"?

Everyone in a network must use the exact same ecal handler. However, say you wanted to test a fork of Fuel-chain with some custom operation added. You don't have to fork the VM for that, and maybe we could even parameterize fuel-core as well. Or you want to do some local testing and want to be able to write results out from the VM in a convenient way. Essentially ecal transforms fuel-vm from a smart-contract-only vm to a generic vm that can be used for anything, and that can be useful for some stuff.

The internal state is needed when doing external operations. Imagine we want to write programs that for some reason run inside the fuel-vm but need an access to a database. The access can be provided by ecal, but we need to keep the connection open somehow. The state can be used to contain the connection.

xgreenx (approved):

MitchTurner:
> Everyone in a network must use the exact same ecal handler.

Okay. That helps a ton. Thanks.

I still have to guess specific use cases, but I'm getting a better idea of what those might be.

PR #668, opened 2024-01-30
Remove non-exhaustive attribute for versionable types

MitchTurner:
We decided that the convenience of having these backward compatible isn't worth the security cost of losing type guarantees.


xgreenx:
It would be nice to revert all changes from https://github.com/FuelLabs/fuel-vm/pull/654/files =)

xgreenx (approved):

PR #670, opened 2024-02-01
DA compression for fuel-tx types

Dentosal:
Part of https://github.com/FuelLabs/fuel-core/issues/1605. See the corresponding fuel-core PR as well: https://github.com/FuelLabs/fuel-core/pull/1609

This PR adds `Compact` derive macro and `fuel-compression` machinery, that are used to convert `Transaction` and contained types into `CompressedTransaction` and `Compressed*`. The compact types are equivalent to the original types, except that malleable fields are removed, and often-repeated types like `AssetId` and `Address` are converted to three-byte keys to a separate database, called temporal storage. `UtxoId`s (which are used only once) are converted to use `TxPointer` instead.


fuel-compression/src/registry/block_section.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+impl<T: Table> WriteTo<T>
+where
+    T::Type: PartialEq,
+{
+    /// Reverse lookup.
+    /// TODO: possibly add a lookup table for this, if deemed necessary
+    pub fn lookup_value(&self, needle: &T::Type) -> Option<Key<T>> {
```
xgreenx, line 46: I think we put too much business logic here with overflowing and counters. I feel like it should be part of the `fuel-core` while `fuel-compression` is responsible only for iterating over the fields and calling `next_key::<Table>(value)`. The type that implements `next_key` may be decided in which way he wants to do that.

The `count` function also sounds like something related to the exact algorithm implementation.

xgreenx (commented):
Thank you nice PR!=) It is cool that `Compact` can be derived almost for all types.

I feel like we put too many details about the implementation of the algorithm and how keys are created, and it seems we can simplify that part a lot to be agnostic. But the main with compaction looks good; I will review it one more time, more precisely when PR is in final shape.

BTW, compilation fails for me:

image

fuel-compression/src/compaction.rs
```diff
@@ -0,0 +1,382 @@
...
+where
+    T: Compactable + Clone + Serialize + for<'a> Deserialize<'a>,
+{
+    type Compact = ArrayWrapper<S, T::Compact>;
+
+    fn count(&self) -> CountPerTable {
+        let mut count = CountPerTable::default();
+        for item in self.iter() {
+            count += item.count();
+        }
```
bvrooman, line 193:
```suggestion
        self.iter().fold(count, |acc, item| acc + item.count());
```

bvrooman (commented):
Reviewed 6/41 so far, will continue review at the next chance I get

fuel-compression/src/compaction.rs
```diff
@@ -0,0 +1,382 @@
...
+    Ok(tmp.map(|v| unsafe { v.assume_init() }))
+}
+
+impl<const S: usize, T> Compactable for [T; S]
+where
+    T: Compactable + Clone + Serialize + for<'a> Deserialize<'a>,
+{
+    type Compact = ArrayWrapper<S, T::Compact>;
+
+    fn count(&self) -> CountPerTable {
```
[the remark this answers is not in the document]
Dentosal: This fully monomorphizes. Zero-cost abstraction.

fuel-compression/src/registry/block_section.rs
```diff
@@ -0,0 +1,197 @@
...
+        if *needle == T::Type::default() {
+            return Some(Key::DEFAULT_VALUE);
+        }
+
+        let mut key = self.start_key;
+        for v in &self.values {
+            if v == needle {
+                return Some(key);
+            }
+            key = key.next();
```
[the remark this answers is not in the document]
Dentosal: It's allowed to overflow. Wrapping around is the expected behavior.

fuel-compression/src/registry/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+                fn get_mut(&mut self) -> &mut WriteTo<tables::$name> {
+                    &mut self.$name
+                }
+            }
+        )*
+    };
+}
+
+tables!(
+    AssetId: [u8; 32],
```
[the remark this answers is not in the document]
Dentosal: Yep.

fuel-tx/src/transaction/types/input/message.rs
```diff
@@ -143,20 +177,24 @@ pub mod specifications {
...
-    Specification: MessageSpecification,
+    Specification: MessageSpecification + Clone,
 {
     /// The sender from the L1 chain.
+    #[cfg_attr(feature = "da-compression", da_compress(registry = "Address"))]
     pub sender: Address,
     /// The receiver on the `Fuel` chain.
+    #[cfg_attr(feature = "da-compression", da_compress(registry = "Address"))]
     pub recipient: Address,
     pub amount: Word,
```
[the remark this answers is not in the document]
Dentosal: Because we can find get from L1 message?

fuel-tx/src/transaction/types/input/message.rs
```diff
@@ -143,20 +177,24 @@ pub mod specifications {
...
     pub nonce: Nonce,
     #[derivative(Debug(format_with = "fmt_as_field"))]
     pub witness_index: Specification::Witness,
     #[derivative(Debug(format_with = "fmt_as_field"))]
+    #[cfg_attr(feature = "da-compression", da_compress(skip))]
```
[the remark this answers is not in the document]
Dentosal: Ok sure. We can keep it for now.

fuel-compression/src/compaction.rs
Dentosal: I find the plain loop way easier to read.

fuel-compression/src/registry/block_section.rs
Dentosal: I agree; if I can get the logic separated it would be really nice. I'll look into it.

fuel-tx/src/transaction/types/utxo_id.rs
```diff
@@ -20,6 +20,7 @@ use rand::{
 #[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "typescript", wasm_bindgen::prelude::wasm_bindgen)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "da-compression", derive(fuel_compression::Compact))]
 #[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
 pub struct UtxoId {
```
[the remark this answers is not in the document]
Dentosal: I'm not planning to do that here, since that's separate from the BiggerType -> Key separation I'm truing to achieve in the PR. Maybe I'll add it here if I can refactor like this https://github.com/FuelLabs/fuel-vm/pull/670#discussion_r1502669109

fuel-tx/src/transaction/types/output.rs
```diff
@@ -23,25 +21,34 @@ pub use repr::OutputRepr;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, strum_macros::EnumCount)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Deserialize, Serialize)]
+#[cfg_attr(feature = "da-compression", derive(fuel_compression::Compact))]
+#[derive(canonical::Deserialize, canonical::Serialize)]
 pub enum Output {
     Coin {
+        #[cfg_attr(feature = "da-compression", da_compress(registry = "Address"))]
```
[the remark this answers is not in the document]
Dentosal: Proc macros had some limits around this, but maybe they have been lifted? I'll take a look.

fuel-compression/src/registry/mod.rs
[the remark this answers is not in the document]
Dentosal: da7819fc

fuel-tx/src/transaction/types/output.rs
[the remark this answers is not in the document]
Dentosal: Done in 42e86341

fuel-compression/src/registry/block_section.rs
Dentosal: I've been trying to figure out  what parts could be moved out of fuel-vm. Most of the algorithm implementation, including this whole file, can probably be transferred.

The typed keys have to remain here, as the proc macro cannot be generic over key type. This means the implementation of `Key` has to live here.

The `count` function of the trait also has to live here. The only way to obtain the number of keys in an entity must be part of the trait. While all key allocation algorithms might not use it, many of the sensible ones do.

fuel-compression/src/compaction.rs
```diff
@@ -0,0 +1,382 @@
...
+        {
+            return Ok(key);
+        }
+
+        // Check if the registry contains this value already
+        if let Some(key) = self.reg.index_lookup::<T>(&value)? {
+            let start: Key<T> = self.start_keys.value();
+            let end: Key<T> = self.safe_keys_start.value();
+            // Check if the value is in the possibly-overwritable range
+            if !key.is_between(start, end) {
```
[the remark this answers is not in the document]
Dentosal: (that code is now test only)

The overflow should work correctly in that case. I should probably add a test case showing that it actually works.

fuel-compression/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies]
+anyhow = { version = "1.0" }
+fuel-derive = { workspace = true }
+paste = "1.0"
+postcard = { version = "1.0", features = ["use-std"], optional = true }
+serde = { version = "1.0", features = ["derive"] }
+serde-big-array = "0.5"
+
+[dev-dependencies]
+fuel-compression = { workspace = true, features = ["test-helpers"] }
```
Dentosal, line 22: These self-dependencies to enable features are again causing issues. This time it's the publish-crates-check, which requires the crate to be published to be able to do this.

fuel-tx/src/transaction/types/utxo_id.rs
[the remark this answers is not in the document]
Dentosal: This currently lives in #808

fuel-compression/src/registry/block_section.rs
Dentosal: Moved most of the extra stuff away.  `count` is gone as well :/

fuel-tx/src/transaction/types/input/message.rs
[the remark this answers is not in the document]
Dentosal: @xgreenx can you clarify this?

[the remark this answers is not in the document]
xgreenx: Yes, anyone who sync with the network should listen for events and fetch messages from them.

In the case of fraud-proof, the L1 contract should store this kind of information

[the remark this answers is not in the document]
Dentosal: Added `da_compress(skip)` in 2a7904a8

fuel-tx/src/transaction/types/utxo_id.rs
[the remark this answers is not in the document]
Dentosal: Merged into this pr

fuel-derive/src/compressed.rs
```diff
@@ -0,0 +1,430 @@
...
+        gen impl ::fuel_compression::Compressible for @Self #w_impl {
+            type Compressed = #compressed_name #g;
+        }
+
+        gen impl<Ctx, E> ::fuel_compression::CompressibleBy<Ctx, E> for @Self #w_impl_field_bounds_compress {
+            async fn compress(&self, ctx: &mut Ctx) -> Result<Self::Compressed, E> {
+                Ok(match self { #compress_per_variant })
+            }
+        }
+        gen impl<Ctx, E> ::fuel_compression::DecompressibleBy<Ctx, E> for @Self #w_impl_field_bounds_decompress {
```
xgreenx, line 420: We need to have a separate derive macro for the `DecompressibleBy` because not all types that derive `CompressibleBy` want to derive `DecompressibleBy`. The Coin or Message is an example. They don't have some fields and during decompressing, we need to fetch them from database

fuel-tx/src/lib.rs
```diff
@@ -125,6 +127,9 @@ pub use transaction::{
     Witness,
 };
 
+#[cfg(feature = "da-compression")]
+pub use transaction::CompressedTransaction;
```
xgreenx, line 131: I think we need to export all types related to comression

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,242 @@
...
+
+impl DecompressibleBy<TestCompressionCtx, Infallible> for CompressibleTxId {
+    async fn decompress(
+        key: &TxPointer,
+        ctx: &TestCompressionCtx,
+    ) -> Result<CompressibleTxId, Infallible> {
+        Ok(*ctx.tx_blocks.get_by_left(key).expect("key not found"))
+    }
+}
+
```
xgreenx, line 113: It would be nice to test full compression and decompression with non zero values of the transaction with different inputs and outputs and check that results equal. We could reuse `TransactionFactory` to generate random transactions=)

fuel-tx/src/transaction/types/input/coin.rs
```diff
@@ -30,6 +37,19 @@ mod private {
 }
 
 /// Specifies the coin based on the usage context. See [`Coin`].
+#[cfg(feature = "da-compression")]
+pub trait CoinSpecification: private::Seal {
+    type Witness: AsField<u16>
+        + for<'a> Compressible<Compressed: Clone + Serialize + Deserialize<'a>>;
+    // TODO: We need to compress the predicate as well
```
xgreenx, line 44: We also need to compress predicates, it is another good field for optimization besides script

fuel-tx/src/transaction/types/utxo_id.rs
```diff
@@ -20,10 +24,42 @@ use rand::{
 #[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "typescript", wasm_bindgen::prelude::wasm_bindgen)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+#[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
+pub struct CompressibleTxId(TxId);
```
xgreenx, line 29: I think we just need to implement `CompressibleBy` for `UtxoId` itself, because in the `fuel-core` we have table `UtxoId` -> `Coin` where `Coin` contains `TxPointer`. In this case we don't need `CompressibleTxId`

fuel-tx/src/lib.rs
Dentosal: Do we? I would think the `fuel-core` part only operates on the transaction, and the types inside are never named.

xgreenx: We don't need them. Just if we want to export `CompressedTransaction`, then it also could be useful to export other types to be consistent.

It can be useful during decompression to allows syntaxic like:

```rust
let ComressedCoin { utxod, predicate, ... } = value;
```

fuel-derive/src/compressed.rs
Dentosal: Split the macro in b8011ce3

fuel-tx/src/transaction/types/utxo_id.rs
Dentosal: Makes sense. Changed to use whole `UtxoId` in b3e0718b.

fuel-tx/src/transaction/types/input/coin.rs
Dentosal: This will be a breaking change, since we have to add a wrapper type for the predicate, and the field is public. (If we were using the attribute-based approach, this wouldn't be an issue.)

fuel-tx/src/lib.rs
Dentosal: I'm leaning towards exporting just the `CompressedTransaction`, and all types that do not derive `Decompress`

fuel-compression/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+mod key;
+mod traits;
+
+pub use key::RegistryKey;
+pub use traits::*;
+
+pub use fuel_derive::{
+    Compress,
+    Decompress,
+};
```
xgreenx, line 19: What do you think about renaming macros into `CompressibleBy` and `DecompressibleBy` to show which traits are implemented?

fuel-derive/src/compression/attribute.rs
```diff
@@ -0,0 +1,100 @@
+use proc_macro2::TokenTree as TokenTree2;
+
+use syn::parse::{
+    Parse,
+    ParseStream,
+};
+
+const ATTR: &str = "da_compress";
```
xgreenx, line 8: Maybe it is better to use the name of the macro to link derive macro and its annotations?

fuel-types/src/array_types.rs
```diff
@@ -346,6 +346,7 @@ key!(Address, 32);
 key!(AssetId, 32);
 key!(BlobId, 32);
 key!(ContractId, 32);
+key!(TxId, 32);
```
xgreenx, line 349: It would be nice to also add tests for this type(like we have for other types)

fuel-tx/src/transaction/types/utxo_id.rs
```diff
@@ -28,6 +28,11 @@ pub struct UtxoId {
     output_index: u16,
 }
 
+#[cfg(feature = "da-compression")]
+impl fuel_compression::Compressible for UtxoId {
+    type Compressed = (crate::TxPointer, u16);
```
xgreenx, line 33: Maybe it is better to define `CompressedUtxoId` instead of tuple

fuel-tx/src/transaction/types/script.rs
```diff
@@ -45,16 +50,51 @@ pub struct ScriptMetadata {
...
+impl DerefMut for ScriptCode {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.bytes
+    }
+}
+
+#[cfg(feature = "da-compression")]
+impl fuel_compression::Compressible for ScriptCode {
+    type Compressed = fuel_compression::RegistryKey;
+}
```
xgreenx, line 83: We also need to implement `CompressibleBy` manually by using `CompressionContext`

fuel-tx/src/transaction/types/mint.rs
```diff
@@ -49,11 +49,16 @@ impl MintMetadata {
...
+#[cfg_attr(
+    feature = "da-compression",
+    derive(fuel_compression::Compress, fuel_compression::Decompress)
+)]
 #[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
 #[canonical(prefix = TransactionRepr::Mint)]
 #[derivative(Eq, PartialEq, Hash)]
 pub struct Mint {
     /// The location of the transaction in the block.
+    #[cfg_attr(feature = "da-compression", da_compress(skip))]
```
xgreenx, line 61: I think we need to skip this field during compression, but during decompression, we need to set it. Maybe it is better to not derive `DecompressibleBy` and implement it via `DecompressionContext`

fuel-tx/src/transaction/types/input/predicate.rs
```diff
@@ -0,0 +1,38 @@
...
+impl DerefMut for PredicateCode {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.bytes
+    }
+}
+
+#[cfg(feature = "da-compression")]
+impl fuel_compression::Compressible for PredicateCode {
+    type Compressed = fuel_compression::RegistryKey;
+}
```
xgreenx, line 38: We need to implement `CompressibleBy` by using `CompressionContext`

fuel-tx/src/transaction/types/input/message.rs
```diff
@@ -143,15 +183,20 @@ pub mod specifications {
...
 where
-    Specification: MessageSpecification,
+    Specification: MessageSpecification + Clone,
 {
     /// The sender from the L1 chain.
     pub sender: Address,
     /// The receiver on the `Fuel` chain.
     pub recipient: Address,
+    #[cfg_attr(feature = "da-compression", da_compress(skip))] // Stored on L1
     pub amount: Word,
```
xgreenx, line 200: I think we can skip `sender`, `recipient`, `amount` and `data` since it is available via event form L1

fuel-tx/src/transaction/types/input/message.rs
```diff
@@ -143,15 +183,20 @@ pub mod specifications {
 #[derive(Default, Derivative, Clone, PartialEq, Eq, Hash)]
 #[derivative(Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "da-compression",
+    derive(fuel_compression::Compress, fuel_compression::Decompress)
```
xgreenx, line 188: I think we need tp implement `DecompressibleBy` via `DecompressionContext` manually since on the decompressor side we need to fetch skipped fields

fuel-tx/src/transaction/types/input/coin.rs
```diff
@@ -95,6 +131,10 @@ impl CoinSpecification for Full {
 #[derive(Default, Derivative, Clone, PartialEq, Eq, Hash)]
 #[derivative(Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "da-compression",
+    derive(fuel_compression::Compress, fuel_compression::Decompress)
```
xgreenx, line 136: I think we need tp implement `DecompressibleBy` via `DecompressionContext` manually since on the decompressor side we need to fetch skipped fields

fuel-tx/src/transaction/types/input/coin.rs
```diff
@@ -95,6 +131,10 @@ impl CoinSpecification for Full {
...
+)]
 #[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
 pub struct Coin<Specification>
```
xgreenx, line 139: Fields like `owner`, `amount`, `asset_id`, `tx_pointer` can be skipped since this information should be available from the past

fuel-compression/src/lib.rs
Dentosal: I see the point in having the same name here. I dislike it since the names are confusing. When I see

```rust
#[derive(fuel_compression::Compress)]
```

I know this type can be compressed by some `fuel_compression`-related machinery. When I see

```rust
#[derive(fuel_compression::CompressibleBy)]
```

I'm like compressible by what? `Compressible` might be an improvement.

Also, `Compress` implements both `Compressible` and `CompressibleBy`, and the define macros having the same names as the trait obfuscates this fact even more.

fuel-derive/src/compression/attribute.rs
Dentosal: The same attribute is used in both `Compress` and `Decompress`. Renaming it to `compress` might be a good idea.

fuel-types/src/array_types.rs
Dentosal: Yep, will do.

fuel-tx/src/transaction/types/utxo_id.rs
Dentosal: Definitely. This is a draft PR for a reason.

fuel-derive/src/compression/attribute.rs
Dentosal: Renamed to `compress` in efd2dc67

fuel-tx/src/transaction/types/input/coin.rs
Dentosal: Removed in d23d6cd3c5d78ad0115aab5df39627902cfab374

fuel-tx/src/transaction/types/input/message.rs
Dentosal: Added skips in d23d6cd3c5d78ad0115aab5df39627902cfab374

fuel-tx/src/transaction/types/input/predicate.rs
Dentosal: I'm trying to figure out why not doing that still allows the current tests to compile.

fuel-tx/src/transaction/types/utxo_id.rs
Dentosal: Done in b60e38a4

fuel-tx/src/tests/da_compression.rs
Dentosal: I'm a bit worried about the `TransactionFactory` not producing all kinds of transactions. This code fails for some case, but when I use `TransactionFactory` all txs pass.

Dentosal: Added the full test in 89654f1f

fuel-tx/src/transaction/types/script.rs
Dentosal: Apparently not needed. Removed `De/CompressionContext` traits  in 9b33143.

fuel-tx/src/transaction/types/mint.rs
Dentosal: Apparently not needed. Removed `De/CompressionContext` traits  in 9b33143.

fuel-tx/src/transaction/types/input/predicate.rs
Dentosal: Apparently not needed. Removed `De/CompressionContext` traits  in 9b33143.

fuel-tx/src/transaction/types/input/coin.rs
Dentosal: 9b33143

fuel-tx/src/transaction/types/input/message.rs
Dentosal: 9b33143

fuel-derive/src/compression/compress.rs
```diff
@@ -0,0 +1,270 @@
...
+                FieldAttrs::Skip => {}
+                FieldAttrs::Normal => {
+                    where_clause_push(
+                        &mut w_impl_field_bounds_decompress,
+                        syn::parse_quote! { #ty: ::fuel_compression::DecompressibleBy<Ctx> },
+                    );
+                }
+            }
+        }
+    }
```
xgreenx, line 206:
```suggestion
```

It seems is not used=)

fuel-derive/src/compression/compress.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    quote! {
+        #bound_fields
+        #compressed #construct_fields
+    }
+}
+
+/// Generate a match arm for each variant of the compressed structure
+/// using the given function to generate the pattern body.
+pub fn each_variant_compressed<
```
xgreenx, line 108: It seems we can move it to `decomrpess` file=)

fuel-derive/src/compression/decompress.rs
```diff
@@ -0,0 +1,141 @@
...
+        .bindings()
+        .iter()
+        .map(|binding| {
+            let attrs = FieldAttrs::parse(&binding.ast().attrs);
+            let ty = &binding.ast().ty;
+            let cname = format_ident!("{}_c", binding.binding);
+
+            match attrs {
+                FieldAttrs::Skip => quote! {
+                    let #cname = Default::default();
```
xgreenx, line 32:
```suggestion
                    let #cname = ::core::default::Default::default();
```

fuel-compression/src/lib.rs
xgreenx: > Also, Compress implements both Compressible and CompressibleBy, and the define macros having the same names as the trait obfuscates this fact even more.

Yeah, it is true. But `Compressible` is a super trait of `CompressibleBy`, so it is not so bad=D But I agree that it is hidden property of the macro that can be discovered only from the comment.

But in the case of `DecompressibleBy` everything is clear=) The macro implements the same trait.

But it is up to you, I will not push it futher=)

fuel-derive/src/lib.rs
```diff
@@ -9,22 +10,40 @@
...
+    pub mod compress;
+    pub mod decompress;
+}
+
 synstructure::decl_derive!(
-    [Serialize, attributes(canonical)] =>
-    /// Derives `Serialize` trait for the given `struct` or `enum`.
-    serialize_derive
+    [Compress, attributes(compress)] =>
+    /// Derives `Compressed` trait for the given `struct` or `enum`.
```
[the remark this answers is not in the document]
Dentosal: 635618994

fuel-derive/src/lib.rs
```diff
@@ -9,22 +10,40 @@
...
+    compression::compress::derive
+);
+synstructure::decl_derive!(
+    [Decompress, attributes(compress)] =>
+    /// Derives `Compressed` trait for the given `struct` or `enum`.
```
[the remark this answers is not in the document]
Dentosal: 635618994

fuel-derive/src/compression/compress.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// Construct compressed version of the struct from the original one
+fn construct_compressed(
+    // The structure to construct, i.e. struct name or enum variant path
+    compressed: &TokenStream2,
+    variant: &synstructure::VariantInfo<'_>,
+) -> TokenStream2 {
+    let bound_fields: TokenStream2 = variant
+        .bindings()
+        .iter()
```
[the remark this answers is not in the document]
Dentosal: 635618994

fuel-derive/src/compression/compress.rs
```diff
@@ -0,0 +1,270 @@
...
+                        let #cname = <#ty as ::fuel_compression::CompressibleBy<_>>::compress_with(&#binding, ctx).await?;
+                    }
+                }
+            }
+        })
+        .collect();
+
+    let construct_fields: TokenStream2 = variant
+        .bindings()
+        .iter()
```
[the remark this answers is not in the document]
Dentosal: 635618994

fuel-derive/src/compression/compress.rs
Dentosal: 635618994

fuel-derive/src/compression/decompress.rs
Dentosal: 635618994

xgreenx (approved):

fuel-tx/src/transaction/types/script.rs
```diff
@@ -115,7 +156,7 @@ impl UniqueFormatValidityChecks for Script {
         consensus_params: &ConsensusParameters,
     ) -> Result<(), ValidityError> {
         let script_params = consensus_params.script_params();
-        if self.body.script.len() as u64 > script_params.max_script_length() {
+        if self.body.script.bytes.len() as u64 > script_params.max_script_length() {
```
rafal-ch, line 159: I think this change is technically not needed?

fuel-compression/src/key.rs
```diff
@@ -0,0 +1,69 @@
...
+impl TryFrom<u32> for RegistryKey {
+    type Error = &'static str;
+
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        let v = value.to_be_bytes();
+        if v[0] != 0 {
+            return Err("RegistryKey must be less than 2^24");
+        }
+
+        let mut bytes = [0u8; 3];
```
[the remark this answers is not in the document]
Dentosal: Both very likely compile down to same instructions, my version feels a bit more readable to me.

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,446 @@
...
+async fn can_decompress_compressed_transaction_upload() {
+    let iter = TransactionFactory::<_, Upload>::from_seed(1234).map(|(tx, _)| tx);
+    assert_can_decompress_compressed_transaction(iter).await;
+}
+
+#[tokio::test]
+async fn can_decompress_compressed_transaction_blob() {
+    let iter = TransactionFactory::<_, Blob>::from_seed(1234).map(|(tx, _)| tx);
+    assert_can_decompress_compressed_transaction(iter).await;
+}
```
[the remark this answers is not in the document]
Dentosal: I don't think it makes much sense to do this, it's not like the timing impact is meaningful anyway, and now the code is less readable.

fuel-compression/src/impls.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+    };
+}
+
+identity_compression!(u8);
+identity_compression!(u16);
+identity_compression!(u32);
+identity_compression!(u64);
+identity_compression!(u128);
```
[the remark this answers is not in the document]
Dentosal: We futher serialize these structs with `postcard`, which indeed applies varint encoding.

fuel-compression/src/key.rs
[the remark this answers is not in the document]
Dentosal: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=7962e819b48c5054f7afb378be0ac539

```
playground::a:
	bswap	esi
	test	sil, sil
	je	.LBB4_1
	mov	qword ptr [rdi + 8], 34
	lea	rax, [rip + .L__unnamed_2]
	mov	qword ptr [rdi], rax
	ret

.LBB4_1:
	mov	eax, esi
	shr	eax, 8
	shr	esi, 24
	mov	byte ptr [rdi + 10], sil
	mov	word ptr [rdi + 8], ax
	xor	eax, eax
	mov	qword ptr [rdi], rax
	ret

playground::b:
	cmp	esi, 16777215
	ja	.LBB5_2
	bswap	esi
	mov	eax, esi
	shr	eax, 8
	shr	esi, 24
	mov	byte ptr [rdi + 10], sil
	mov	word ptr [rdi + 8], ax
	xor	eax, eax
	mov	qword ptr [rdi], rax
	ret

.LBB5_2:
	mov	qword ptr [rdi + 8], 34
	lea	rax, [rip + .L__unnamed_2]
	mov	qword ptr [rdi], rax
	ret
```

fuel-tx/src/transaction/types/script.rs
Dentosal: That's true. Removed in b0f52dd.

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,446 @@
...
+#[derive(Debug, PartialEq, Default, Compress, Decompress)]
+pub struct ExampleStruct {
+    pub asset_id: AssetId,
+    pub array: [u8; 32],
+    pub vec: Vec<u8>,
+    pub integer: u32,
+}
+
+#[derive(Debug, PartialEq, Compress, Decompress)]
+pub struct InnerStruct {
```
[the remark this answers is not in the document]
Dentosal: Good catch. I changed the test functions around that a bit, now it's used again. See cbe48f9.

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,446 @@
...
+                    *input.nonce().unwrap(),
+                    MessageInfo {
+                        sender: *input.sender().unwrap(),
+                        recipient: *input.recipient().unwrap(),
+                        amount: input.amount().unwrap(),
+                        data: input.input_data().unwrap_or_default().to_vec(),
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
Dentosal: Removed in 160172991

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,446 @@
...
+                (
+                    *input.utxo_id().unwrap(),
+                    CoinInfo {
+                        owner: *input.input_owner().unwrap(),
+                        amount: input.amount().unwrap(),
+                        asset_id: *input.asset_id(&AssetId::default()).unwrap(),
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
Dentosal: Removed in 160172991

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,446 @@
...
+            asset_id: AssetId::new(rng.gen()),
+            array: rng.gen(),
+            vec: (0..rng.gen_range(0..32)).map(|_| rng.gen::<u8>()).collect(),
+            integer: rng.gen(),
+        };
+        let compressed = original
+            .compress_with(&mut ctx)
+            .await
+            .expect("compression failed");
+        let postcard_compressed =
```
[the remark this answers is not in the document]
Dentosal: The point of the name is to highlight that the value stored here is both:
1. compressed using the da-compression mechanism, and
2. serlialized with postcard.

I renamed the variables in 59ffbd5ee to highlight this.

fuel-tx/src/tests/da_compression.rs
```diff
@@ -0,0 +1,446 @@
...
+    pub array: [u8; 32],
+    pub vec: Vec<u8>,
+    pub integer: u32,
+}
+
+#[derive(Debug, PartialEq, Compress, Decompress)]
+pub struct InnerStruct {
+    pub asset_id: AssetId,
+    pub count: u64,
+    #[compress(skip)]
```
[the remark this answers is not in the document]
Dentosal: We did not, in an unit-test manner. Now we do, added in 476c898fc.

xgreenx (approved):

rafal-ch (approved):
LGTM :+1:

Issue #655, opened 2024-01-11
Use `Vec<u8>` as a value for the `ContractsState` table


xgreenx:
It is a preparation for the smooth update to support [dynamic storage](https://github.com/FuelLabs/fuel-specs/issues/517).

Required changes:
- [x] Change `StorageSlot::value` type to `Vec<u8>` (https://github.com/FuelLabs/fuel-vm/pull/671/)
- [x] Update `fuel-core` to implement `StorageRead` and `StorageWrite` for the `ContractsState` (https://github.com/FuelLabs/fuel-core/pull/1714)

PR #671, opened 2024-02-06
feat: Use Vec of Bytes for Contract State

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-vm/issues/655


Voxelot:
@xgreenx would this negate the performance benefit of being able to directly copy the contract bytecode into vm memory out of the db? I.e. seems like this would require copying the whole bytecode onto the heap first?

fuel-tx/src/transaction/types/storage.rs
```diff
@@ -69,7 +58,7 @@ impl Distribution<StorageSlot> for Standard {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> StorageSlot {
```
Dentosal, line 58: Note: now `sample` doesn't generate from all possible values anymore, which might affect test coverage. Seems to be a non-issue for now.

xgreenx:
> would this negate the performance benefit of being able to directly copy the contract bytecode into vm memory out of the db? I.e. seems like this would require copying the whole bytecode onto the heap first?


@Voxelot It affects the `ContractsState` table, not `ContractsRawCode` table=)

fuel-tx/src/transaction/types/storage.rs
```diff
@@ -124,4 +116,42 @@ mod tests {
             serde_json::from_reader(storage_slots_file).expect("read file");
         assert_eq!(storage_slots.len(), 1);
     }
+
+    #[test]
+    fn test_storage_slot_canonical_serialization() {
```
MitchTurner, line 121: Could you included given/when/then sections for these tests?

fuel-tx/src/transaction/types/storage.rs
```diff
@@ -18,49 +18,38 @@ use rand::{
 
 use core::cmp::Ordering;
 
+pub type StorageData = Vec<u8>;
```
MitchTurner, line 21: I'm curious about the naming of this. Are we intending _all_ storage to move to `Vec` or just for contract state? Or is there no distinction between storage and contract storage?

Just trying to understand the scope of all this.

fuel-tx/src/transaction/types/storage.rs
```diff
@@ -18,49 +18,38 @@ use rand::{
...
     key: Bytes32,
-    value: Bytes32,
+    value: StorageData,
 }
 
 impl StorageSlot {
-    pub const SLOT_SIZE: usize = Bytes32::LEN + Bytes32::LEN;
-
-    pub const fn new(key: Bytes32, value: Bytes32) -> Self {
+    pub const fn new(key: Bytes32, value: StorageData) -> Self {
```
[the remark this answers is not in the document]
bvrooman: This PR now includes manual implementations of serialization/deserialization to maintain a fixed size for storage slots. These implementations will be removed when fully supporting dynamic storage.

fuel-vm/src/storage/interpreter.rs
```diff
@@ -145,17 +146,23 @@ pub trait InterpreterStorage:
         &mut self,
         contract: &ContractId,
         key: &Bytes32,
-        value: &Bytes32,
-    ) -> Result<Option<Bytes32>, Self::DataError> {
-        StorageMutate::<ContractsState>::insert(self, &(contract, key).into(), value)
+        value: &StorageData,
+    ) -> Result<Option<StorageData>, Self::DataError> {
+        // StorageWrite::<ContractsState>::write(self, &(contract, key).into(),
+        // value.into())
```
Dentosal, line 152: Maybe these comments should be removed?

bvrooman: This part specifically is unclear in my mind. We want to use a higher level of abstraction  for the `ContractsState` table, on top of `StorageMutate` and `StorageInspect` by implementing `StorageWrite` and `StorageRead`. This should consolidate the read and write patterns for vector based storage. I'm not sure I understand how/why that's used in this specific context of Merkle storage.

fuel-tx/src/transaction/types/storage.rs
bvrooman: This can be named better - something like `ContractsStateData` could be more congruent with the table, where the key is `ContractsStateKey`. I will make that change, but let me know if you have a suggestion you prefer.

fuel-vm/src/storage/interpreter.rs
Voxelot: Summarizing from the call earlier today - it sounds like we need to implement StorageRead and StorageWrite for the contracts state table. This will also allow us to structure the table using `Vec<u8>` or `[u8]` under the hood and avoid the need for the newtype wrapper.

bvrooman: We will still use the new wrapper type to provide some convenience functions for converting between types, as well as to control the canonical serialization implementation while we need a fixed size.

fuel-tx/src/transaction/types/storage.rs
bvrooman: I renamed it to `ContractsStateData` to illustrate that this is part of the `ContractsState` table alongside the `ContractsState` key.

fuel-tx/src/transaction/types/storage.rs
```diff
@@ -16,51 +15,157 @@ use rand::{
...
+impl IntoIterator for ContractsStateData {
+    type IntoIter = alloc::vec::IntoIter<Self::Item>;
+    type Item = u8;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+// TODO: Remove manual serialization when implementing dynamic storage
```
xgreenx, line 96: We don't need to use this type in the transaction. It should be used only inside of the `ConstractsState` table. In this case you don't need to have custom serialization and deserialziation

fuel-tx/src/transaction/types/storage.rs
```diff
@@ -16,51 +15,157 @@ use rand::{
...
+}
 
 #[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "typescript", wasm_bindgen::prelude::wasm_bindgen)]
 #[derive(Deserialize, Serialize)]
 pub struct StorageSlot {
     key: Bytes32,
-    value: Bytes32,
+    value: ContractsStateData,
```
xgreenx, line 145: Let's not change the type for the `Transaction` to avoid breaking changes and to not affect the serialziation

fuel-tx/src/transaction/types/storage.rs
bvrooman: Ok, I can revert the changes to `StorageSlot` and keep the value type as `Bytes32`.

xgreenx (approved):

xgreenx (approved):

PR #672, opened 2024-02-08
Update fee calculation logic to accept price

MitchTurner:
Closes: https://github.com/FuelLabs/fuel-vm/issues/666


fuel-tx/src/transaction/fee.rs
```diff
@@ -180,15 +173,18 @@ pub trait Chargeable: field::Inputs + field::Witnesses + field::Policies {
...
         used_gas: Word,
+        gas_price: u64,
     ) -> Option<Word> {
         // We've already charged the user for witnesses as part of the minimal gas and all
         // execution required to validate transaction validity rules.
         let min_gas = self.min_gas(gas_costs, fee);
 
         let total_used_gas = min_gas.saturating_add(used_gas);
-        let used_fee = gas_to_fee(total_used_gas, self.price(), fee.gas_price_factor);
+        let used_fee = gas_to_fee(total_used_gas, self.price(), gas_price);
```
Voxelot, line 183:
```suggestion
        let used_fee = gas_to_fee(total_used_gas, gas_price, fee.gas_price_factor);
```

The gas_price_factor is a global consensus constant param, and shouldn't be removed.

fuel-asm/src/args.rs
```diff
@@ -252,7 +252,7 @@ crate::enum_try_from! {
         PolicyTypes = 0x500,
 
         /// Set `$rA` to `tx.policies[0x00].gasPrice`
-        PolicyGasPrice = 0x501,
+        PolicyTip = 0x501,
```
Voxelot, line 255: maybe we should call this "GasTip" in case we have other unit dimensions in the future.

fuel-tx/src/transaction/fee.rs
```diff
@@ -152,21 +145,31 @@ pub trait Chargeable: field::Inputs + field::Witnesses + field::Policies {
...
+    fn min_fee(
+        &self,
+        gas_costs: &GasCosts,
+        fee: &FeeParameters,
+        gas_price: Word,
+    ) -> u128 {
         gas_to_fee(
             self.min_gas(gas_costs, fee),
-            self.price(),
+            gas_price,
```
Voxelot, line 156: should we be adding the tip here?

MitchTurner: I figured I'd do that in the next PR.

fuel-asm/src/args.rs
MitchTurner: Do we want the tip denominated in gas or in the base asset?

fuel-tx/src/transaction/types/mint.rs
```diff
@@ -73,7 +73,7 @@ pub struct Mint {
 impl crate::UniqueIdentifier for Mint {
     fn id(&self, chain_id: &ChainId) -> Bytes32 {
         if let Some(id) = self.cached_id() {
-            return id
+            return id;
```
xgreenx, line 76: It would be nice if we skipped changes like this=D

fuel-vm/src/checked_transaction.rs
```diff
@@ -130,13 +130,18 @@ impl<Tx: IntoChecked + UniqueIdentifier> Checked<Tx> {
 }
 
 #[cfg(feature = "test-helpers")]
-impl<Tx: IntoChecked + Default> Default for Checked<Tx>
+impl<Tx: IntoChecked + Default> Checked<Tx>
```
xgreenx, line 133: You still can implement `Default` trait with default gas price 0 or 1.

fuel-vm/src/checked_transaction.rs
```diff
@@ -880,22 +900,21 @@ mod tests {
     // input
     #[quickcheck]
     fn max_fee_message_input(
```
xgreenx, line 902: The purpose of this test to test different `gas_price` as well, could you return it back, please?=)

fuel-vm/src/checked_transaction.rs
```diff
@@ -841,7 +863,6 @@ mod tests {
     // input
     #[quickcheck]
     fn min_fee_coin_input(
```
xgreenx, line 865: The purpose of this test to test different `gas_price` as well, could you return it back, please?=)

fuel-vm/src/checked_transaction.rs
```diff
@@ -945,7 +964,6 @@ mod tests {
     // input
     #[quickcheck]
     fn min_fee_message_input(
```
xgreenx, line 966: The purpose of this test to test different `gas_price` as well, could you return it back, please?=)

fuel-vm/src/checked_transaction/builder.rs
```diff
@@ -20,27 +20,35 @@ where
     Tx: IntoChecked,
 {
     /// Finalize the builder into a [`Checked<Tx>`] of the correct type
-    fn finalize_checked(&mut self, height: BlockHeight) -> Checked<Tx>;
+    fn finalize_checked(&mut self, height: BlockHeight, gas_price: u64) -> Checked<Tx>;
```
xgreenx, line 23: It seems `gas_price` can be part of the builder as well along with `block_height`

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -474,7 +476,10 @@ where
         Ok(())
     }
 
-    pub(crate) fn run(&mut self) -> Result<ProgramState, InterpreterError<S::DataError>> {
+    pub(crate) fn run(
+        &mut self,
+        gas_price: Word,
```
xgreenx, line 481: It seems can be part of the `InterpreterParams` along with another data=)

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -692,6 +701,7 @@ where
     pub fn deploy(
         &mut self,
         tx: Checked<Create>,
+        gas_price: Word,
```
xgreenx, line 704: The same here, it can be part of the `InterpreterParams`

fuel-tx/src/transaction/types/mint.rs
MitchTurner: It's happening automatically by my formatter. I'm not sure what is different with my fmt configuration.

fuel-vm/src/checked_transaction.rs
MitchTurner: I removed it because it isn't used, since the `gas_price` isn't attached to the tx anymore.

fuel-vm/src/interpreter/executors/main.rs
MitchTurner: I guess this is a question I had about this api in general. Will each instance of the interpreter only use a single price? Do we never share an interpreter between blocks? That's why I went this route.

fuel-tx/src/tests/valid_cases/input.rs
```diff
@@ -495,7 +507,11 @@ fn message_message_coin() {
 
     let block_height = rng.gen();
     let err = tx
-        .check(block_height, &ConsensusParameters::standard())
+        .check(
```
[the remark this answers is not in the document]
MitchTurner: Thinking about this more. There are no `impl`s of `check` and only 4 `impl`s of `check_without_signature`. I'm not too worried about it. It would be a lot more work rn to create a new type--all the places we call those functions. And those places would still need to be changed in the future.

This also goes back to the work I did a while ago breaking down the params so functions only use what they need. As soon as we start grouping things together again it gets harder to uphold the Interface Segregation Principle.

fuel-vm/src/checked_transaction.rs
MitchTurner: Maybe we want both? I'm just worried that someone will write a test that needs the gas_price to be parameterized and there will be some hidden gas_price instead.

fuel-vm/src/checked_transaction/builder.rs
MitchTurner: `gas_price` _isn't_ part of the transaction, though. It's part of the block.

fuel-tx/src/transaction/id.rs
```diff
@@ -277,7 +277,6 @@ mod tests {
 
     fn assert_id_common_attrs<Tx: Buildable>(tx: &Tx) {
         use core::ops::Deref;
-        assert_id_ne(tx, |t| t.set_gas_price(t.gas_price().not()));
```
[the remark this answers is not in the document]
MitchTurner: Yeah. I was planning on doing a followup story for tips.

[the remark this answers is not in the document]
xgreenx: I meant that you don't need to remove and add this part in the other PR. You simply can rename functions and you already are testing it. In this case, you will not lose the code that worked before.

It's okay to update the business logic in the next PR, but here we are testing serialization and deserialization that are the same.

fuel-tx/src/transaction/types/input/snapshot_tests.rs
```diff
@@ -19,7 +19,6 @@ fn tx_with_signed_coin_snapshot() {
             predicate: Empty::new(),
             predicate_data: Empty::new(),
         }))
-        .gas_price(1)
```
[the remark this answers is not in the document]
xgreenx: The idea of this test to serialise all possible fields

fuel-vm/src/checked_transaction.rs
xgreenx: But this test is for testing how fee calculation works in general, not how it works within the transaction.

fuel-vm/src/checked_transaction/builder.rs
xgreenx: It's true, but this type is used only in tests, and for simplicity it could be better=) But it is up to you since you need to update all palces in the `fuel-core`=)

fuel-vm/src/interpreter/executors/main.rs
xgreenx: We always create a new instance of the `Interpreter` for each transaction. We don't have plans to share interpreter, since it is faster to allocate a new memory for `Interpreter` instead of clearing manually the old one, based on our benchmarks=)

fuel-tx/src/transaction/types/input/snapshot_tests.rs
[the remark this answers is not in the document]
MitchTurner: Hmm. Looks like there still is a diff. I'm not sure what this `assertion_line` means.

fuel-vm/src/interpreter/executors/main.rs
MitchTurner: Okay. I'm convinced that putting it on the `Interpreter` is a good idea, but I'm not going to put it on `InterpreterParams` because we have conversions from `ConsensusParams` that would need an additional parameter, if we just put it on `Interpreter` then we can keep those conversions.

I might rename `InterpreterParams` to `InterpreterConsensusParams` because that's a bit more accurate.

MitchTurner: Hmm. I'm having a similar problem with `Interpreter`. We have some `From` conversions for other types that don't have `gas_price` so I'm going have to replace a bunch of `from` methods with new functions. 

It might end up being a better interface long term to have it held on the `Interpreter`, but it's not obvious to me right now if that's the right design. 

I don't hate having the extra param on the execution methods for now.

fuel-vm/src/checked_transaction/builder.rs
MitchTurner: I think I prefer it this way. I'll just have to deal with any costs on the `fuel-core` side :)

fuel-tx/src/tests/valid_cases/input.rs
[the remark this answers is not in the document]
MitchTurner: I'm pretty ambivalent about this one. I agree that it's a smell if we keep changing the interface, but I don't know if it's worth the work changing now.

fuel-vm/src/interpreter/executors/main.rs
Voxelot: What about something like?

```
impl impl From<(ConsensusParameters, GasPrice)> for InterpreterParams {
```

MitchTurner: I thought about going with the tuple approach, but I don't feel like it is very idiomatic.

I'll give the constructor another shot.

fuel-vm/src/interpreter.rs
```diff
@@ -167,30 +166,48 @@ impl Default for InterpreterParams {
...
-            fee_params: value.fee_params,
-            base_asset_id: value.base_asset_id,
-        }
-    }
-}
-
-impl From<CheckPredicateParams> for InterpreterParams {
-    fn from(params: CheckPredicateParams) -> Self {
-        InterpreterParams {
+// impl From<ConsensusParameters> for InterpreterParams {
```
Voxelot, line 169: Should we remove this commented out code?

MitchTurner: Gottit. Thanks.

fuel-vm/src/checked_transaction.rs
```diff
@@ -1606,8 +1703,14 @@ mod tests {
     where
         Tx: Chargeable + field::Inputs + field::Outputs,
     {
-        let available_balances =
-            balances::initial_free_balances(tx, gas_costs, fee_params, base_asset_id)?;
+        let arb_gas_price = 1;
```
Voxelot, line 1706: Why did we stop using quickcheck to set the gas price?

MitchTurner: Oh. I thought I fixed these spots. @xgreenx pointed it out but looks like I missed this one. Thanks.

fuel-vm/src/checked_transaction.rs
```diff
@@ -135,8 +135,30 @@ where
...
     fn default() -> Self {
+        let gas_price = 0;
         Tx::default()
-            .into_checked(Default::default(), &ConsensusParameters::standard())
+            .into_checked(
+                Default::default(),
+                &ConsensusParameters::standard(),
+                gas_price,
+            )
+            .expect("default tx should produce a valid fully checked transaction")
```
xgreenx, line 145: You can replace it with `default_with_gas_price(0)`

fuel-vm/src/checked_transaction.rs
```diff
@@ -192,21 +214,23 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
...
     ) -> Result<Checked<Self>, CheckError>
     where
         Checked<Self>: CheckPredicates,
     {
         let check_predicate_params = consensus_params.into();
-        self.into_checked_basic(block_height, consensus_params)?
+        self.into_checked_basic(block_height, consensus_params, gas_price)?
             .check_signatures(&consensus_params.chain_id)?
-            .check_predicates(&check_predicate_params)
+            .check_predicates(gas_price, &check_predicate_params)
```
xgreenx, line 225: We don't need `gas_price` to execute the predicate

xgreenx (changes requested):

fuel-vm/src/tests/encoding.rs
```diff
@@ -217,7 +217,6 @@ fn transaction() {
             vec![0xfa],
             vec![0xfb, 0xfc],
             Policies::new()
-                .with_gas_price(Word::MAX >> 1)
```
xgreenx, line 220: Could you also use tip here instead of the removing it, please?=) It is related to the encoding, so it is okay to keep

xgreenx (approved):

PR #674, opened 2024-02-10
Update CI Rust version to 1.75

bvrooman:
Upgrades the version of Rust used in CI to 1.75 to address a compiler error for `clap` during CI: `package clap v4.5.0 cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,8 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


MitchTurner (approved):
LGTM. Do we usually keep our nightly up to date as well?

bvrooman:
> LGTM. Do we usually keep our nightly up to date as well?

I don't think we have a policy on our nightly version, but it should probably be a version at least as recent as the stable version

Issue #650, opened 2023-12-29
Should be able to get the AssetId & To for a Change output


dmihal:
Currently, it's not possible to get the "AssetId" or "To" parameters of a change output, even though their value is static and specified in the transaction. It should be possible for any Sway program to view these values.

The issue derives from these lines:

https://github.com/FuelLabs/fuel-vm/blob/0d4fd4500b5f7aa1281a7078b86e593f19ae151f/fuel-vm/src/interpreter/metadata.rs#L479-L488

https://github.com/FuelLabs/fuel-vm/blob/0d4fd4500b5f7aa1281a7078b86e593f19ae151f/fuel-vm/src/interpreter/metadata.rs#L495-L504

Specifically, filtering for `o.is_coin()` should be replaced with something like `o.is_coin() || o.is_change()`

dmihal:
User report:

https://forum.fuel.network/t/obtaining-the-to-address-for-a-change-output/4134/1

PR #675, opened 2024-02-12
Add access to {asset_id,to,amount} fields for Variable and Change outputs

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/650


fuel-vm/src/interpreter/metadata.rs
```diff
@@ -489,14 +489,14 @@ impl<Tx> GTFInput<'_, Tx> {
             GTFArgs::OutputCoinAmount => tx
                 .outputs()
                 .get(b)
-                .filter(|o| o.is_coin())
+                .filter(|o| o.is_coin() || o.is_variable() || o.is_change())
```
Voxelot, line 492: Since the change amount isn't set until post-execution, this will always be zero for change outputs.

Dentosal: True. Do you think we should panic here rather than returning the zero? The old error is `OutputNotFound` which seems a bit misleading. So is returning zero.

Dentosal:
After https://github.com/FuelLabs/fuel-vm/pull/678 accessing fields of `Output::Variable` no longer make sense. Also, as Voxelot pointed out above, `Output::Change` amount is always zero, so better leave that one out as well.

fuel-vm/src/interpreter/metadata.rs
Dentosal: Reverted to old behavior in f7855c9.

xgreenx (approved):

PR #683, opened 2024-02-22
chore: Simplify `InterpreterStorage`

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-vm/issues/655

Currently, the `InterpreterStorage` trait has two issues:
- It imposes an unnecessary constraint on `MerkleRootStorage` which can be removed.
- The getter and setter methods for assets and state are prefixed with `merkle_`; however, implementations of these methods do not necessarily need to Merklize the data, and it is the responsibility of specific implementations to understand when Merklization is needed. 

By removing the `merkle_` prefix, we make it more clear that the methods are simply methods to query or mutate data in interpreter storage. Any Merklization is an internal detail. Additionally, this means that implementations of these methods that do not apply Merklization, i.e. for `MemoryStorage`, are more congruent with the method name.

This enables future interface changes to be done more easily and with greater confidence in their correctness.


fuel-vm/src/storage/memory.rs
```diff
@@ -248,23 +242,6 @@ impl StorageMutate<ContractsAssets> for MemoryStorage {
     }
 }
 
-impl MerkleRootStorage<ContractId, ContractsAssets> for MemoryStorage {
```
Dentosal, line 251: I'm a bit surprised that this can just be removed.

Dentosal (approved):
Seem like this is just renaming and removals. Correctly marked as breaking change, although it's mostly internal.

xgreenx (approved):

Issue #665, opened 2024-01-25
Use `max_fee` policy to charge user


MitchTurner:

MitchTurner:
This was on the Miro board, but I'm not sure what it means :P

xgreenx:
The current codebase converts `max_gas` into `max_fee` and deducts this amount from available balances. 
<img width="1275" alt="image" src="https://github.com/FuelLabs/fuel-vm/assets/18346821/958d98e4-1bcf-432f-97f5-d375b9125e19">

It was okay to do because `gas_price` was statically defined by the transaction itself. But in a new module, since the base gas price is dynamic, the `max_fee` based on the `max_gas` is dynamic too. Instead of using this dynamic value, we need to use `max_fee` defined by the user in the policies if it is available.

MitchTurner:
Are we keeping `script_gas_limit`? I'm still confused about the relationship between gas limit and `max_fee`.

Voxelot:
yes we need to keep script_gas_limit as a way for users to budget how much of their fee goes into script execution vs witness data etc

PR #685, opened 2024-02-23
Use max fee policy to charge user

MitchTurner:
Closes https://github.com/FuelLabs/fuel-vm/issues/665


MitchTurner:
@xgreenx did we also want some check somewhere that the `gas_price` on the `FullyCheckedTx` is correct?

xgreenx:
I think `Interpreter` should verify that `gas_price` is the same.

fuel-vm/src/checked_transaction.rs
```diff
@@ -121,6 +125,68 @@ impl<Tx: IntoChecked> Checked<Tx> {
...
+pub struct Ready<Tx: IntoChecked> {
+    gas_price: Word,
+    transaction: Tx,
+    metadata: Tx::Metadata,
+    checks_bitmask: Checks,
+}
+
+impl<Tx: IntoChecked> Ready<Tx> {
+    /// Consume and decompose components of the `Immutable` transaction.
+    pub fn decompose(self) -> (Word, Tx, Tx::Metadata, Checks) {
```
xgreenx, line 139: I think it makes sense to return `(Word, Checked<Tx>)`. It will give more flexibility to the consumer of this method. They can decide do they need to break down it more or not without losing checkness=)

fuel-vm/src/checked_transaction.rs
```diff
@@ -693,99 +748,89 @@ mod tests {
     fn checked_tx_accepts_valid_tx() {
         // simple smoke test that valid txs can be checked
         let rng = &mut StdRng::seed_from_u64(2322u64);
-        let gas_price = 10;
         let gas_limit = 1000;
         let input_amount = 1000;
         let output_amount = 10;
-        let tx = valid_coin_tx(rng, gas_limit, input_amount, output_amount);
+        let max_fee_limit = 0;
```
xgreenx, line 754: Maybe we want to use non-zero value here=)

fuel-vm/src/checked_transaction.rs
```diff
@@ -693,99 +748,89 @@ mod tests {
...
     #[test]
-    fn checked_tx_accepts_valid_signed_message_input_fees() {
+    fn into_checked__tx_accepts_valid_signed_message_coin_for_fees() {
         // simple test to ensure a tx that only has a message input can cover fees
         let rng = &mut StdRng::seed_from_u64(2322u64);
         let input_amount = 100;
-        let gas_price = 100;
         let gas_limit = 1000;
-        let tx = signed_message_coin_tx(rng, gas_limit, input_amount);
+        let zero_fee_limit = 0;
```
xgreenx, line 778: Maybe we want to use non-zero value here

fuel-vm/src/checked_transaction.rs
```diff
@@ -693,99 +748,89 @@ mod tests {
...
     #[test]
-    fn checked_tx_excludes_message_output_amount_from_fee() {
+    fn into_checked__tx_excludes_message_output_amount_from_fee() {
         // ensure message outputs aren't deducted from available balance
         let rng = &mut StdRng::seed_from_u64(2322u64);
         let input_amount = 100;
-        let gas_price = 100;
         let gas_limit = 1000;
-        let tx = signed_message_coin_tx(rng, gas_limit, input_amount);
+        let zero_fee_limit = 0;
```
xgreenx, line 798: Maybe we want to use non-zero value here

fuel-vm/src/checked_transaction.rs
```diff
@@ -1383,179 +1416,237 @@ mod tests {
...
+            .expect_err("overflow expected");
 
         // then
-        assert_eq!(min_fee_without_tip + tip, min_fee_with_tip);
+        assert!(matches!(err, CheckError::InsufficientMaxFee { .. }));
     }
 
     #[test]
-    fn into_checked_basic__max_fee_calc_includes_tip() {
+    fn into_immutable__tx_fails_if_tip_not_covered() {
```
xgreenx, line 1575: Maybe we want to change the name of all test from `into_immutable_...` -> `into_ready_...`

fuel-vm/src/checked_transaction.rs
```diff
@@ -1567,8 +1658,9 @@ mod tests {
         let rng = &mut StdRng::seed_from_u64(2322u64);
         let secret = SecretKey::random(rng);
         let any_asset = rng.gen();
-        let arb_gas_price = 1;
+        let zero_gas_limit = 0;
         let tx = TransactionBuilder::script(vec![], vec![])
+            .max_fee_limit(zero_gas_limit)
```
xgreenx, line 1663:
```suggestion
        let tx = TransactionBuilder::script(vec![], vec![])
```

fuel-vm/src/error.rs
```diff
@@ -38,6 +39,9 @@ pub enum InterpreterError<StorageError> {
     /// The provided transaction isn't valid.
     #[display(fmt = "Failed to validate the transaction: {_0}")]
     TransactionValidity(ValidityError),
+    /// Failed while checking the transaction.
+    #[display(fmt = "Failed to check the transaction: {_0:?}")]
+    CheckError(CheckError),
```
xgreenx, line 44: Hmm, `CheckError` already contains `ValidityError` inside. Maybe it makes sense to move a new error to `ValidityError`. In this case, you don't need to add `CheckError` here

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -673,7 +676,7 @@ where
     /// that can be referenced from the interpreter instance itself.
     pub fn transact(
```
xgreenx, line 677: Maybe you want to update `deploy` method as well

fuel-vm/src/interpreter/initialization.rs
```diff
@@ -102,13 +102,13 @@ where
...
         &mut self,
-        checked: Checked<Tx>,
+        immutable: Ready<Tx>,
     ) -> Result<(), InterpreterError<S::DataError>> {
         let block_height = self.storage.block_height().map_err(RuntimeError::Storage)?;
 
         self.context = Context::Script { block_height };
 
-        let (tx, metadata): (Tx, Tx::Metadata) = checked.into();
+        let (_, tx, metadata, _) = immutable.decompose();
```
xgreenx, line 111: You need to check that `gas_price` from `Read` is the same like in `Interpreter`. The same for the `deploy` method

fuel-vm/src/error.rs
MitchTurner: It would make more sense to remove the `ValidityError` variant because there is a `ValidityError` variant of `CheckError`.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Breaking
 
+- [#685](https://github.com/FuelLabs/fuel-vm/pull/685):
+  Use the `MaxFee` policy to check that the transaction is valid. Introduce a new type `Ready` that can 
+  will be constructed with the `gas_price` before being transacted by the `Interpreter`.
```
Voxelot, line 19: - can will be

```suggestion
   be constructed with the `gas_price` before being transacted by the `Interpreter`. 
```

fuel-vm/src/error.rs
MitchTurner: Okay. I reread what you said.

Maybe `into_ready` could return a `ValidityError`? Is that what you're saying. I had to add this because we call `into_ready` inside the `Transactor`. TBH, I don't know why we are cluttering the `InterpreterError`s with a test harnesses stuff, like `Transactor`.

I'd almost say that `Transactor` should have its own error that can wrap `CheckError` and `InterpreterError`. The whole caught error thing in `Transactor` is kinda ugly IMO, but it's kinda out of scope right now.

fuel-vm/src/interpreter/initialization.rs
MitchTurner: I think I addressed this. But maybe I misunderstood.

xgreenx (approved):

PR #688, opened 2024-02-29
feat: Return `Arc` in `StorageRead` and `StorageWrite` methods

bvrooman:


Voxelot:
@xgreenx this seems like a strange modification to me, why are we wrapping everything from storage into arcs?

xgreenx:
Yeah, maybe it is overkill=D Let's close for now

PR #696, opened 2024-03-11
dev: refactor smt node visibility

bvrooman:


fuel-merkle/src/sparse/merkle_tree/merkle_tree.rs
```diff
@@ -0,0 +1,1676 @@
+pub(super) use super::{
```
xgreenx, line 1: You don't need to move the actual `MerkleTree` into a sub-module. It can live in the same place as before, but `node` and `branch` will be submodels of it.

PR #697, opened 2024-03-12
Store stack and heap separately

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/452

Spec PR: https://github.com/FuelLabs/fuel-specs/pull/566


fuel-vm/src/interpreter/memory.rs
```diff
@@ -19,13 +19,231 @@ use fuel_types::{
...
+        let len = self.len();
+
+        if new_len > len {
+            // It is how `Vec::resize` calculates a new capacity.
+            let cap = core::cmp::max(len * 2, new_len);
+            let cap = core::cmp::min(cap, MEM_SIZE);
+            let diff = cap - len;
+            let mut new_vec = vec![value; cap];
+            new_vec[diff..].copy_from_slice(self.as_slice());
+            *self = new_vec;
```
Dentosal, line 58: This at least just copies the bytes just once, but it also forces a new allocation every time. I guess that cannot be avoided if we want to use a single `Vec`. We could maybe be a bit smarter than just doubling the size every time, e.g. starting with a kilobyte-sized allocation and not allowing the last doubling to `MEM_SIZE`, instead allocating with `reserve_exact`.

In any case, `ALOC` is currently constant cost and will allow copying the amount of entire memory, in parts. What used to be a cheap subtract and compare operation, is now a memcpy.

Dentosal:
There's a subtle bug in the implementation here: It's possible to extend the stack, then shrink the stack, and then extend the heap to be where the stack once was. Then reading from that memory area would read from the memory that was in the stack, and not from the heap that's on that region now. Fortunately we never shrink the heap, so just truncating the `stack´ field every time we grow heap over it fixes that.

fuel-vm/src/constraints.rs
```diff
@@ -101,16 +13,16 @@ pub struct InstructionLocation {
     pub offset: u64,
 }
 
-impl<const LEN: usize> Deref for CheckedMemConstLen<LEN> {
-    type Target = MemoryRange;
+// impl<const LEN: usize> Deref for CheckedMemConstLen<LEN> {
```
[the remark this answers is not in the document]
Dentosal: 3f082f4e

fuel-vm/src/interpreter/memory.rs
```diff
@@ -19,23 +21,322 @@ use fuel_types::{
...
+    new_vec.extend(core::iter::repeat(0).take(cap - vec.len()));
+    new_vec.extend(vec.iter().copied());
+    *vec = new_vec;
+}
+
+/// The memory of the VM, represented as stack and heap.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Memory {
+    /// Stack. Grows upwards.
+    stack: Vec<u8>,
```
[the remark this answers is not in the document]
Dentosal: 3f082f4e

fuel-vm/src/interpreter/blockchain/test/scwq.rs
```diff
@@ -98,39 +106,11 @@ fn test_state_clear_qword(
...
-#[test_case(
-    VM_MAX_RAM-1, 1
-    => matches Err(_)
-    ; "Fail when $rA + 32 > VM_MAX_RAM"
-)]
-#[test_case(
-    VM_MAX_RAM-32, 1
-    => matches Ok(_)
-    ; "Pass when $rA + 32 == VM_MAX_RAM"
-)]
```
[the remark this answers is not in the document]
Dentosal: Re-added in b054694c.

fuel-vm/src/interpreter/blockchain/test/swwq.rs
```diff
@@ -131,66 +239,11 @@ fn test_state_write_qword(
...
-#[test_case(
-    VM_MAX_RAM - 1, 0, 1
-    => matches Err(_)
-    ; "Fail when rA + 32 > VM_MAX_RAM"
-)]
-#[test_case(
-    VM_MAX_RAM - 32, 0, 1
-    => matches Ok(_)
-    ; "Pass when rA + 32 == VM_MAX_RAM"
-)]
```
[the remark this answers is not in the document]
Dentosal: Re-added in b054694c.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -19,23 +22,324 @@ use fuel_types::{
...
+            .saturating_sub(self.heap.len());
+        let uninit_memory = vec![0u8; uninit_memory_size];
+        let mut memory = self.stack;
+        memory.extend(uninit_memory);
+        memory.extend(self.heap);
+        memory
+    }
+
+    /// Grows the stack to be at least `new_sp` bytes.
+    pub fn grow_stack(&mut self, hp: Reg<HP>, new_sp: Word) -> Result<(), PanicReason> {
```
xgreenx, line 120: Hmm, it seems we don't need to pass `Reg<HP>` since we have `self.hp` that represents the same. I know that you are using it in the tests to manipulate the heap, but maybe tests shouldn't affect the method API.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -19,23 +22,324 @@ use fuel_types::{
...
+        count: B,
+    ) -> Result<MemoryRange, PanicReason> {
+        let start = addr.to_addr()?;
+        let len = count.to_addr()?;
+        let end = start.saturating_add(len);
+        if end > MEM_SIZE {
+            return Err(PanicReason::MemoryOverflow)
+        }
+
+        if end <= self.stack.len() || (start >= self.hp && start >= self.stack.len()) {
```
xgreenx, line 167: I'm just curious why we need `start >= self.stack.len()` if `hp >= sp`? For the cases where `sp < self.stack.len()`? If yes, then it is a bug in the memory implementation(on our side), and we need to return something else than `UninitalizedMemoryAccess`

fuel-vm/src/interpreter/memory.rs
```diff
@@ -73,96 +377,55 @@ impl ToAddr for i32 {
...
     }
 
-    /// Convert to a raw `usize` range.
-    pub fn usizes(&self) -> ops::Range<usize> {
-        self.start..self.end
-    }
+    /// Moves range to be relative to the given address.
+    /// Panics if the address is before the offset.
+    pub fn relative_to(self, index: usize) -> Self {
+        assert!(self.start() >= index);
```
xgreenx, line 424: We use this function to places where we verify that `start >= index`, so it shouldn't fail. But since this function is public, I would prefer to remove this function and use:

```rust
let start = self.start() - index;
let end = self.end() - index;
```

Directly in the place of this function usage=)

fuel-vm/src/interpreter/memory.rs
```diff
@@ -334,11 +575,15 @@ pub(crate) fn try_update_stack_pointer(
     ssp: Reg<SSP>,
     hp: Reg<HP>,
     new_sp: Word,
+    memory: &mut Memory,
 ) -> SimpleResult<()> {
-    if new_sp >= *hp || new_sp < *ssp {
+    if new_sp < *ssp {
         Err(PanicReason::MemoryOverflow.into())
+    } else if new_sp > *hp {
```
xgreenx, line 582: So before, we had a check `new_sp >= *hp`, while now it is  `new_sp >*hp`. Is it expected? Did I get right that stack is `[0..SP)` and the heap is `[HP..MEM_SIZE)` and it is okay that `SP == HP`?

fuel-vm/src/interpreter/memory.rs
```diff
@@ -379,11 +625,14 @@ pub(crate) fn push_selected_registers(
 
     // First update the new stack pointer, as that's the only error condition
     let count: u64 = bitmask.count_ones().into();
-    let stack_range = MemoryRange::new(*sp, count * 8)?;
-    try_update_stack_pointer(sp, ssp, hp, stack_range.words().end)?;
+    let write_size = count * (core::mem::size_of::<Word>() as u64);
+    let write_at = *sp;
+    try_update_stack_pointer(sp, ssp, hp, write_at + write_size, memory)?;
```
xgreenx, line 630: I know that it shouldn't be possible to overflow here, but it would be nice to add `#![deny(clippy::arithmetic_side_effects)]` for `fuel-vm`. But it is definitely not a part of the current PR=)

fuel-vm/src/interpreter/memory.rs
Dentosal: True. I find it quite distasteful to even duplicate `$hp` inside this struct, but now that we do it, why not use it for this as well. Will fix.

fuel-vm/src/interpreter/memory.rs
Dentosal: Good catch. I added the check before I realized that we have to truncate the stack if heap extends on top of it. WIll remove.

fuel-vm/src/interpreter/memory.rs
Dentosal: I prefer having the function, but we can make it private.

fuel-vm/src/interpreter/memory.rs
Dentosal: Exactly correct.

fuel-vm/src/interpreter/memory.rs
Dentosal: Agreed. It's definitely not too easy to see that this cannot overflow.

fuel-vm/src/interpreter/memory.rs
Dentosal: Done in 3baee80c

fuel-vm/src/interpreter/memory.rs
Dentosal: Removed in e8f5b451

fuel-vm/src/interpreter/memory.rs
Dentosal: Ah nevermind, just removed/inlined the function in ad312a8b

fuel-vm/src/interpreter/flow.rs
```diff
@@ -627,39 +616,47 @@ where
...
+    let content_size = len.saturating_sub(frame_len_with_padding);
+    memory
+        .write_noownerchecks(*fp, content_size)?
         .copy_from_slice(&frame_bytes);
 
-    let mut code_range = code_mem_range.clone();
-    code_range.grow_start(CallFrame::serialized_size());
-    code_range.shrink_end(frame.code_size_padding());
+    let code_start = fp.saturating_add(CallFrame::serialized_size() as Word);
+    let code_len = frame.code_size();
```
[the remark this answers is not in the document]
Dentosal: The old version constructed`code_range` essentially like this

```rust
code_frame_range.shrink_end(frame.code_size() + frame.code_size_padding());
let mut code_range = code_mem_range.clone();
code_range.grow_start(CallFrame::serialized_size());
code_range.shrink_end(frame.code_size_padding());
```

Note the `shrink_end` which is used to remove the padding. So here we never add the padding, instead of subtracting it away after adding it.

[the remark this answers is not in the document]
xgreenx: Oh, I see=D

xgreenx (commented):
Looks super good to me=) Approved! (I can't approve since I created the draft PR)

Dentosal (approved):

PR #700, opened 2024-03-20
Add GM for base asset ID

Dentosal:
Spec PR: https://github.com/FuelLabs/fuel-specs/pull/561

### DIFF
--- a/fuel-tx/src/transaction/consensus_parameters.rs
+++ b/fuel-tx/src/transaction/consensus_parameters.rs
@@ -251,9 +251,10 @@ impl TxParameters {
     /// Transaction memory offset in VM runtime
     pub const fn tx_offset(&self) -> usize {
         Bytes32::LEN // Tx ID
-            + WORD_SIZE // Tx size
-            // Asset ID/Balance coin input pairs
-            + self.max_inputs as usize * (AssetId::LEN + WORD_SIZE)
+        + AssetId::LEN // Base asset ID
+        // Asset ID/Balance coin input pairs
+        + self.max_inputs as usize * (AssetId::LEN + WORD_SIZE)
+        + WORD_SIZE // Tx size
     }
 
     /// Replace the max inputs with the given argument
--- a/fuel-vm/src/interpreter/metadata.rs
+++ b/fuel-vm/src/interpreter/metadata.rs
@@ -108,53 +108,37 @@ pub(crate) fn metadata(
     chain_id: ChainId,
     tx_offset: Word,
 ) -> SimpleResult<()> {
-    let external = context.is_external();
     let args = GMArgs::try_from(imm)?;
+    let parent = context
+        .is_internal()
+        .then(|| frames.last().map(|f| f.registers()[RegId::FP]))
+        .flatten();
 
-    if external {
-        match args {
-            GMArgs::GetVerifyingPredicate => {
-                *result = context
-                    .predicate()
-                    .map(|p| p.idx() as Word)
-                    .ok_or(PanicReason::TransactionValidity)?;
-            }
-
-            GMArgs::GetChainId => {
-                *result = chain_id.into();
-            }
-
-            GMArgs::TxStart => {
-                *result = tx_offset;
-            }
-
-            _ => return Err(PanicReason::ExpectedInternalContext.into()),
+    match (args, parent) {
+        (GMArgs::GetVerifyingPredicate, None) => {
+            *result = context
+                .predicate()
+                .map(|p| p.idx() as Word)
+                .ok_or(PanicReason::TransactionValidity)?;
         }
-    } else {
-        let parent = frames
-            .last()
-            .map(|f| f.registers()[RegId::FP])
-            .expect("External context will always have a frame");
 
-        match args {
-            GMArgs::IsCallerExternal => {
-                *result = (parent == 0) as Word;
-            }
-
-            GMArgs::GetCaller if parent != 0 => {
-                *result = parent;
-            }
-
-            GMArgs::GetChainId => {
-                *result = chain_id.into();
-            }
-
-            GMArgs::TxStart => {
-                *result = tx_offset;
-            }
-
-            _ => return Err(PanicReason::ExpectedInternalContext.into()),
+        (GMArgs::GetChainId, _) => {
+            *result = chain_id.into();
         }
+
+        (GMArgs::TxStart, _) => {
+            *result = tx_offset;
+        }
+
+        (GMArgs::GetCaller, Some(parent)) if parent != 0 => {
+            *result = parent;
+        }
+
+        (GMArgs::IsCallerExternal, Some(parent)) => {
+            *result = (parent == 0) as Word;
+        }
+
+        _ => return Err(PanicReason::ExpectedInternalContext.into()),
     }
 
     inc_pc(pc)?;
--- a/fuel-vm/src/interpreter/metadata.rs
+++ b/fuel-vm/src/interpreter/metadata.rs
@@ -126,6 +126,10 @@ pub(crate) fn metadata(
             *result = chain_id.into();
         }
 
+        (GMArgs::BsaeAssetId, _) => {
+            *result = tx_offset;
+        }
+
         (GMArgs::TxStart, _) => {
             *result = tx_offset;
         }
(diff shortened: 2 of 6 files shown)


fuel-vm/src/interpreter/metadata.rs
```diff
@@ -108,53 +108,41 @@ pub(crate) fn metadata(
...
-                *result = (parent == 0) as Word;
-            }
+        (GMArgs::GetChainId, _) => {
+            *result = chain_id.into();
+        }
 
-            GMArgs::GetCaller if parent != 0 => {
-                *result = parent;
-            }
+        (GMArgs::BsaeAssetId, _) => {
```
Voxelot, line 129: `BsaeAssetId` ??

```suggestion
        (GMArgs::BsaeAssetId, _) => {
```

fuel-vm/src/interpreter/metadata.rs
```diff
@@ -108,53 +108,41 @@ pub(crate) fn metadata(
...
+        (GMArgs::GetCaller, Some(parent)) if parent != 0 => {
+            *result = parent;
+        }
 
-            _ => return Err(PanicReason::ExpectedInternalContext.into()),
+        (GMArgs::IsCallerExternal, Some(parent)) => {
+            *result = (parent == 0) as Word;
         }
+
+        _ => return Err(PanicReason::ExpectedInternalContext.into()),
```
Voxelot, line 145: This doesn't seem very future-proof, as the context error as a catch-all could happen even when it's an unrelated error.

Dentosal: Even before, it could actually occur during both contexts, and definitely doesn't seem to be the correct error. I'll rework this part.

fuel-vm/src/interpreter/metadata.rs
Dentosal: Oops, fixed in b78c16c7

fuel-vm/src/interpreter/metadata.rs
Dentosal: Fixed in 3fe06ca. I changed the match to an exhaustive one and added a new panic variant `ExpectedNestedCaller` for the relevant case.

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -251,9 +251,10 @@ impl TxParameters {
     /// Transaction memory offset in VM runtime
     pub const fn tx_offset(&self) -> usize {
         Bytes32::LEN // Tx ID
-            + WORD_SIZE // Tx size
-            // Asset ID/Balance coin input pairs
-            + self.max_inputs as usize * (AssetId::LEN + WORD_SIZE)
```
MitchTurner, line 256: Why flip the order of these values?

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -251,9 +251,10 @@ impl TxParameters {
...
+        + AssetId::LEN // Base asset ID
+        // Asset ID/Balance coin input pairs
+        + self.max_inputs as usize * (AssetId::LEN + WORD_SIZE)
+        + WORD_SIZE // Tx size
```
MitchTurner, line 257: Why is the `Tx size` spot flipped from before?

fuel-vm/src/tests/metadata.rs
```diff
@@ -268,6 +268,53 @@ fn get_metadata_chain_id() {
     }
 }
 
+#[test]
+fn get_metadata_base_asset_id() {
```
MitchTurner, line 272: So, I'd say that the SUT of this test is `op::gm_args` and the condition is `GMArgs::BaseAssetId`. It's confusing since we can't call `gm_args` natively and instead we call `transact`, so you could also say that `transact` is the SUT in that sense.

Either way, I think the structure of the test should reflect what the SUT is. In this case I include it in a `gm_args_script` method, so it's a little indirect but cleans up the test in a good way IMO:

```
// given
let base_asset_id = AssetId::from([5; 32]);
let params = ConsensusParameters {
        base_asset_id,
        ..Default::default()
    };

// when
let arg = GMArgs::BaseAssetId;
let size = AssetId::LEN;
let script = gm_args_script(arg, size);

let logged_data = run_script_and_get_logged_data(params, script);

// then
let expected = params.base_asset_id.to_bytes();
let actual = logged_data;
assert!(expected, actual);
```

So the name of the test could be something like `gm_args__base_asset_id_retrieves_value_in_params`

fuel-tx/src/transaction/consensus_parameters.rs
Dentosal: It's after the balance pairs in memory, so I made the order here match the order in memory.

fuel-vm/src/interpreter/metadata.rs
```diff
@@ -108,54 +108,29 @@ pub(crate) fn metadata(
...
-
-            GMArgs::TxStart => {
-                *result = tx_offset;
-            }
-
-            _ => return Err(PanicReason::ExpectedInternalContext.into()),
-        }
-    }
+    *result = match GMArgs::try_from(imm)? {
+        GMArgs::GetVerifyingPredicate => context
```
xgreenx, line 117: Before we were returning an `ExpectedInternalContext` error in the case of the internal context. Do we need to do the same?

fuel-vm/src/tests/validation.rs
```diff
@@ -62,9 +60,8 @@ fn malleable_fields_do_not_affect_validity() {
 
     let params = ConsensusParameters::default();
 
-    let tx_size_ptr =
-        32 + (params.tx_params.max_inputs as usize * (AssetId::LEN + WORD_SIZE));
-    let tx_start_ptr = tx_size_ptr + 8;
+    let tx_start_ptr = params.tx_params.tx_offset();
+    let tx_size_ptr = tx_start_ptr - 8;
```
xgreenx, line 64: Why do we need to do `-8`?

Voxelot:
looks like there are merge conflicts now

fuel-tx/src/transaction/consensus_parameters.rs
MitchTurner: Ah. It looks like it wasn't matching the specs before but it is now AFAICT.

fuel-vm/src/interpreter/metadata.rs
Dentosal: Good catch. The new behavior is what the specs say. I see no reason to disallow internal contexts here.

fuel-vm/src/tests/metadata.rs
Dentosal: I think doing that would make the test case less readable. Note how this is already an integration test, not a unit test, and it still has minimal indirection which is quite nice. I experimented a bit with something like `run_script_and_get_logged_data` and going that way seems to lead essentially duplicating `TransactionBuilder` if we want to make it reusable.

fuel-vm/src/interpreter/metadata.rs
xgreenx: The specification says that it is allowed only in the context of the predicate:

image

And in our codebase predicate context is external context:

image

So it looks like we need ti check that it is external context=)

xgreenx (changes requested):

fuel-vm/src/interpreter/metadata.rs
Dentosal: But the code already performs the check: The `context.predicate()` returns `None` outside predicate contexts, leading to `TransactionValidity` panic. Previously calling `gm/GetVerifyingPredicate` panicked with `ExpectedInternalContext` which was definitely incorrect. Now if we check with something like
```rust
if context.is_external() { // pseudocode
  result = context.predicate().map(...).ok_or(PanicReason::TransactionValidity)?;
} else {
  return PanicReason::ExpectedExternalContext;
}
```
then the only additional check we're actually performing is that `Call` contexts return `ExpectedExternalContext` instead of `TransactionValidity`. Is the issue just returning a wrong panic reason? Am I missing something here?

xgreenx: Ohh, I missed the moment that `predicate` already is doing this check inside=D Hmm, maybe using of another error will make more sense, because `TransactionValidity` is a strange choice

xgreenx (approved):

fuel-vm/src/interpreter/metadata.rs
Dentosal: I'm going to merge this as-is, since `TransactionValidity` is (only) used for this kind of things currently. We can do a rename in a separate PR if needed.

PR #702, opened 2024-03-21
Versioning of sub parameters

xgreenx:
Wrapped `FeeParameters`, `PredicateParameters`, `TxParameters`, `ScriptParameters` and `ContractParameters` into an enum to support versioning.


xgreenx (commented):
Ignore JS files please, it is auto formatter=)

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -320,50 +467,156 @@ pub struct PredicateParameters {
     pub max_gas_per_predicate: u64,
 }
 
-impl PredicateParameters {
-    /// Default consensus parameters with settings suggested in fuel-specs
+impl PredicateParametersV1 {
+    /// Default parameters just for testing.
```
Dentosal, line 471: If these are just for testing, shouldn't we feature-gate them?

xgreenx: I will put them under `test-helpers` in a separate PR

Dentosal (approved):
This is ok as-is, but needs follow-up PR as discussed above.

PR #706, opened 2024-03-24
Unified `Create` and `Script` logic via `ChargeableTransaction`

xgreenx:
Created `ChargeableTransaction` that accumulates common logic of the `Create` and `Script` transactions. The same logic can be applied to any chargeable transaction. The `Mint` transaction is free and created by the block producer, so it is not chargeable and doesn't have policies, inputs, outputs, and witnesses.

Side changes:
- Updated the canonical serialization and deserialization logic to ignore fields marked with `canonical(skip)`. We already partially supported that, but if the skipped field used a generic, we required that generic to implement the corresponding trait. After the change, skipped fields don't impact the generic's constraints.

The change is breaking because affects JSON serialization and deserialization. Now `Script` and `Create` transactions have `body` fields that include unique transactions.


fuel-tx/src/transaction/types/chargeable_transaction.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub common: CommonMetadata,
+    pub body: Body,
+}
+
+#[derive(Clone, Derivative)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
+#[derivative(Eq, PartialEq, Hash, Debug)]
+pub struct ChargeableTransaction<Body, MetadataBody> {
+    pub(crate) body: Body,
```
xgreenx, line 50: The change is breaking because now old fields live under the `body` subfield. We could avoid it by flattening it, but it is [not supported](https://github.com/bincode-org/bincode/issues/245) properly by serde for the bincode.

xgreenx:
Hmm, basically no difference=D We also could use the name `ExecutableTransasction`.

I decided to call it "chargeable" because any executable transaction is chargeable by default, but not any chargeable transaction is executable. Executable transactions mean that they support predicate execution as well.

Dentosal:
I mean there's already an `ExecutableTransaction` trait:

https://github.com/FuelLabs/fuel-vm/blob/d5824ddd138cf13f08e0d61b1f2c237bef3a7448/fuel-vm/src/interpreter.rs#L359

This is only implemented for `Script` and `Create`. So I was confused about the naming. But apparently `ChargeableTransaction` is a concrete type containing the common fields instead. The naming makes sense, but it's still a bit confusing.

xgreenx:
[This](https://github.com/FuelLabs/fuel-vm/pull/707) PR `Upgrade` transaction that also implements the `ExecutableTransaction` trait. The idea is that `ExecutableTransaction` is defined by the `Interpreter`, and it is requirements from the VM to the transaction. While `ChargeableTransaction` is defined in the `fuel-tx` without any execution context.

fuel-tx/src/transaction/id.rs
```diff
@@ -21,6 +21,12 @@ use fuel_types::{
     ChainId,
 };
 
+/// Prepares transaction for signing.
+pub trait PrepareSign {
+    /// Prepares transaction for signing zeroing required fields.
```
Dentosal, line 26:
```suggestion
    /// Prepares transaction for signing, zeroing required fields.
```

fuel-tx/src/transaction/types/mint.rs
```diff
@@ -51,7 +51,6 @@ impl MintMetadata {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(fuel_types::canonical::Deserialize, fuel_types::canonical::Serialize)]
 #[canonical(prefix = TransactionRepr::Mint)]
-#[cfg_attr(feature = "typescript", wasm_bindgen::prelude::wasm_bindgen)]
```
Dentosal, line 54: Why is this removed?

xgreenx: The `wasm_bindgen` doesn't know how to work with generics, so it is not possible to inherit for `Create` and `Script` transactions. So, I reworked how we are passing any transaction to typescript.

Dentosal (approved):

Dentosal:
Now has conflicts, but looks correct otherwise.

Dentosal (approved):

PR #707, opened 2024-03-25
A new `Upgrade` transaction to perform network upgrades

xgreenx:
Related https://github.com/FuelLabs/fuel-core/issues/1753 and https://github.com/FuelLabs/fuel-specs/pull/567

The change adds a new `Upgrade` transaction that allows upgrading either consensus parameters or state transition function used by the network to produce future blocks.

The purpose of the upgrade is defined by the `Upgrade Purpose` type:

```rust
pub enum UpgradePurpose {
    /// The upgrade is performed to change the consensus parameters.
    ConsensusParameters {
        /// The index of the witness in the [`Witnesses`] field that contains
        /// the serialized consensus parameters.
        witness_index: u16,
        /// The hash of the serialized consensus parameters.
        /// Since the serialized consensus parameters live inside witnesses(malleable
        /// data), any party can override them. The `checksum` is used to verify that the
        /// data was not modified.
        checksum: Bytes32,
    },
    /// The upgrade is performed to change the state transition function.
    StateTransition {
        /// The hash of the new bytecode of the state transition function.
        /// The bytecode must be present on the blockchain(should be known by the
        /// network) at the moment of inclusion of this transaction.
        bytecode_hash: Bytes32,
    },
}
```

The `Upgrade` transaction is chargeable, and the sender should pay for it. Transaction inputs should contain only base assets.

Only the privileged address can upgrade the network. The privileged address can be either a real account or a predicate.

Since serialized consensus parameters are small(< 2kb), they can be part of the upgrade transaction and live inside of witness data. The bytecode of the blockchain state transition function is huge ~1.6MB(relative to consensus parameters), and it is impossible to fit it into one transaction. So when we perform the upgrade of the state transition function, it should already be available on the blockchain. The transaction to actually upload the bytecode(`Upload` transaction) will implemented in the https://github.com/FuelLabs/fuel-core/issues/1754.

Side changes:
- Moved `fuel-tx-test-helpers` logic into the `fuel_tx::test_helpers` module.
- The combination of `serde` and `postcard` is used to serialize and deserialize `ConsensusParameters` during the upgrade. This means the protocol and state transition function requires the `serde` feature by default for `ConsensusParameters` and `fuel-types`.
- Added a new rule for `Create` transaction: all inputs should use base asset otherwise it returns `TransactionInputContainsNonBaseAssetId` error.
- Renamed some errors because now they are used for several transactions(`Upgrade` uses some errors from `Create` and some from `Script` transactions):
  - `TransactionScriptOutputContractCreated` -> `TransactionOutputContainsContractCreated`.
  - `TransactionCreateOutputContract` -> `TransactionOutputContainsContract`.
  - `TransactionCreateOutputVariable` -> `TransactionOutputContainsVariable`.
  - `TransactionCreateOutputChangeNotBaseAsset` -> `TransactionChangeChangeUsesNotBaseAsset`.
  - `TransactionCreateInputContract` -> `TransactionInputContainsContract`.
  - `TransactionCreateMessageData` -> `TransactionInputContainsMessageData`.
- Used the same pattern everywhere: 
  ```rust
               Self::Script(tx) => tx.encode_static(buffer),
               Self::Create(tx) => tx.encode_static(buffer),
               Self::Mint(tx) => tx.encode_static(buffer),
               Self::Upgrade(tx) => tx.encode_static(buffer),
  ```
  
  Instead of:
  ```rust
               Transaction::Script(script) => script.encode_static(buffer),
               Transaction::Create(create) => create.encode_static(buffer),
               Transaction::Mint(mint) => mint.encode_static(buffer),
               Transaction::Upgrade(upgrade) => upgrade.encode_static(buffer),
  ```


fuel-tx/src/transaction/types/upgrade.rs
```diff
@@ -0,0 +1,315 @@
...
+pub type Upgrade = ChargeableTransaction<UpgradeBody, UpgradeMetadata>;
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum UpgradeMetadata {
+    ConsensusParameters {
+        /// Deserialized consensus parameters from the witness.
+        consensus_parameters: Box<ConsensusParameters>,
+        /// The actual checksum of the serialized consensus parameters.
+        calculated_checksum: Bytes32,
+    },
```
Dentosal, line 42: These are just cached from the witness?

fuel-tx/src/transaction/types/upgrade.rs
```diff
@@ -0,0 +1,315 @@
...
+        let tx_id_gas = gas_cost.s256().resolve(bytes as u64);
+
+        let purpose_gas = match &self.body.purpose {
+            UpgradePurpose::ConsensusParameters { witness_index, .. } => {
+                let len = self
+                    .witnesses
+                    .get(*witness_index as usize)
+                    .map_or(0, |w| w.as_vec().len());
+                gas_cost.s256().resolve(len as u64)
+            }
```
Dentosal, line 164: I'm not sure why we charge for this. Having a spec update PR would help.

xgreenx: We charge for checksum calculation. I will update the specification with the `Upgrade` transaction soon

xgreenx: https://github.com/FuelLabs/fuel-specs/pull/567

fuel-tx/src/transaction.rs
```diff
@@ -190,6 +194,58 @@ impl Transaction {
...
+        consensus_parameters: &ConsensusParameters,
+        policies: Policies,
+        inputs: Vec<Input>,
+        outputs: Vec<Output>,
+        mut witnesses: Vec<Witness>,
+    ) -> Result<Upgrade, ValidityError> {
+        let serialized_consensus_parameters = postcard::to_allocvec(consensus_parameters)
+            .map_err(|_| {
+                ValidityError::TransactionUpgradeConsensusParametersSerialization
+            })?;
```
Dentosal, line 228: nit: I'm pretty sure this is never supposed to fail

Dentosal (dismissed):

Dentosal (dismissed):

fuel-tx/src/tests/valid_cases/transaction/upgrade.rs
```diff
@@ -0,0 +1,729 @@
...
+    };
+
+    // When
+    let result = tx.check(block_height, &test_params());
+
+    // Then
+    assert_eq!(Err(ValidityError::TransactionMetadataMismatch), result);
+}
+
+mod inputs {
```
MitchTurner, line 616: These tests are missing the `check` prefix.

Also, why are these in a separate mod?

fuel-tx/src/transaction/id.rs
```diff
@@ -682,6 +686,15 @@ mod tests {
         let scripts = vec![vec![], generate_bytes(rng), generate_bytes(rng)];
         let script_data = vec![vec![], generate_bytes(rng), generate_bytes(rng)];
         let storage_slots = vec![vec![], vec![rng.gen(), rng.gen()]];
+        let purposes = [
```
MitchTurner, line 689: This test is giant. We should probably split each `Vec` into its own little helper...? 


We should also probably come up with a better name for this test, `id` isn't really a behavior. `id__changing_fields_in_the_tx_changes_the_generated_id`? or something like that?

_You don't need to do this, but we wouldn't need all the nested `for` loops if we just used prop test strategies to generate the groups. But that would be kinda a lot of work to set up, so I think this is fine for now_

fuel-tx/src/transaction/types/upgrade.rs
```diff
@@ -0,0 +1,317 @@
...
+        /// The actual checksum of the serialized consensus parameters.
+        calculated_checksum: Bytes32,
+    },
+    /// Currently there is no metadata for state transition upgrades, so leave it empty.
+    #[default]
+    StateTransition,
+}
+
+impl UpgradeMetadata {
+    pub fn compute(tx: &Upgrade) -> Result<Self, ValidityError> {
```
MitchTurner, line 50: If the `Upgrade` (`ChargeableTransaction`) already has `Some` metadata, would it make sense to use the "cached" value ever?

fuel-vm/src/tests/encoding.rs
```diff
@@ -395,4 +410,114 @@ fn transaction() {
             vec![],
         ),
     ]);
+    assert_transactions_encoding_correct(&[
```
MitchTurner, line 413: It's kinda hard to see what is different about each tx or why there are the number there are. Maybe move to a helper that says what the set is? I can't say how you chose them, but whatever the process was can be the name of the helper.

fuel-tx/src/transaction/types/upgrade.rs
xgreenx: No, this function calculates new metadata from the `Upgrade` transaction. We shouldn't use old metadata since it may not be actual anymore because of changes that we made.

fuel-tx/src/tests/offset.rs
```diff
@@ -432,6 +434,57 @@ fn tx_offset_script() {
     assert!(cases.output_contract_created_id);
 }
 
+#[test]
+fn tx_offset_upgrade() {
```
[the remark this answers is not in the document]
xgreenx: Added a comment in the test file to highlight the use case.

image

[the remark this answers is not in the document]
xgreenx: We are testing public getters to the offset of each field

fuel-tx/src/tests/valid_cases/transaction/upgrade.rs
xgreenx: It is a separate module since all tests below used different input types but testing the same thing, while above we are testing validity rules  from specification.

fuel-tx/src/tests/valid_cases/transaction/upgrade.rs
```diff
@@ -0,0 +1,729 @@
...
+
+    // When
+    let result = tx.check(block_height, &test_params());
+
+    // Then
+    assert_eq!(Err(ValidityError::TransactionMaxFeeNotSet), result);
+}
+
+#[test]
+fn max_iow_reached_max_inputs() {
```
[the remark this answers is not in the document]
xgreenx: Input/Output/Witness=) Removed it

fuel-vm/src/tests/encoding.rs
xgreenx: https://github.com/FuelLabs/fuel-vm/issues/717

fuel-tx/src/tests/bytes.rs
```diff
@@ -636,3 +746,94 @@ fn script_input_coin_data_offset() {
         }
     }
 }
+
+#[test]
+fn upgrade_input_coin_data_offset() {
```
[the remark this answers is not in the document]
xgreenx: https://github.com/FuelLabs/fuel-vm/issues/717

fuel-tx/src/transaction/id.rs
xgreenx: https://github.com/FuelLabs/fuel-vm/issues/717

MitchTurner (approved):

PR #709, opened 2024-03-26
Removed `bytecode_length` from the `Create` transaction

xgreenx:
Spec: https://github.com/FuelLabs/fuel-specs/pull/563


Dentosal (dismissed):
I was always wondering why we have this. No more.

bvrooman (approved):

Qinlan7:
> I was always wondering why we have this. No more.

Please add me

PR #712, opened 2024-04-02
Process `Upgrade` transaction inside of the `Interpreter`

xgreenx:
The `Interpreter` supports the processing of the `Upgrade` transaction. The change affects `InterpreterStorage`, adding 5 new methods that must be implemented.


fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -466,6 +473,113 @@ where
...
+        gas_costs: &GasCosts,
+        fee_params: &FeeParameters,
+        base_asset_id: &AssetId,
+        gas_price: Word,
+    ) -> Result<(), InterpreterError<S::DataError>> {
+        let metadata = upgrade.metadata().as_ref();
+        debug_assert!(
+            metadata.is_some(),
+            "`upgrade_inner` is called without cached metadata"
+        );
```
Dentosal, line 493: Can't we compute the metadata here?

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -466,6 +473,113 @@ where
...
+                    let metadata = UpgradeMetadata::compute(upgrade)?;
+                    Self::get_consensus_parameters(&metadata)?
+                };
+
+                let current_version = storage
+                    .consensus_parameters_version()
+                    .map_err(RuntimeError::Storage)?;
+                let next_version = current_version
+                    .checked_add(1)
+                    .ok_or(InterpreterError::Panic(PanicReason::ArithmeticOverflow))?;
```
Dentosal, line 509: Introducing a new `PanicReason` for this would be nice. `ArithmeticOverflow` is meant to be used for arithmetic opcodes only.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -466,6 +473,113 @@ where
...
+                        PanicReason::UnknownStateTransactionBytecodeHash,
+                    ))
+                }
+
+                let current_version = storage
+                    .state_transition_version()
+                    .map_err(RuntimeError::Storage)?;
+                let next_version = current_version
+                    .checked_add(1)
+                    .ok_or(InterpreterError::Panic(PanicReason::ArithmeticOverflow))?;
```
Dentosal, line 537: As above

fuel-vm/src/memory_client.rs
```diff
@@ -83,6 +84,11 @@ impl<Ecal: EcalHandler> MemoryClient<Ecal> {
         self.transactor.deploy(tx).ok()
     }
 
+    /// Executes `Upgrade` transaction.
+    pub fn upgrade(&mut self, tx: Checked<Upgrade>) -> Option<Upgrade> {
```
Dentosal, line 88: Why are we erasing the error type here?

Dentosal (commented):
Needs some clarifications

fuel-vm/src/interpreter/executors/main.rs
xgreenx: We can, but it was calculated before in the `Checked` transaction. It is just a debug assert for the future if someone decides to change the code or behavior later.

Recalculation of it breaks the idea of the metadata=D

fuel-asm/src/panic_reason.rs
```diff
@@ -137,6 +137,12 @@ enum_from! {
         /// Attempting to read or write uninitialized memory.
         /// Also occurs when boundary crosses from stack to heap.
         UninitalizedMemoryAccess = 0x30,
+        /// It is not allowed to override the consensus parameters.
```
MitchTurner, line 140: nit:
```suggestion
        /// Overriding consensus parameters is not allowed.
```

fuel-asm/src/panic_reason.rs
```diff
@@ -137,6 +137,12 @@ enum_from! {
...
+        OverridingConsensusParameters = 0x31,
+        /// The storage doesn't know about the hash of the state transition bytecode.
+        UnknownStateTransactionBytecodeHash = 0x32,
+        /// It is not allowed to override the state transition bytecode.
```
MitchTurner, line 144:
```suggestion
        /// Overriding the state transition bytecode is not allowed.
```

fuel-vm/src/tests/upgrade.rs
```diff
@@ -0,0 +1,418 @@
...
+        assert_eq!(
+            Err(InterpreterError::Panic(
+                PanicReason::UnknownStateTransactionBytecodeHash
+            )),
+            result
+        );
+    }
+
+    #[test]
+    fn transact_fails_when_try_to_override_state_bytecode() {
```
MitchTurner, line 196: This case is confusing me. Isn't the point of the upgrade to override some previous state? 

Other than the first time it's run, under what circumstances will this not fail?

MitchTurner: Same question for the consensus params.

MitchTurner (commented):
Made it part way through.

fuel-vm/src/tests/upgrade.rs
xgreenx: Each version contains only one associated consensus parameter or state transition function. If it was set before, it is not allowed to be overridden for the same version. You can only increase the version and set it for a new one.

fuel-vm/src/interpreter/executors/main/tests.rs
```diff
@@ -199,3 +199,51 @@ fn deploy__tx_with_wrong_gas_price_causes_error() {
...
+    let input_amount = 1000;
+    let arb_max_fee = input_amount;
+    let consensus_params = ConsensusParameters::standard();
+
+    let interpreter_params = InterpreterParams {
+        gas_price: interpreter_gas_price,
+        ..Default::default()
+    };
+
+    let ready_tx = TransactionBuilder::upgrade(UpgradePurpose::StateTransition {
```
MitchTurner, line 217: Similar to the "Upload" PR, this could be moved to a helper to declutter this test.

fuel-vm/src/interpreter/executors/main/tests.rs
```diff
@@ -199,3 +199,51 @@ fn deploy__tx_with_wrong_gas_price_causes_error() {
...
+        tx_gas_price,
+        &interpreter_params.gas_costs,
+        &interpreter_params.fee_params,
+    )
+    .unwrap();
+
+    // when
+    let mut transactor =
+        Transactor::<_, Upgrade>::new(MemoryStorage::default(), interpreter_params);
+    let err = transactor.execute_ready_upgrade_tx(ready_tx).unwrap_err();
```
MitchTurner, line 242: This test isn't testing `Upgrade` it's testing `execute_ready_upgrade_tx`. I know this is what `upgrade` uses under the hood, but is that actually the public function we want to support?

Dentosal (approved):

MitchTurner (approved):

PR #714, opened 2024-04-07
A new `Upload` transaction to upload the huge bytecode on the chain

xgreenx:
Related https://github.com/FuelLabs/fuel-core/issues/1754 and https://github.com/FuelLabs/fuel-specs/pull/570

The change adds a new `Upload` transaction that allows uploading huge byte code on chain subsection by subsection.

The `Upload` transaction is chargeable and is twice as expensive as the `Create` transaction. Anyone can submit this transaction.

### Before requesting review
- [x] I have reviewed the code myself

Issue #251, opened 2022-10-28
Update `Interpreter::transact` work with `Script` only


xgreenx:
<img width="1434" alt="image" src="https://user-images.githubusercontent.com/18346821/198716626-496eb47d-3e4e-41bf-a619-5ae36875d8c3.png">

`Interpreter::transact` accepts `Create` and `Script` transactions. But we don't execute any script on `Create`, and it seems redundant and not clear that `transact` support both transactions. `Interpreter` already provides the `deploy` method for the `Create` transaction, so `transact` now can work only with `Script`, and we can remove support of `Create`.

It is breaking change and requires the adaptation of `fuel-core` and `fuel-rs`.

Voxelot:
One thing to note, we'd need to come up with an alternate way to set the change outputs on create txs for any unused fees. Deploy currently only inserts the contract into state without any alterations on the transaction itself.

xgreenx:
Deploy return altered `Create` with actual outputs=)

MitchTurner:
Digging into this some, I'm not sure what work is still required here.

Would love more details @xgreenx

PR #715, opened 2024-04-08
Process `Upload` transaction inside the `Interpreter`

xgreenx:
Part of the https://github.com/FuelLabs/fuel-core/issues/1754

The `Interpreter` supports the processing of the `Upload` transaction. The change affects `InterpreterStorage`, adding `StorageMutate<UploadedBytes>` constrain.

## Checklist

### Before requesting review


fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -580,6 +594,102 @@ where
...
+            .unwrap_or_else(|| UploadedBytecode::Uncompleted {
+                bytecode: vec![],
+                uploaded_parts_number: 0,
+            });
+
+        let new_bytecode = match bytecode {
+            UploadedBytecode::Uncompleted {
+                mut bytecode,
+                uploaded_parts_number,
+            } => {
```
MitchTurner, line 625: Can we put this branch into a helper function? `upload_bytecode_subsection` or something like that.

fuel-vm/src/error.rs
```diff
@@ -367,6 +367,16 @@ pub enum BugVariant {
...
+
+    /// Witness index is out of bounds.
+    #[strum(message = "Witness index is out of bounds.")]
+    WitnessIndexOutOfBounds,
+
+    /// The witness part index is higher than the total number of parts.
+    #[strum(
+        message = "The witness part index is higher than the total number of parts."
+    )]
+    NextPartIndexIsHigherThanTotalNumberOfParts,
```
MitchTurner, line 379: I know this was already decided and there are a lot of places that use this term (this applies to any place using the term `Part`), but I think I'd prefer something like `Subsection` over the term `Part`. `Part` is really vague, and `Subsection` is a bit more clear IMO.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -889,6 +1010,38 @@ where
...
+        tx: Ready<Upload>,
+    ) -> Result<Upload, InterpreterError<S::DataError>> {
+        self.verify_ready_tx(&tx)?;
+
+        let (_, checked) = tx.decompose();
+        let (mut upload, metadata): (Upload, <Upload as IntoChecked>::Metadata) =
+            checked.into();
+        let base_asset_id = *self.base_asset_id();
+        let gas_price = self.gas_price();
+        Self::upload_inner(
```
MitchTurner, line 1032: What's the purpose of having an `inner` function here. And it's so far from the "outer", despite having the same `impl` bounds AFAICT. There are a few places in the code I've noticed that do this and it deepens my context stack while reading the code and makes it harder to read IMO.

I'm a big fan of keeping functions short, but I also like keeping the context as local as possible and making things declarative as possible so you can just read the function signature to get enough context as possible. This code just makes me wonder, what's the difference between `upload` and `upload_inner`.

I suggested adding a `upload_bytecode_subsection` method above. Maybe with the combination of that as well as a `get_bytecode`, etc, you could fit this all in the "outer" `upload` function while keeping things short.

fuel-vm/src/interpreter/executors/main/tests.rs
```diff
@@ -247,3 +247,51 @@ fn upgrade__tx_with_wrong_gas_price_causes_error() {
...
+
+    let interpreter_params = InterpreterParams {
+        gas_price: interpreter_gas_price,
+        ..Default::default()
+    };
+
+    let parts =
+        UploadPart::split_bytecode(&vec![123; 1024], 24).expect("Should split bytecode");
+    let part = parts[0].clone();
+    let ready_tx = TransactionBuilder::upload(UploadBody {
```
MitchTurner, line 268: Maybe move this to a small helper called `upload_tx_with_wrong_gas_price`. That's all the test reader needs to know about this `ready_tx` (which could be renamed to `bad_tx` or something like that. The `parts` could even go with it.

This will declutter the test a lot, shortening it by like 20 lines.

fuel-vm/src/tests/upload.rs
```diff
@@ -0,0 +1,366 @@
...
+        vec![Output::change(Default::default(), 0, AssetId::BASE)],
+        vec![],
+    )
+    .into_checked_basic(Default::default(), &Default::default())
+    .expect("Failed to generate checked tx")
+    .test_into_ready()
+}
+
+#[test]
+fn transact_uploads_bytecode_with_one_part() {
```
MitchTurner, line 76: These tests are beautiful. Thanks!

Nit: I prefer to put double `__` between the SUT and the test case: `transact__uploads_bytecode_with_one_part`. If you don't like that then no pressure, it's not industry standard, but we already have a `#![allow(non_snake_case)]` on the file, so it wouldn't hurt.

Regardless, I don't want to be too picky since (like I said) these tests are already above par and the test name would be semantically identical.

fuel-vm/src/interpreter/executors/main.rs
xgreenx: It is related to [this](https://github.com/FuelLabs/fuel-vm/issues/251).

image


The idea is that the `run` function should only work with the `Script` type since you can run only script transactions. All other types of transactions don't require `Interpreter`'s features or context.

I'm keeping each section separately because I hope that in the future, we will refactor `Interpreter`=D

fuel-vm/src/error.rs
xgreenx: Subsection sounds better, I will update it. I will start with https://github.com/FuelLabs/fuel-vm/pull/714 first.

MitchTurner (dismissed):

fuel-vm/src/error.rs
xgreenx: https://github.com/FuelLabs/fuel-vm/pull/714/commits/b4f2cdd65888a98ee045eefe9ecd4cf066593aeb

fuel-vm/src/memory_client.rs
```diff
@@ -96,6 +97,11 @@ impl<Ecal: EcalHandler> MemoryClient<Ecal> {
         self.transactor.upgrade(tx)
     }
 
+    /// Executes `Upload` transaction.
+    pub fn upload(&mut self, tx: Checked<Upload>) -> Option<Upload> {
```
Dentosal, line 101: nit: Consider returning the error here instead of erasing it

Dentosal (dismissed):

Dentosal (approved):

PR #719, opened 2024-04-10
Handle padded_len_* overflows gracefully

Dentosal:
There was a bug in `padded_len_word` and `padded_len_usize` functions that would overflow on inputs close to max value of the type. While inputs around the max value were never sensible, at least the LDC opcode could be used to trigger the overflow.

## Checklist

### Before requesting review


fuel-tx/src/tests/bytes.rs
```diff
@@ -733,8 +733,8 @@ fn script_input_coin_data_offset() {
                             .inputs_predicate_offset_at(offset)
                             .expect("Failed to fetch offset");
 
-                        assert_ne!(bytes::padded_len(&predicate), predicate.len());
-                        assert_eq!(bytes::padded_len(&predicate), len);
+                        assert_ne!(bytes::padded_len(&predicate), Some(predicate.len()));
```
MitchTurner, line 736: Why even include the `ne` case? There are many things that it's not equal to :)

fuel-tx/src/tests/bytes.rs
```diff
@@ -826,8 +826,8 @@ fn upgrade_input_coin_data_offset() {
                     .inputs_predicate_offset_at(offset)
                     .expect("Failed to fetch offset");
 
-                assert_ne!(bytes::padded_len(&predicate), predicate.len());
-                assert_eq!(bytes::padded_len(&predicate), len);
+                assert_ne!(bytes::padded_len(&predicate), Some(predicate.len()));
```
MitchTurner, line 829: Same question.

fuel-tx/src/transaction/types/chargeable_transaction.rs
```diff
@@ -265,7 +265,11 @@ mod field {
                     .and_then(|predicate| {
                         self.inputs_offset_at(idx).map(|inputs| inputs + predicate)
                     })
-                    .zip(input.predicate_len().map(bytes::padded_len_usize))
+                    .zip(
+                        input
+                            .predicate_len()
+                            .map(|l| bytes::padded_len_usize(l).unwrap_or(usize::MAX)),
```
MitchTurner, line 271: You can use `and_then` instead of `map` here to keep an `Option` and not need the `unwrap_or`. 

The `usize::MAX` feels arbitrary to me.

fuel-tx/src/transaction/types/input.rs
```diff
@@ -455,9 +455,9 @@ impl Input {
             Input::CoinPredicate(_) => InputRepr::Coin.coin_predicate_offset(),
             Input::MessageCoinPredicate(_) => InputRepr::Message.data_offset(),
             Input::MessageDataPredicate(MessageDataPredicate { data, .. }) => {
-                InputRepr::Message
-                    .data_offset()
-                    .map(|o| o + bytes::padded_len(data))
+                InputRepr::Message.data_offset().map(|o| {
+                    o.saturating_add(bytes::padded_len(data).unwrap_or(usize::MAX))
```
MitchTurner, line 459: The `usize::Max` kinda makes more sense here with the `saturating_add`, but I haven a hard time with `None` for len being equivalent to `usize::Max`.

You could do a:
```rs
bytes::padded_len(data).map(|len| InputRepr::Message.data_offset().map(|o| o.saturating_add(len))
```
Not really an improvement on readability though :/ So up to you.

fuel-tx/src/transaction/types/input.rs
```diff
@@ -470,9 +470,11 @@ impl Input {
         match self {
             Input::CoinPredicate(CoinPredicate { predicate, .. })
             | Input::MessageCoinPredicate(MessageCoinPredicate { predicate, .. })
-            | Input::MessageDataPredicate(MessageDataPredicate { predicate, .. }) => self
-                .predicate_offset()
-                .map(|o| o + bytes::padded_len(predicate)),
+            | Input::MessageDataPredicate(MessageDataPredicate { predicate, .. }) => {
+                self.predicate_offset().map(|o| {
+                    o.saturating_add(bytes::padded_len(predicate).unwrap_or(usize::MAX))
```
MitchTurner, line 475: same here

fuel-tx/src/transaction/types/script.rs
```diff
@@ -233,7 +233,9 @@ mod field {
                 return body.script_data_offset;
             }
 
-            self.script_offset() + bytes::padded_len(self.body.script.as_slice())
+            self.script_offset().saturating_add(
+                bytes::padded_len(self.body.script.as_slice()).unwrap_or(usize::MAX),
```
MitchTurner, line 237: same here

fuel-tx/src/transaction/types/script.rs
```diff
@@ -247,8 +249,9 @@ mod field {
         }
 
         fn body_offset_end(&self) -> usize {
-            self.script_data_offset()
-                + bytes::padded_len(self.body.script_data.as_slice())
+            self.script_data_offset().saturating_add(
+                bytes::padded_len(self.body.script_data.as_slice()).unwrap_or(usize::MAX),
```
MitchTurner, line 253: and here

fuel-types/src/bytes.rs
```diff
@@ -68,9 +40,23 @@ pub unsafe fn from_slice_unchecked<const N: usize>(buf: &[u8]) -> [u8; N] {
...
-    assert_eq!(WORD_SIZE * 2, padded_len(&[0; WORD_SIZE * 2]));
+    assert_eq!(Some(WORD_SIZE * 0), padded_len(&[]));
+    assert_eq!(Some(WORD_SIZE * 1), padded_len(&[0]));
+    assert_eq!(Some(WORD_SIZE * 1), padded_len(&[0; WORD_SIZE]));
+    assert_eq!(Some(WORD_SIZE * 2), padded_len(&[0; WORD_SIZE + 1]));
+    assert_eq!(Some(WORD_SIZE * 2), padded_len(&[0; WORD_SIZE * 2]));
+}
+
+#[test]
+fn padded_len_value() {
```
MitchTurner, line 51: Can we name this something like `padded_len_usize__gives_values_in_increments_of_8` or something like that. To say what the tested behavior is.

fuel-types/src/bytes.rs
```diff
@@ -68,9 +40,23 @@ pub unsafe fn from_slice_unchecked<const N: usize>(buf: &[u8]) -> [u8; N] {
 #[allow(clippy::erasing_op)]
 #[allow(clippy::identity_op)]
 fn padded_len_to_fit_word_len() {
-    assert_eq!(WORD_SIZE * 0, padded_len(&[]));
-    assert_eq!(WORD_SIZE * 1, padded_len(&[0]));
-    assert_eq!(WORD_SIZE * 1, padded_len(&[0; WORD_SIZE]));
-    assert_eq!(WORD_SIZE * 2, padded_len(&[0; WORD_SIZE + 1]));
-    assert_eq!(WORD_SIZE * 2, padded_len(&[0; WORD_SIZE * 2]));
+    assert_eq!(Some(WORD_SIZE * 0), padded_len(&[]));
```
MitchTurner, line 43: This test could use a clearer name of what the expected behavior is.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -587,7 +587,13 @@ where
...
             .map_err(|_| PanicReason::MemoryOverflow)?;
 
-        let length = bytes::padded_len_word(length_unpadded);
+        let length = bytes::padded_len_usize(
+            length_unpadded
+                .try_into()
+                .map_err(|_| PanicReason::MemoryOverflow)?,
+        )
+        .map(|len| len as Word)
+        .unwrap_or(Word::MAX);
```
xgreenx, line 596: Maybe we also want to return `PanicReason::MemoryOverflow`?

fuel-tx/src/tests/bytes.rs
Dentosal: No idea. It was there already, and I just kept the old tests as-is.

fuel-tx/src/tests/bytes.rs
Dentosal: Same answer.

fuel-tx/src/transaction/types/chargeable_transaction.rs
Dentosal: Using `and_then` here would mean combining together (a) input has no `predicate_len` and (b) that the padding overflows. These are fundamentally different things. The idea behind using `usize::MAX` is that the value was already quite close to it, and impossibly large. Any actual usage of the return value will correctly cause `MemoryOverflow` or similar panic. The alternative would be to return `Option<Result<usize, PanicReason>>` , and that feels a bit excessive. Alternatively we could just Rust-panic on such sizes, but I'm not confident that we actually verify the sizes beforehands.

xgreenx (dismissed):

xgreenx (approved):

PR #720, opened 2024-04-10
A new `Upload` transaction to upload the huge bytecode on the chain

xgreenx:
Related https://github.com/FuelLabs/fuel-core/issues/1754 and https://github.com/FuelLabs/fuel-specs/pull/570

The change adds a new `Upload` transaction that allows uploading huge byte code on chain subsection by subsection.

The `Upload` transaction is chargeable and is twice as expensive as the `Create` transaction. Anyone can submit this transaction.

### Before requesting review


fuel-tx/src/tests/display.rs
```diff
@@ -32,6 +32,15 @@ fn to_from_str() {
 
             assert_eq!(tx, tx_p);
         });
+    TransactionFactory::<_, Upload>::from_seed(1295)
```
MitchTurner, line 35: I was gonna suggest maybe drying this up a little with a shared function:
```
fn transaction_round_trip_serde<T>()
    where 
        TransactionFactory<StdRng, T>: Iterator,
        T: Into<Transaction>
     {
    TransactionFactory::<_, Upgrade>::from_seed(1295)
        .take(20)
        .for_each(|(tx, _)| {
            // given
            let tx: Transaction = tx.into();
            // when
            let tx_p = tx.to_json();
            let tx_p = Transaction::from_json(tx_p).expect("failed to restore tx");
            // then
            assert_eq!(tx, tx_p);
}
```
I don't know if this is much better with all the generics.

I think the name `to_from_str` could be a little clearer either way though.

fuel-tx/src/tests/valid_cases/transaction/upload.rs
```diff
@@ -0,0 +1,857 @@
...
+
+    // When
+    let result = tx.check(block_height, &test_params());
+
+    // Then
+    assert_eq!(Err(ValidityError::TransactionMaxFeeNotSet), result);
+}
+
+#[test]
+fn max_iow_reached_max_inputs() {
```
MitchTurner, line 244: I think we should not use `iow` like on the other tests.

MitchTurner: There are the multiple spots.

fuel-tx/src/tests/valid_cases/transaction/upload.rs
```diff
@@ -0,0 +1,857 @@
...
+    let result = tx.check(block_height, &test_params());
+
+    // Then
+    assert_eq!(
+        Err(ValidityError::TransactionUploadRootVerificationFailed),
+        result
+    );
+}
+
+// TODO: Remove `#[ignore]` when https://github.com/FuelLabs/fuel-vm/issues/716 is resolved
```
MitchTurner, line 689: I don't  see the `#[ignore]`.

fuel-tx/src/transaction.rs
```diff
@@ -1165,9 +1303,31 @@ mod tests {
...
-            create_with_no_witnesses.metered_bytes_size()
-                + create_with_witnesses.witnesses.size_dynamic()
+            tx_with_witnesses.metered_bytes_size(),
+            tx_with_no_witnesses.metered_bytes_size()
+                + tx_with_witnesses.witnesses.size_dynamic()
+        );
+    }
+
+    #[test]
+    fn upload_metered_data_includes_witnesses() {
```
MitchTurner, line 1313: This is just testing that the size of the tx bytes increases by 8 bytes + witness bytes size. I feel like that's what the test should be named.

Is there a way to inspect the bytes to show that they are equal except for the added witness + 8 bytes?

MitchTurner (commented):
Tangentially related:

Something I've said before, and I still think would be good is instead of `Upload` and `Script` and `Create`, etc, we should have our transactions named `UploadTx` and `ScriptTx` and `CreateTx`, etc. I know the context is in the namespace, but it is never with it's namespace, so unless you understand the context well, you have to make couple clicks before you understand what they are.

Event then, it's far from obvious that `Script` would be a type of transaction, to me `Script` suggests that it is the code run by smart contracts, and never what you'd use for a simple transfer (but it is what you'd use). So even `ScriptTx` probably isn't clear enough.

fuel-tx/src/transaction.rs
MitchTurner: Oh sorry. I meant, not just check that it's longer by `witness.size()` (I'm confused why `witness.size()` is equal to 64 + 8 (where does the 8 come from)).

I was saying that it might make sense to see that the witness bytes were added to the correct spot in the metered bytes. The test name kinda implies that it's checking that the bytes match.

MitchTurner (approved):

PR #728, opened 2024-05-02
Add a single stepping example

Dentosal:
Adds a small example to demonstrate single-stepping to show execution trace.

### Before requesting review

### DIFF
--- a/fuel-vm/examples/single_step.rs
+++ b/fuel-vm/examples/single_step.rs
@@ -1,0 +1,96 @@
+//! This example shows how you can run the VM in single-stepping mode,
+//! allowing for e.g. visualization of the execution state at each step.
+
+use fuel_asm::{
+    op,
+    RawInstruction,
+    RegId,
+};
+use fuel_tx::{
+    ConsensusParameters,
+    TransactionBuilder,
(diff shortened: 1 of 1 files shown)


xgreenx (approved):

fuel-vm/examples/single_step.rs
```diff
@@ -0,0 +1,96 @@
...
+                                .unwrap_or_else(|| "???".to_owned()),
+                        );
+                    }
+                    DebugEval::Continue => {}
+                }
+                t = vm.resume().expect("panicked");
+            }
+            ProgramState::VerifyPredicate(d) => {
+                println!("paused on debugger {d:?} (in predicate)");
+                t = vm.resume().expect("panicked");
```
MitchTurner, line 92: Nit: I feel like we should be using `unwrap` in all these places that call `expect`. `expect` is meant to explain why it's expected to be `Ok`. In all these cases we just say that it's failing.

Dentosal: This is an example, so I'm using expect to highlight that in real-world usage you're supposed to handle the errors.

PR #729, opened 2024-05-02
Deny clippy::arithmetic_side_effects for fuel-merkle

Dentosal:
Closes https://github.com/FuelLabs/fuel-vm/issues/170

Denies the following lints:
* `clippy::arithmetic_side_effects`
* `clippy::cast_sign_loss`
* `clippy::cast_possible_truncation`
* `clippy::cast_possible_wrap`

## Checklist

### Before requesting review


fuel-merkle/src/common/path.rs
```diff
@@ -1,44 +1,36 @@
...
 where
     T: Msb,
 {
-    fn get_instruction(&self, index: u32) -> Option<Instruction> {
+    fn get_instruction(&self, index: u32) -> Option<Side> {
         self.get_bit_at_index_from_msb(index).map(Into::into)
     }
-}
 
-impl<T> ComparablePath for T
```
Dentosal, line 37: Merged into `Path` trait, as they were not used separately.

fuel-merkle/src/common/path.rs
```diff
@@ -1,44 +1,36 @@
-use crate::common::{
-    Bit,
-    Msb,
-};
+use crate::common::Msb;
 
-pub enum Instruction {
```
Dentosal, line 6: Renamed to `Side`, and used in more places

fuel-merkle/src/common/msb.rs
```diff
@@ -1,50 +1,32 @@
-#[derive(Debug, Eq, PartialEq)]
-pub enum Bit {
```
Dentosal, line 2: Replaced with `bool` or `Side` enum, depending on which made sense.

fuel-merkle/src/common/msb.rs
```diff
@@ -1,50 +1,32 @@
...
-    _0 = 0,
-    _1 = 1,
-}
-
-trait GetBit {
-    fn get_bit(&self, bit_index: u32) -> Option<Bit>;
-}
```
Dentosal, line 9: This was used only once, below. I just inlined it.

fuel-merkle/src/common/position.rs
```diff
@@ -110,24 +109,38 @@ impl Position {
...
-        let index = self.in_order_index() as i64 + shift * self.direction();
-        Self::from_in_order_index(index as u64)
+        #[allow(clippy::arithmetic_side_effects)]
+        let shift = 1 << (self.height() + 1); // height() < u32::MAX
+        let this = self.in_order_index();
+        #[allow(clippy::arithmetic_side_effects)]
+        Self::from_in_order_index(match self.orientation() {
+            // height in relation to in_order_index cannot underflow
+            Side::Left => this - shift,
+            // it's not possible to actually construct a tree so larget his would overflow
```
bvrooman, line 127:
```suggestion
            // it's not possible to actually construct a tree so large that this would overflow
```

fuel-merkle/src/common/position.rs
```diff
@@ -137,16 +150,34 @@ impl Position {
...
+    /// The child position of the current position given by the direction.
+    /// A child position has a height less 1 than the current position.
+    ///
+    /// A child position is calculated as a function of the current position's
+    /// index and height, and the supplied direction. The left child
+    /// position has the in-order index arriving before the current index;
+    /// the right child position has the in-order index arriving after the
+    /// current index.
+    pub fn child(self, side: Side) -> Self {
+        assert!(self.is_node());
```
bvrooman, line 162: Should we consider using `debug_assert!` here instead? Realistically, the way we use this class means that we never call `child` on a leaf, but theortically, this can be called in any context, leading to a panic. At the same time, without the `assert!`, we run the risk of overflowing arithmetic. Alternatively, we could make this a fallible function returning `Result`, but that has many downstream effects.

fuel-merkle/src/common/position.rs
```diff
@@ -374,20 +388,20 @@ mod test {
...
-        assert_eq!(Position(7).left_child(), Position(3));
-        assert_eq!(Position(3).left_child(), Position(1));
-        assert_eq!(Position(1).left_child(), Position(0));
-        assert_eq!(Position(11).left_child(), Position(9));
-        assert_eq!(Position(9).left_child(), Position(8));
+        assert_eq!(Position(7).child(Side::Left), Position(3));
+        assert_eq!(Position(3).child(Side::Left), Position(1));
+        assert_eq!(Position(1).child(Side::Left), Position(0));
+        assert_eq!(Position(11).child(Side::Left), Position(9));
+        assert_eq!(Position(9).child(Side::Left), Position(8));
```
bvrooman, line 395: Is this necessary? It looks like we still have the higher-level function `left_child` (and similarly below, `right_child`)

fuel-merkle/src/common/msb.rs
```diff
@@ -1,50 +1,32 @@
...
         // The bit within the containing byte
         let byte_bit_index = index % 8;
-        self.get(byte_index as usize)
-            .and_then(|byte| byte.get_bit(byte_bit_index))
+        self.get(byte_index as usize).map(|byte| {
+            #[allow(clippy::arithmetic_side_effects)] // checked above
+            let mask = 1 << (7 - byte_bit_index);
+            byte & mask != 0
+        })
     }
```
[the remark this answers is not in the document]
Dentosal: I'm quite acustomed to using `bool` to mean a single bit, and for me the extra enum hurts readibility way more than using bool in a function that's named `get_bit_...`. I added a type alias in 26eb4001.

fuel-merkle/src/common/position.rs
Dentosal: Not anymore, as I re-added those. Restored in fcf38742.

fuel-merkle/src/binary/root_calculator.rs
```diff
@@ -28,9 +28,13 @@ impl MerkleRootCalculator {
...
-        while self.stack.len() > 1 {
-            let right_node = &self.stack[self.stack.len() - 1];
-            let left_node = &self.stack[self.stack.len() - 2];
+        loop {
+            let Some(i) = self.stack.len().checked_sub(2) else {
+                break;
+            };
+            let left_node = &self.stack[i];
+            #[allow(clippy::arithmetic_side_effects)] // stack.len()-2+1 is valid
+            let right_node = &self.stack[i + 1];
```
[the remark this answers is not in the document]
Dentosal: Restored to original in 16e4ba71

fuel-merkle/src/binary/merkle_tree.rs
```diff
@@ -114,13 +111,14 @@ impl<TableType, StorageType> MerkleTree<TableType, StorageType> {
     }
 
     fn root_position(&self) -> Position {
-        // Define a new tree with a leaf count 1 greater than the current leaf
-        // count.
```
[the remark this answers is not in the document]
Dentosal: Yep. 05b7ce2d

fuel-merkle/src/common/position.rs
Dentosal: This assert already existed in the previous version of the code, and I didn't want to change the extenal behavior here. I feel like this is out-of-scope for this PR. That said, we could at least improve the documentation here.

xgreenx (commented):
I'm okay with unwrapping or ignoring the overflow, assuming that memory is finite and the user can't provide an invalid tree that can cause overflow.

I've added [two tests](https://github.com/FuelLabs/fuel-vm/pull/729/commits/6bf695b0e0f0d9c933da4320d964d6bd93ddb9fb) to show the problem. I think other public functions may also fail for specific inputs. It would be nice if we added tests for overflow test cases and handled them everywhere.

fuel-merkle/src/common/position.rs
```diff
@@ -110,24 +109,44 @@ impl Position {
...
     pub fn parent(self) -> Self {
         let shift = 1 << self.height();
-        let index = self.in_order_index() as i64 + shift * self.direction();
-        Self::from_in_order_index(index as u64)
+        let this = self.in_order_index();
+        #[allow(clippy::arithmetic_side_effects)]
+        Self::from_in_order_index(match self.orientation() {
+            // height in relation to in_order_index cannot underflow
+            Side::Left => this - shift,
+            // it's not possible to actually construct a tree so larget his would overflow
```
xgreenx, line 147: I don't think that we can guarantee that this function will not overflow since it is public, and the `Postiion` could be created with an incorrect value for `from_in_order_index`.

fuel-merkle/src/binary/verify.rs
```diff
@@ -9,32 +9,34 @@ use crate::{
...
-    while (1u64 << path_length) < num_leaves {
-        path_length += 1;
-
-        if path_length >= 64 {
-            break;
-        }
+/// Returns None if:
+/// - `num_leaves` is 0
+/// - the result wouldn't fit into an usize
+fn path_length_from_key(key: u64, num_leaves: u64) -> Option<usize> {
```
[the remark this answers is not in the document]
Dentosal: Because the only place where this is used, it's compared to `proof_set.len()`. I don't understand what you mean by 

> The proof set may contain fewer leaves than `num_leaves` and still be valid and fit into the memory.

fuel-merkle/src/binary/verify.rs
```diff
@@ -9,32 +9,34 @@ use crate::{
...
+    if num_leaves == 0 {
+        return None;
     }
+
+    let path_length =
```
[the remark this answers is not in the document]
Dentosal: Fixed in 0a68b5de

fuel-merkle/src/common/path_iterator.rs
```diff
@@ -139,7 +139,8 @@ where
         // 0      7     00  02  04  06  08  10  12  14   252 254
         //              00  01  02  03  04  05  06  07   126 127
         //
-        let initial_offset = T::key_size_in_bits() - root.height();
+        #[allow(clippy::arithmetic_side_effects)] // Responsibility of the trait implementor
+        let initial_offset = T::key_size_bits() - root.height();
```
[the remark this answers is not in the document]
Dentosal: The size is not always the same. For instance `Position::key_size_bits() == 8` and `Node::key_size_bits() == 32`. Yes we could use `const KEY_SIZE_BITS: usize;` in the trait instead, if we know that the calculation can always be done in a const context.

fuel-merkle/src/common/position.rs
```diff
@@ -110,24 +109,44 @@ impl Position {
...
     }
 
     /// The sibling position.
     /// A position shares the same parent and height as its sibling.
     pub fn sibling(self) -> Self {
-        let shift = 1 << (self.height() + 1);
-        let index = self.in_order_index() as i64 + shift * self.direction();
-        Self::from_in_order_index(index as u64)
+        #[allow(clippy::arithmetic_side_effects)]
+        let shift = 1 << (self.height() + 1); // height() < u32::MAX
```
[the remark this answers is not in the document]
Dentosal: ced9f8cd

fuel-merkle/src/common/position.rs
Dentosal: Ah right. Fixed in ced9f8cd.

fuel-merkle/src/common/path_iterator.rs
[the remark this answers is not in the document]
xgreenx: Got it. Then it sounds like if the implementer did something wrong, it could fail here? ANd maybe it is better to return an error then or at least a panic=)

fuel-merkle/src/common/position.rs
```diff
@@ -55,7 +62,7 @@ use core::convert::Infallible;
 /// This allows us to construct a `Position` (and its in-order index) by
 /// providing either an in-order index directly or, in the case of a leaf, a
 /// leaf index. This functionality is captured by `from_in_order_index()` and
-/// `from_leaf_index()` respectively.
+/// `from_leaf_index_unwrap()` respectively.
```
xgreenx, line 65: I modified comment, sorry=D

```suggestion
/// `from_leaf_index()` respectively.
```

fuel-merkle/src/common/position.rs
```diff
@@ -110,43 +114,84 @@ impl Position {
...
-    pub fn parent(self) -> Self {
-        let shift = 1 << self.height();
-        let index = self.in_order_index() as i64 + shift * self.direction();
-        Self::from_in_order_index(index as u64)
+    pub fn parent(self) -> Result<Self, GetNodeError> {
+        let shift = 1u64
+            .checked_shl(self.height())
+            .ok_or(GetNodeError::CannotExist)?;
+        let this = self.in_order_index();
+        Ok(Self::from_in_order_index(match self.orientation() {
```
xgreenx, line 148: Hmm, we could also panic inside of the `self.orientation`:

image

fuel-merkle/src/common/position.rs
```diff
@@ -110,43 +114,84 @@ impl Position {
...
+        Self::from_leaf_index(index).expect("Index too large")
     }
 
     /// The sibling position.
     /// A position shares the same parent and height as its sibling.
-    pub fn sibling(self) -> Self {
-        let shift = 1 << (self.height() + 1);
-        let index = self.in_order_index() as i64 + shift * self.direction();
-        Self::from_in_order_index(index as u64)
+    pub fn sibling(self) -> Result<Self, GetNodeError> {
```
xgreenx, line 129: You can put this function under `#[cfg(test)]`

fuel-merkle/src/common/position.rs
```diff
@@ -110,43 +114,84 @@ impl Position {
...
     /// relative to this position.
-    pub fn uncle(self) -> Self {
-        self.parent().sibling()
-    }
-
-    /// The left child position.
-    /// See [child](Self::child).
-    pub fn left_child(self) -> Self {
-        self.child(LEFT_CHILD_DIRECTION)
+    pub fn uncle(self) -> Result<Self, GetNodeError> {
```
xgreenx, line 157: You can put this function under `#[cfg(test)]`

fuel-merkle/src/common/path_iterator.rs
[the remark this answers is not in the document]
Dentosal: Yep, a panic would probably be in order.

fuel-merkle/src/binary/merkle_tree.rs
```diff
@@ -313,23 +320,30 @@ where
     TableType: Mappable<Key = u64, Value = Primitive, OwnedValue = Primitive>,
     StorageType: StorageMutate<TableType, Error = StorageError>,
 {
-    pub fn push(&mut self, data: &[u8]) -> Result<(), StorageError> {
-        let node = Node::create_leaf(self.leaves_count, data);
-        self.storage.insert(&node.key(), &node.as_ref().into())?;
+    pub fn push(&mut self, data: &[u8]) -> Result<(), TreeExtendError<StorageError>> {
```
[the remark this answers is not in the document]
Dentosal: Because there's no way to signal "tree too large" using the `StorageError` type

fuel-merkle/src/common/path_iterator.rs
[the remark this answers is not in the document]
Dentosal: 5ea05c09

fuel-merkle/src/binary/merkle_tree.rs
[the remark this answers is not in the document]
xgreenx: image

fuel-merkle/src/common/position.rs
xgreenx: It is used in tests=)

Dentosal: Done in 432e7c97

fuel-merkle/src/binary/merkle_tree.rs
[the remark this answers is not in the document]
Dentosal: Oops. Fixed in 51e37c07.

fuel-merkle/src/binary/merkle_tree.rs
```diff
@@ -370,7 +390,7 @@ where
 {
     let mut head = subtree.clone();
     while let Some(mut head_next) = head.take_next() {
-        head = join_subtrees(&mut head_next, &mut head);
+        head = join_subtrees(&mut head_next, &mut head).expect("Failed to join subtrees");
```
MitchTurner, line 393: Can we specify why we don't expect this to panick? Or is that dependent on the caller of `build_root_node`?

fuel-merkle/src/binary/root_calculator.rs
```diff
@@ -47,9 +49,10 @@ impl MerkleRootCalculator {
             return empty_sum().to_owned()
         }
         while self.stack.len() > 1 {
-            let right_child = self.stack.pop().expect("Unable to pop element from stack");
-            let left_child = self.stack.pop().expect("Unable to pop element from stack");
-            let merged_node = Node::create_node(&left_child, &right_child);
+            let right_child = self.stack.pop().expect("Checked in loop bound");
+            let left_child = self.stack.pop().expect("Checked in loop bound");
+            let merged_node = Node::create_node(&left_child, &right_child)
+                .expect("Unable to create a node");
```
MitchTurner, line 55: Can we explain why we expect this to not panic? (Same with above)

fuel-merkle/src/common/subtree.rs
Dentosal, line 1: The `Subtree`  was just a hand-rolled non-empty linked list. It was always used wrapped in an option, so the use was exactly equivalent to a linked list. I first replaced it with `alloc::collections::LInkedList` but why not use a normal `Vec` instead?

fuel-merkle/src/binary/merkle_tree.rs
Dentosal: Should be fixed now, I did refactor the whole function across several commits.

fuel-merkle/src/binary/root_calculator.rs
Dentosal: Refactored away.

fuel-merkle/src/common/position_path.rs
```diff
@@ -47,9 +50,15 @@ pub struct PositionPathIter {
...
+    /// Panics if leaves_count is zero, as the tree is not valid
     pub fn new(root: Position, leaf: Position, leaves_count: u64) -> Self {
         Self {
-            rightmost_position: Position::from_leaf_index(leaves_count - 1),
+            rightmost_position: Position::from_leaf_index(
+                leaves_count
+                    .checked_sub(1)
+                    .expect("Path to a tree without leaves"),
+            )
+            .unwrap(),
```
xgreenx, line 61: I think we can't guaranty that this will not fail. Is someone provides invalid `leaves_count`, it will fail

xgreenx (commented):
LGTM! It would be nice if @bvrooman reviewed the change=)

bvrooman (approved):
lgtm

fuel-merkle/src/common/position_path.rs
bvrooman: If the tree has a root, then `leaves_count` will be at least 1. In real use cases, this should be okay. But if someone were to provide arguments that are inconsistent with a real tree, this could fail. I don't think that's an issue.

PR #736, opened 2024-05-19
Fix ldc opcode in internal contexts

Dentosal:
LDC opcode was behaving incorrectly in internal contexts, i.e. when called from within a contract.

https://github.com/FuelLabs/fuel-specs/pull/584

## Checklist

### Before requesting review

### DIFF
--- a/fuel-vm/src/interpreter/blockchain.rs
+++ b/fuel-vm/src/interpreter/blockchain.rs
@@ -585,7 +585,7 @@ where
             .try_into()
             .map_err(|_| PanicReason::MemoryOverflow)?;
 
-        let current_contract = current_contract(&self.context, self.fp, self.memory)?;
+        let current_contract = current_contract(self.context, self.fp, self.memory)?;
 
         let length = bytes::padded_len_usize(
             length_unpadded
--- a/fuel-vm/src/interpreter/blockchain.rs
+++ b/fuel-vm/src/interpreter/blockchain.rs
@@ -102,8 +102,6 @@ where
         // We will charge for the contracts size in the `load_contract_code`.
         self.gas_charge(gas_cost.base())?;
         let contract_max_size = self.contract_max_size();
-        let current_contract =
-            current_contract(&self.context, self.registers.fp(), &self.memory)?;
         let (
             SystemRegisters {
                 cgas,
@@ -119,6 +117,7 @@ where
         ) = split_registers(&mut self.registers);
         let input = LoadContractCodeCtx {
             memory: &mut self.memory,
+            context: &self.context,
             profiler: &mut self.profiler,
             storage: &mut self.storage,
             contract_max_size,
@@ -127,7 +126,6 @@ where
                 &mut self.panic_context,
             ),
             gas_cost,
-            current_contract,
             cgas,
             ggas,
             ssp,
@@ -538,10 +536,10 @@ where
 struct LoadContractCodeCtx<'vm, S, I> {
     contract_max_size: u64,
     memory: &'vm mut Memory,
+    context: &'vm Context,
     profiler: &'vm mut Profiler,
     input_contracts: InputContracts<'vm, I>,
     storage: &'vm S,
-    current_contract: Option<ContractId>,
     gas_cost: DependentCost,
     cgas: RegMut<'vm, CGAS>,
     ggas: RegMut<'vm, GGAS>,
@@ -576,7 +574,6 @@ where
     {
         let ssp = *self.ssp;
         let sp = *self.sp;
-        let fp = *self.fp;
         let region_start = ssp;
 
         if ssp != sp {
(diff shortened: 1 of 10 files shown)


fuel-vm/src/interpreter/blockchain/code_tests.rs
```diff
@@ -14,7 +14,7 @@ use crate::{
 use fuel_tx::Contract;
 
 #[test]
-fn test_load_contract() -> IoResult<(), Infallible> {
+fn test_load_contract_in_script() -> IoResult<(), Infallible> {
```
MitchTurner, line 17: Shouldn't these tests be against the `Interpreter` instead of against `LoadContractCodeCtx`?

Dentosal: They should probably be against both. I still hate that `LoadContractCodeCtx` and the others like that even exist.

xgreenx (dismissed):

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -636,19 +636,20 @@ where
...
-            let old_code_size = Word::from_be_bytes(self.memory.read_bytes(fp)?);
-
+        // Update frame code size, if we have a stack frame (i.e. fp > 0)
+        if self.context.is_internal() {
+            let code_size_ptr =
+                (*self.fp).saturating_add(CallFrame::code_size_offset() as Word);
+            let old_code_size =
+                Word::from_be_bytes(self.memory.read_bytes(code_size_ptr)?);
+            let old_code_size = padded_len_word(old_code_size)
+                .expect("Code size cannot overflow with padding");
```
Dentosal, line 646: I would assume the size should already be padded when written in memory. Is it really correct to just compute padding here?

xgreenx: Our code uses unpadded value in the `CallFrame`

image

xgreenx (approved):

fuel-vm/src/interpreter/blockchain.rs
Dentosal: Yep. Shouldn't `CallFrame` use the padded value?

xgreenx: I don't know=) The question is to the compiler team: do they use it somehow or not? Or is it used by anyone=)

Our specification says "Code size in bytes.", that sounds like exact size of the contract without padding.

image

Dentosal: I'll ask the team, but a quick search from Sway codebase seems like they don't use this for anything. Time to remove? :D

Voxelot: Doesnt "word-aligned" mean the same thing as padding here?

Dentosal: Yep.

Moved to use padded size in 7335f9fe.

xgreenx:
Could ou also update the Specification to specify this behaviour explicitly, please?=)

fuel-vm/src/tests/blockchain.rs
```diff
@@ -2047,3 +2047,180 @@ fn coinbase_works() {
 
     assert_eq!(data.as_ref().unwrap(), &*expected);
 }
+
+#[test]
+fn various_ldc_issues_poc() {
```
MitchTurner, line 2052: Is there any way to break this down in to more granular && specific tests? Tests like this are hard to maintain as it's not clear what behavior they are defining.

fuel-vm/src/interpreter/blockchain/code_tests.rs
MitchTurner: Okay. I don't you to fix the architecture in this PR.

fuel-vm/src/interpreter/flow.rs
```diff
@@ -579,27 +584,35 @@ where
...
+        mem_frame.copy_from_slice(&frame.to_bytes());
+        let (mem_code, mem_code_padding) = mem_code.split_at_mut(code_size);
+        read_contract(call.to(), self.storage, mem_code)?;
+        mem_code_padding.fill(0);
+
+        #[allow(clippy::arithmetic_side_effects)] // Checked above
+        let code_start =
+            (*self.registers.system_registers.fp) + CallFrame::serialized_size() as Word;
+
+        *self.registers.system_registers.pc = code_start;
```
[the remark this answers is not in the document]
Dentosal: No? Essentially it was before like this:

```python
content_size = len - frame_len_with_padding
frame_end = fp + content_size # return value
```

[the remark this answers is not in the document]
xgreenx: Ohh, there is `saturating_sub`, I missed that=D

xgreenx (approved):

PR #737, opened 2024-05-22
Panic on instructions with non-zero reserved part

Dentosal:
Spec PR: https://github.com/FuelLabs/fuel-specs/pull/583

See spec pr for extended description.

## Checklist
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (linked above)

### Before requesting review


fuel-asm/src/encoding_tests.rs
```diff
@@ -54,6 +53,29 @@ fn opcode() {
...
+    // Args: 2
+    Instruction::try_from([Opcode::NOT as u8, 0, 0, 0]).unwrap();
+    Instruction::try_from([Opcode::NOT as u8, 0, 0, 1])
+        .expect_err("Reserved part is nonzero");
+
+    // Args: 3
+    Instruction::try_from([Opcode::XOR as u8, 0, 0, 0]).unwrap();
+    Instruction::try_from([Opcode::XOR as u8, 0, 0, 1])
+        .expect_err("Reserved part is nonzero");
+}
```
bvrooman, line 77: While this behaviour is implemented as a macro, and all opcodes inherit this behaviour, it may be worth testing this for all opcodes. From an opcode-centric perspective, we want to show that each opcode fulfills the specification (regardless of implementation). 

Maybe we could have a macro-generated test(s) that tests common behaviours for a given opcode, and maintain a list that invokes this test macro for all opcodes.

But that's up to you - it may be overkill here.

bvrooman (dismissed):

fuel-asm/src/encoding_tests.rs
```diff
@@ -54,6 +53,29 @@ fn opcode() {
...
+
+    // Args: 2
+    Instruction::try_from([Opcode::NOT as u8, 0, 0, 0]).unwrap();
+    Instruction::try_from([Opcode::NOT as u8, 0, 0, 1])
+        .expect_err("Reserved part is nonzero");
+
+    // Args: 3
+    Instruction::try_from([Opcode::XOR as u8, 0, 0, 0]).unwrap();
+    Instruction::try_from([Opcode::XOR as u8, 0, 0, 1])
+        .expect_err("Reserved part is nonzero");
```
xgreenx, line 76: I don't see where are we using arg 2 and arg 3=)

fuel-asm/src/macros.rs
```diff
@@ -713,6 +713,60 @@ macro_rules! op_unpack {
     () => {};
 }
 
+// Generate a method for checking that the reserved part of the
+// instruction is zero. This is private, as invalid instructions
+// cannot be constructed outside this crate.
+macro_rules! op_reserved_part {
```
xgreenx, line 719: I think it is better to move this check on `unpack` level.

Something like this:
image

Dentosal: Can you elaborate why? It was my first instinct as well, but after I realized we can validate this at construction time and never construct invalid `Instruction`s.

xgreenx: From a performance perspective. The check in the constructor repeats the logic in the `unpack`.

If we could remove the `unpack` method and store only required values like `RegId` or `Imm6`, we would be able to check it only once during creation and performance would be not affected=)

fuel-asm/src/encoding_tests.rs
MitchTurner: Makes sense to me.

xgreenx (approved):

xgreenx (approved):

fuel-asm/src/lib.rs
```diff
@@ -49,26 +49,31 @@ pub use panic_reason::PanicReason;
 
 /// Represents a 6-bit register ID, guaranteed to be masked by construction.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "typescript", wasm_bindgen::prelude::wasm_bindgen)]
 pub struct RegId(u8);
 
 /// Represents a 6-bit immediate value, guaranteed to be masked by construction.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
Dentosal, line 58: The derived `Deserialize` impl allows constructing invalid values.

fuel-asm/src/macros.rs
xgreenx: The current approach works faster based on benchmarks than if the instruction is unpacked by default because(the memory layout of the unpacked variant is bigger).

xgreenx (approved):

PR #747, opened 2024-06-04
Use `DependentCost` for `aloc` opcode

xgreenx:
Since we zero bytes during the heap allocation, we need to iterate over `n` bytes of memory. It may be a lot of work so we need to charge for it properly.

image

## Checklist
- [x] If performance characteristic of an instruction change, update gas costs as well or make a follow-up PR for that - https://github.com/FuelLabs/fuel-core/pull/1934

### Before requesting review

### DIFF
--- a/fuel-vm/src/pool.rs
+++ b/fuel-vm/src/pool.rs
@@ -11,7 +11,7 @@ pub trait VmMemoryPool: Sync {
     type Memory: Memory + Send + Sync + 'static;
 
     /// Gets a new VM memory instance from the pool.
-    fn get_new(&self) -> impl core::future::Future<Output = Self::Memory> + Send;
+    fn get_new(&self) -> Self::Memory;
 }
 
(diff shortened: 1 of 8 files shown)


fuel-vm/src/pool.rs
```diff
@@ -1,17 +1,17 @@
...
+use crate::interpreter::MemoryInstance;
 
 /// Trait for a VM memory pool.
 pub trait VmMemoryPool: Sync {
     /// The memory instance returned by this pool.
     type Memory: Memory + Send + Sync + 'static;
 
     /// Gets a new VM memory instance from the pool.
-    fn get_new(&self) -> Self::Memory;
+    fn get_new(&self) -> impl core::future::Future<Output = Self::Memory> + Send;
```
xgreenx, line 14: It is not related to the change, but on `fuel-core` side we are using async function, not sync.

Dentosal (approved):

PR #751, opened 2024-06-05
Add misc tests

Dentosal:
Improves test coverage by testing some code paths that were not previously take.  Will not contain any externally visible changes, only adding more tests and cleaning up existing ones.

### Before requesting review

### DIFF
--- a/fuel-vm/src/interpreter/debug.rs
+++ b/fuel-vm/src/interpreter/debug.rs
@@ -1,9 +1,3 @@
-#[cfg(test)]
-use alloc::{
-    vec,
-    vec::Vec,
-};
-
 use super::Interpreter;
 use crate::prelude::*;
 use fuel_asm::RegId;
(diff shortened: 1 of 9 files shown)


fuel-vm/src/interpreter/debug.rs
```diff
@@ -63,140 +57,192 @@ where
...
+        let gas_limit = 1_000_000;
+        let gas_price = 0;
+        let height = Default::default();
+
+        let script = [
+            op::addi(0x10, RegId::ZERO, 8),
+            op::addi(0x11, RegId::ZERO, 16),
+            op::addi(0x12, RegId::ZERO, 32),
+            op::addi(0x13, RegId::ZERO, 64),
+            op::addi(0x14, RegId::ZERO, 128),
```
xgreenx, line 87: It would be nice to use some constant names for registers. It will connect them here with the values in `suite`

fuel-vm/src/interpreter/debug.rs
```diff
@@ -63,140 +57,192 @@ where
...
+            .add_random_fee_input()
+            .finalize()
+            .into_checked(height, &consensus_params)
+            .expect("failed to generate checked tx")
+            .into_ready(
+                gas_price,
+                consensus_params.gas_costs(),
+                consensus_params.fee_params(),
+            )
+            .unwrap();
```
xgreenx, line 185: It seems we repeat this part several times, maybe makes sense to move it into a function

xgreenx (approved):
Looks good to me, added small [nit]s

Dentosal:
Merging this as-is, those nits will be addressed later when we'll expand the test helper context

PR #752, opened 2024-06-05
Remove Bug type

Dentosal:
Removes `Bug` and `BugVariant` types, opting to `expect` or `unreachable!` instead, as suggested in audit reports. Since we already catch the rust-level panics in fuel-core, we should move the reporting there. This way, we include all crashes and not just those where we had wrapped it in `Bug`.

## Checklist

### Before requesting review


xgreenx:
As we discussed internally: let's stick with `Bug` for now to not change the behavior of the current code

Issue #758, opened 2024-06-14
The `load_contract_code` charges gas after loading of the contract


xgreenx:
`load_contract_code` charges gas [after](https://github.com/FuelLabs/fuel-vm/blob/795d2eebfda2e2ccc407a02bebefa975e1e19b6d/fuel-vm/src/interpreter/blockchain.rs#L617) loading the contract, it should first query its length, charge gas and only then load it in memory. Depending on storage implementation, this could be relevant since users may be able to perform costly actions without paying.

PR #768, opened 2024-06-15
Charge for LDC opcode before loading the contract into memory

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/758

### Before requesting review

### DIFF
--- a/fuel-vm/src/interpreter/blockchain.rs
+++ b/fuel-vm/src/interpreter/blockchain.rs
@@ -613,15 +613,13 @@ where
         self.input_contracts.check(&contract_id)?;
 
         // Fetch the storage contract
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract_bytes = contract.as_ref().as_ref();
-        let contract_len = contract_bytes.len();
         let profiler = ProfileGas {
             pc: self.pc.as_ref(),
             is: self.is,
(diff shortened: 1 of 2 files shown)


fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -613,22 +613,22 @@ where
...
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract_bytes = contract.as_ref().as_ref();
-        let contract_len = contract_bytes.len();
         let profiler = ProfileGas {
             pc: self.pc.as_ref(),
             is: self.is,
             current_contract,
             profiler: self.profiler,
         };
+        let contract_len = contract_size(&self.storage, &contract_id)?;
```
Voxelot, line 622: Does this really save us much expense? We don't know the size of a contract until we load it from disk and check the length of the bytecode slice anyways. 

This is probably going to make us have to load the contract twice from disk in actuality.

xgreenx: Yes, we will load the contract twice, but it will be cached on the rocksdb level, so once=) Plus, we don't load the contract into memory to get its size; we use an optimized function that just gets's size from RocksDB.

Voxelot (approved):

Issue #761, opened 2024-06-14
Verify the impact of inputs/outputs on the performance of the VM


xgreenx:
Some opcodes iterate input / output to find specific entries, but doesn't charge linearly to input / output count. Given that input / output count is relatively limited, it is possible that this is absorbed into the constant gas cost. ([1](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/contract.rs#L81)) / ([2](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/contract.rs#L258)) / ([3](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/blockchain.rs#L125)) / ([4](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/blockchain.rs#L209)) / ([5](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/blockchain.rs#L272)) / ([6](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/blockchain.rs#L309)) / ([7](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/flow.rs#L373)) / ([8](https://github.com/FuelLabs/fuel-vm/blob/v0.49.0/fuel-vm/src/interpreter/metadata.rs#L323))

PR #770, opened 2024-06-15
Cache contract inputs in the VM

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/761

Updated benchmark to cover worst case https://github.com/FuelLabs/fuel-core/pull/1973

## Checklist

### Before requesting review


fuel-vm/src/interpreter.rs
```diff
@@ -644,22 +636,25 @@ impl CheckedMetadata for UploadCheckedMetadata {
...
-            tx_input_contracts,
+            input_contracts,
             panic_context,
         }
     }
 
     /// Checks that the contract is declared in the transaction inputs.
     pub fn check(&mut self, contract: &ContractId) -> SimpleResult<()> {
-        if !self.tx_input_contracts.any(|input| input == contract) {
+        if !self.input_contracts.contains(contract) {
```
MitchTurner, line 657: Essentially, the idea of this PR is making this lookup more performant? Is that right?

fuel-vm/src/interpreter/initialization.rs
```diff
@@ -41,6 +45,33 @@ where
     ) -> Result<(), RuntimeError<S::DataError>> {
         tx.prepare_init_execute();
         self.tx = tx;
+        self.input_contracts = self
```
MitchTurner, line 48: So we might be taking a bit of a hit on performance because of this, but at least its constant time and not an attack vector?

xgreenx: Yes, we calculate it each time now during initialization and later all related operations will be fast

fuel-vm/src/interpreter.rs
xgreenx: Yep

MitchTurner (approved):

MitchTurner (approved):

Issue #756, opened 2024-06-14
Predicate estimation is, almost, unbounded


xgreenx:
The predicate estimation allows `MAX_GAS_PER_TRANSACTION` for each predicate in the transaction. 
<img width="844" alt="image" src="https://github.com/FuelLabs/fuel-vm/assets/18346821/91854fd1-c73c-4c58-85ad-e25bee32ae0f">

Currently, we allow 255 inputs per transaction. It allows the attacker to consume a lot of resources during the estimation of the predicate by submitting transactions with 255 predicates that are using `30M` gas each, which can lead to 85 seconds of execution.

PR #771, opened 2024-06-15
Take into account spent gas during synchronous predicates estimation

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/756

The `fuel-core` already uses synchronous predicate estimation, so we don't need any other changes. 

image

### Before requesting review

### DIFF
--- a/fuel-vm/src/tests/predicate.rs
+++ b/fuel-vm/src/tests/predicate.rs
@@ -570,10 +570,12 @@ fn synchronous_estimate_predicates_respects_total_tx_gas_limit() {
     let limit = 1_000_000;
     let rng = &mut StdRng::seed_from_u64(2322u64);
 
-    let mut params = CheckPredicateParams::default();
-    params.max_gas_per_predicate = limit;
-    params.max_gas_per_tx = limit;
-    params.gas_costs = GasCosts::unit();
+    let params = CheckPredicateParams {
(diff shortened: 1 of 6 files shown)


fuel-vm/src/tests/predicate.rs
```diff
@@ -563,3 +563,48 @@ async fn gas_used_by_predicates_more_than_limit() {
...
+            vec![],
+        ));
+    }
+    let mut transaction = builder.finalize();
+
+    // When
+    let result = transaction.estimate_predicates(&params, MemoryInstance::new());
+
+    // Then
+    assert_eq!(Ok(()), result);
```
Voxelot, line 609: if the estimation runs out of gas due to each predicate running an infinite loop, why is this an `Ok(())`?

xgreenx: It was done a long time ago to allow estimate predicates even if they return false. For cases when signature incorrect and predicate fails

I also don't like it=D I was thinking of modifying the code and returning an error in all cases except false predicate. But it will change how estimation works, so I decided not to include it in this PR.

Voxelot (approved):

fuel-vm/src/tests/predicate.rs
Voxelot: Ok - this seems odd but makes sense given the context

Issue #746, opened 2024-06-03
Test opcodes via external API instead of testing internal functions


xgreenx:

MitchTurner:
I think we could improve the description of this item some. Since part of the work is already done.

PR #780, opened 2024-06-21
Blob tx and instructions

Dentosal:
Work towards https://github.com/FuelLabs/fuel-specs/issues/589. Spec PR: https://github.com/FuelLabs/fuel-specs/pull/592. Core PR https://github.com/FuelLabs/fuel-core/pull/1988.

Open for the first round of reviews, but:
- Typescipt support and tests are still missing
- Some open questions need to be resolved
  - I commented the PR extensively to mark these

Open questions include:
- [ ] Should BSIZ charge dependent on size? (same for CSIZ as well)
- [ ] Should we implement this for predicates?

## Checklist

### Before requesting review

- [ ] [Typescript SDK](https://github.com/FuelLabs/fuels-ts/)
- [ ] Someone else?


fuel-tx/src/transaction/consensus_parameters/gas.rs
```diff
@@ -949,6 +1057,25 @@ impl GasCostsValues {
...
+            GasCostsValues::V3(_v3) => todo!(), // TODO
+            GasCostsValues::V4(v4) => v4.bsiz,
+        }
+    }
+
+    pub fn bldd(&self) -> DependentCost {
+        match self {
+            GasCostsValues::V1(_v1) => todo!(), // TODO
+            GasCostsValues::V2(_v2) => todo!(), // TODO
+            GasCostsValues::V3(_v3) => todo!(), // TODO
```
Dentosal, line 1077: I'm not sure what should be done for gas prices on old versions, where this opcode doesn't exist yet. Possible options include:

- `0`, carries risk of free gas use and generally seems not suitable
- `u64::MAX`, causes gas to immediately run out which would at least disallow transacting with incorrect behavior
- rust-level panic, and to disallow incorrect configurations somewhere else
- vm-level panic

fuel-tx/src/transaction/consensus_parameters/gas/default_gas_costs.rs
```diff
@@ -174,6 +174,14 @@ pub fn default_gas_costs() -> GasCostsValues {
...
             units_per_gas: 5,
         },
+        bsiz: DependentCost::LightOperation {
+            base: 17,
+            units_per_gas: 790,
+        },
+        bldd: DependentCost::LightOperation {
+            base: 15,
+            units_per_gas: 272,
+        },
```
Dentosal, line 184: Needs a fuel-core benches run on reference hardware

fuel-vm/src/tests/coins.rs
Dentosal, line 1: These are moved to `test_helpers`, so the same `RunResult` type can be used by multiple tests

fuel-vm/src/storage/predicate.rs
Dentosal, line 1: These are moved to `test_helpers`, so the same `RunResult` type can be used by multiple tests

fuel-vm/src/checked_transaction/types.rs
```diff
@@ -401,3 +402,79 @@ pub mod upload {
...
+        Cacheable,
+        Chargeable,
+        ConsensusParameters,
+        FormatValidityChecks,
+    };
+    use fuel_types::BlockHeight;
+
+    /// Metadata produced by checking [`fuel_tx::Blob`].
+    #[derive(Debug, Clone, Eq, PartialEq, Hash)]
+    pub struct CheckedMetadata {
```
Dentosal, line 432: This struct seems to repeat with same contents for almost all tx types does it make sense to have separate copy for each?

fuel-vm/src/interpreter/blob.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    pub(crate) fn blob_load_data(
+        &mut self,
+        dst_ptr: Word,
+        blob_id_ptr: Word,
+        blob_offset: Word,
+        len: Word,
+    ) -> IoResult<(), S::DataError> {
+        self.dependent_gas_charge(self.interpreter_params.gas_costs.bldd(), len)?;
```
Dentosal, line 66: Should we charge for the size of the contract as well, and not just the len copied? Essentially same question applies for the `bsiz`, which could possibly be constant cost.

fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -59,7 +59,7 @@ where
                 ))
             })?,
         );
-        if pc < self.registers[RegId::IS] || pc >= self.registers[RegId::SSP] {
+        if pc < self.registers[RegId::IS] || pc >= self.registers[RegId::SP] {
```
Dentosal, line 62: Note: this alone relaxes the stack data cannot be excutable -restriction.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -972,11 +972,10 @@ impl OwnershipRegisters {
 
     /// Empty range is owned iff the range.start is owned
     pub(crate) fn has_ownership_heap(&self, range: &Range<Word>) -> bool {
-        // TODO implement fp->hp and (addr, size) validations
-        // fp->hp
-        // it means $hp from the previous context, i.e. what's saved in the
-        // "Saved registers from previous context" of the call frame at
-        // $fp`
```
Dentosal, line 979: The todo comment here is completely outdated. We do indeed check for this condition below, but for some reason this comment was never removed.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -972,11 +972,10 @@ impl OwnershipRegisters {
...
+        if range.is_empty() && range.start == self.hp {
```
Dentosal, line 975: Emtpy range at `HP..HP` is a valid range, but that wasn't allowed previously.

fuel-vm/src/storage/blob_data.rs
```diff
@@ -0,0 +1,92 @@
...
+use rand::{
+    distributions::{
+        Distribution,
+        Standard,
+    },
+    Rng,
+};
+
+/// The storage table for blob data bytes.
+pub struct BlobData;
```
Dentosal, line 20: Unsure whether this should live here or in `fuel-core`. The pre-existing types are scattered around these two repos.

fuel-vm/src/storage/predicate.rs
```diff
@@ -40,6 +41,8 @@ use super::{
 /// operations. However, predicates, as defined in the protocol, cannot execute contract
 /// opcodes. This means its storage backend for predicate execution shouldn't provide any
 /// functionality.
+///
+/// TODO: blob storage should be implemented
```
Dentosal, line 45: Should we do this? Should we do this now? Mentioned as an open issue in the PR body.

fuel-vm/src/tests/alu.rs
Dentosal, line 1: The `AluResult` is moved to test_helpers, so the same `RunResult` type can be used by multiple tests

xgreenx:
@Dentosal Could you apply fix from https://github.com/FuelLabs/fuel-vm/pull/789 to your change here as well, please?=)

Dentosal:
> @Dentosal Could you apply fix from #789 to your change here as well, please?=)

Merged master into this, should be applied now.

fuel-vm/src/tests/blockchain.rs
```diff
@@ -2261,3 +2261,444 @@ fn various_ldc_issues_poc() {
         }
     });
 }
+
+// === Begin temporary tests for ldcv2 ===
+
+#[test]
+fn ldcv2__offset_affects_read_code() {
```
MitchTurner, line 2268: How does it affect it? Based on the test, it looks like as offset increases, the gas used decreases. G/W/T would help clarify the name and vice versa.

It looks like a lot of the tests in this suite are missing G/W/T and names that capture the specific intended behavior.

For example, `ldcv2_ssp_not_sp` doesn't say what should be `ssp`, or in what scenario, just that it's not `sp`. Shouldn't it be something like `ldcv2__when_increase_sp_...something`

fuel-vm/src/tests/test_helpers.rs
```diff
@@ -55,10 +55,19 @@ pub fn run_script(script: Vec<Instruction>) -> Vec<Receipt> {
 }
 
 /// Assert that transaction didn't panic
+#[track_caller]
```
MitchTurner, line 58: Why do we need this?

MitchTurner (commented):
Lots of little things. Nothing too big.

fuel-tx/src/transaction.rs
```diff
@@ -954,6 +1013,16 @@ pub mod field {
         fn bytecode_root_offset_static() -> usize;
     }
 
+    pub trait BlobId {
+        fn blob_id(&self) -> &fuel_types::BlobId;
+        fn blob_id_mut(&mut self) -> &mut fuel_types::BlobId;
```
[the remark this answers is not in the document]
Dentosal: Because every other trait here has it as well, is why I added it. Even the other traits seem to use `_mut` methods only for testing, maybe we should `#[cfg(tests)]` those as well?

fuel-tx/src/transaction/consensus_parameters/gas/default_gas_costs.rs
```diff
@@ -1,9 +1,9 @@
 use super::*;
 /// File generated by fuel-core: benches/src/bin/collect.rs:440. With the following git
 /// hash
-pub const GIT: &str = "98341e564b75d1157e61d7d5f38612f6224a5b30";
+pub const GIT: &str = "98341e564b75d1157e61d7d5f38612f6224a5b30"; // TODO: modified manually
```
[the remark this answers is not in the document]
Dentosal: We need to generate new values using `fuel-core` benches, but I'm not sure what the correct update order is. We cannot release `fuel-vm` without introducing new values, but should we run the benches against this branch before we publish it? Likely.

fuel-tx/src/transaction/types/blob.rs
```diff
@@ -0,0 +1,227 @@
...
+        gas_cost.s256().resolve(bytes as u64)
+    }
+}
+
+impl UniqueFormatValidityChecks for Blob {
+    fn check_unique_rules(
+        &self,
+        consensus_params: &ConsensusParameters,
+    ) -> Result<(), ValidityError> {
+        let index = self.body.witness_index as usize;
```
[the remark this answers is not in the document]
Dentosal: "subsection" and "bytecode" are just incorrect words here, anyway. In d42c469ef78617d6ca19ca3d447dece48dcbb23d I've changed the comment  to say

> The witness index of the payload.

Which is correct. Let me know if renaming still seems useful after this.

fuel-tx/src/transaction/types/blob.rs
```diff
@@ -0,0 +1,227 @@
...
+        let witness = self
+            .witnesses
+            .get(index)
+            .ok_or(ValidityError::InputWitnessIndexBounds { index })?;
+
+        // Verify that blob id is correct
```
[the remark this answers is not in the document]
Dentosal: None of the other `UniqueFormatValidityChecks` implementations does this splitting either, so I'd rather update them all in a single follow-up PR.

fuel-vm/src/interpreter/blob.rs
```diff
@@ -0,0 +1,89 @@
...
+        else {
+            return Err(PanicReason::BlobNotFound.into());
+        };
+        self.dependent_gas_charge_without_base(
+            self.interpreter_params.gas_costs.bsiz(),
+            size as Word,
+        )?;
+        let (SystemRegisters { pc, .. }, mut w) = split_registers(&mut self.registers);
+        let result = &mut w[WriteRegKey::try_from(dst)?];
+        *result = size as Word;
```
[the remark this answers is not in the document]
Dentosal: Naming the index would hurt readibility. The index cannot be used in any other context than when indexing into `ProgramRegisters`. The pattern `&mut program_registers[WriteRegKey::try_from(dst)?]` repeats everywhere, because that's the only sensible way to use `WriteRegKey`. Giving it a name would only be more confusing. That said, I'm pushing to get rid of `WriteRegKey` and `ProgramRegisters`, so we wouldn't have to do this.

Abbreviations are vital for code readibility. I cannot easily parse lines that have multiple long identifiers. Particularly things that get used all the time deserve a short name. The register names are used in every single instruction implementation, and having `program_counter` in every place where we have `pc` now would be awful. It's stored as `pc` in my head, I do not expand it every time.

`result` is the reference where the result is written to. What would be a better name? Maybe `result_register_value`?

The `w` is defined on the line above, and used once. Giving it a longer name would make things worse. Whice one you prefer?

```rust
let mut acc = 0;
for i in 0..limit {
  acc += values[i];
}
```

or

```rust
let mut accumulator = 0;
for values_index in 0..limit {
  accumulator += values[values_index];
}
```

fuel-vm/src/tests/blob.rs
```diff
@@ -0,0 +1,360 @@
...
+use test_case::test_case;
+
+use super::test_helpers::{
+    assert_success,
+    RunResult,
+};
+use crate::tests::test_helpers::set_full_word;
+
+#[rstest::rstest]
+fn blob_size_and_load_whole(
```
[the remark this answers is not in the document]
Dentosal: I'll look into separating them.

rstest is used to compress 16 lines into 2, significantly helping reability and reducing boilerplate.

fuel-vm/src/tests/blockchain.rs
```diff
@@ -2261,3 +2261,444 @@ fn various_ldc_issues_poc() {
         }
     });
 }
+
+// === Begin temporary tests for ldcv2 ===
```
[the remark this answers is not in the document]
Dentosal: Since the LDCv2 isn't supposed to exist for a long time, these tests are temporary. We're planning to migrate to a separate opcode instead of overloading ldc like this. This section should be removed when we separate the instructions, which is likely before merging this PR.

fuel-vm/src/tests/blockchain.rs
Dentosal: Really I just copied the tests for original LDC and adapted them when necessary. The original cases for LDC should be fixed as well.

The naming is indeed unclear, I'll try to fix those. I consider G/W/T unhelpful, but I can add format the tests that way if you want to.

fuel-vm/src/tests/test_helpers.rs
Dentosal: The `track_caller` attribute retargets panic messages to show the caller location by default, which is the desired behavior when using this function.

fuel-vm/src/interpreter/blob.rs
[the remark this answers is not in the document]
MitchTurner: Sorry. I think I can give a more helpful recommendation.

Let's just write a helper that hides all the abbreviations from the reader:

Instead of 
```
        let (SystemRegisters { pc, .. }, mut w) = split_registers(&mut self.registers);
        let result = &mut w[WriteRegKey::try_from(dst)?];
        *result = size as Word;
```
just declaratively say

```
      self.write_value_to_result_register(size)
```

An order of magnitude easier to read.

>The w is defined on the line above, and used once. Giving it a longer name would make things worse. Whice one you prefer?

`acc` means accumulator to me. `i` means index. `w` means nothing.

I prefer the first in your example, but the second isn't painful for me. If there was any business context in the example, then I'd prefer the latter.

WRT readabiliy, I always prefer a declarative function over having to read a bunch of implementation details.

fuel-tx/src/tests/valid_cases/transaction/blob.rs
```diff
@@ -0,0 +1,625 @@
...
+}
+
+#[test]
+fn valid_blob_transaction_can_pass_check() {
+    let block_height: BlockHeight = 1000.into();
+    let tx = valid_blob_transaction()
+        .finalize()
+        .check(block_height, &test_params());
+    assert_eq!(tx, Ok(()));
+}
```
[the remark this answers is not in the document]
MitchTurner: You mean it's adding more noise and you aren't convinced that the it is improving our tests?

Writing good tests takes discipline, requires developing good habits, and has an overhead, yes. It's an investment. Happy to hear alternatives to G/W/T, but we our tests in general are in a desperate need of more discipline.

fuel-tx/src/transaction.rs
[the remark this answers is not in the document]
MitchTurner: Yeah. That makes sense to me. Either that or call YAGNI.

fuel-tx/src/transaction/types/blob.rs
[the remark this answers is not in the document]
MitchTurner: That makes much more sense.

fuel-vm/src/tests/blockchain.rs
MitchTurner: > The naming is indeed unclear, I'll try to fix those. 

Cool. 

> I consider G/W/T unhelpful, but I can add format the tests that way if you want to.

From my perspective, each test should be only be testing one thing. I find G/W/T to be helpful to quickly parsing the tests to see what's being tested in that specific test. As I scroll down a test suite, I expect there to be many tests that do similar but subtly different things. 

`#[test_case]` also allows that, but I think a lot of engineers try to make it do too much in the name of DRY. So I prefer G/W/T clarity over `test_case` brevity.

fuel-tx/src/transaction.rs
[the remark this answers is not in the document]
Dentosal: Actually, that's a breaking change to modify those traits. I'm going to keep this as-is for now, but we can open a follow-up issue if you want to.

fuel-tx/src/tests/valid_cases/transaction/blob.rs
```diff
@@ -0,0 +1,625 @@
...
+
+#[test]
+fn maturity() {
```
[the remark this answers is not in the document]
Dentosal: The case above is the happy path test.

fuel-vm/src/tests/blockchain.rs
Dentosal: I added G/W/T where applicable. I'm not happy with that, especailly since the distinction between Given and When is quite tricky when data is code.

fuel-tx/src/transaction/consensus_parameters/gas.rs
Dentosal: Changed to VM level panic.

fuel-vm/src/interpreter/executors/instruction.rs
Dentosal: This should probably be reverted before merging, if we're not going to make use of is after all.

fuel-vm/src/storage/predicate.rs
Dentosal: TODO: link to an issue

fuel-vm/src/interpreter/blob.rs
Dentosal: TODO: review code to check if we actualy do this correctly, i.e. it really should depend on the size

fuel-tx/src/transaction/types/blob.rs
```diff
@@ -0,0 +1,235 @@
...
+impl BlobIdExt for BlobId {
+    fn compute(data: &[u8]) -> Self {
+        Self::new(*fuel_crypto::Hasher::hash(data))
+    }
+}
+
+pub type Blob = ChargeableTransaction<BlobBody, BlobMetadata>;
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct BlobMetadata;
```
xgreenx, line 45: I think it would be nice to cache the hash of the blob

fuel-tx/src/transaction/types/blob.rs
```diff
@@ -0,0 +1,235 @@
...
+    #[inline(always)]
+    fn metered_bytes_size(&self) -> usize {
+        Serialize::size(self)
+    }
+
+    #[inline(always)]
+    fn gas_used_by_metadata(&self, gas_cost: &GasCosts) -> Word {
+        let bytes = Serialize::size(self);
+        // Gas required to calculate the `tx_id`.
+        gas_cost.s256().resolve(bytes as u64)
```
xgreenx, line 79: And we can also charge for `BlobId` calculation=)

fuel-tx/src/transaction/validity/error.rs
```diff
@@ -90,6 +90,10 @@ pub enum ValidityError {
     TransactionOutputContainsContractCreated {
         index: usize,
     },
+    /// The output contains a `Output::Coin` which is not allowed.
+    TransactionOutputContainsCoin {
```
xgreenx, line 94: It is not described anywhere, but for backward compatibility, we need to add all new variants of errors at the end. I want to add a test later to ensure this behaviour.

But for now could you please add all new variants to the end?=)

fuel-tx/src/transaction/validity/error.rs
```diff
@@ -112,6 +116,8 @@ pub enum ValidityError {
     /// The total number of bytecode subsections in the `Upload` transaction exceeds the
     /// limit.
     TransactionUploadTooManyBytecodeSubsections,
+    /// Blob id of the transaction differs from the data.
+    TransactionBlobIdVerificationFailed,
```
xgreenx, line 120: The same here, the error should be at the end

fuel-tx/src/transaction/validity/error.rs
xgreenx: https://github.com/FuelLabs/fuel-vm/issues/797

fuel-vm/src/interpreter/blob.rs
```diff
@@ -0,0 +1,94 @@
...
+        let gas_cost = self
+            .interpreter_params
+            .gas_costs
+            .bldd()
+            .map_err(PanicReason::from)?;
+        self.dependent_gas_charge(gas_cost, len)?;
+
+        let blob_offset: usize = blob_offset
+            .try_into()
+            .map_err(|_| PanicReason::MemoryOverflow)?;
```
xgreenx, line 75: It produces different result for 32 and 64 bit platforms.

fuel-vm/src/interpreter/blob.rs
```diff
@@ -0,0 +1,94 @@
...
+        blob_id_ptr: Word,
+        blob_offset: Word,
+        len: Word,
+    ) -> IoResult<(), S::DataError> {
+        let gas_cost = self
+            .interpreter_params
+            .gas_costs
+            .bldd()
+            .map_err(PanicReason::from)?;
+        self.dependent_gas_charge(gas_cost, len)?;
```
xgreenx, line 71: We need to charge for maximum between `len` and size of the blob

fuel-vm/src/interpreter/blob.rs
```diff
@@ -0,0 +1,94 @@
...
+        let blob_id = BlobId::from(self.memory.as_ref().read_bytes(blob_id_ptr)?);
+        let owner = self.ownership_registers();
+        let dst = self.memory.as_mut().write(owner, dst_ptr, len)?;
+
+        let blob = <S as StorageInspect<BlobData>>::get(&self.storage, &blob_id)
+            .map_err(RuntimeError::Storage)?
+            .ok_or(PanicReason::BlobNotFound)?;
+        let blob = blob.as_ref().as_ref();
+
+        let end = blob_offset.saturating_add(dst.len()).min(blob.len());
```
xgreenx, line 86: The `blob_offset` could be `u32::MAX` that could cause different result for 32 and 64 bit platforms. We need to check that ensures that `blob_offset` less than blob size. In the code copy case we check that it is less than max contract size(What also could be incorrect, maybe we want to compare it with the size of the contract).

fuel-vm/src/interpreter/blob.rs
```diff
@@ -0,0 +1,94 @@
...
+        let data = blob.get(blob_offset..end).unwrap_or_default();
```
xgreenx, line 87: Could we replace `unwrap_or_default` with `unwrap_or` please?=) It would be simpler to read

fuel-vm/src/interpreter/blob.rs
xgreenx: Hmm, even if saturation occurs it will be handled by `min` later. But maybe it is still better to add this check to avoid potential problems in the future

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -661,6 +678,113 @@ where
...
+        let region_start = ssp;
+
+        if ssp != sp {
+            return Err(PanicReason::ExpectedUnallocatedStack.into())
+        }
+
+        let blob_id = BlobId::from(self.memory.read_bytes(blob_id_addr)?);
+        let blob_offset: usize = blob_offset
+            .try_into()
+            .map_err(|_| PanicReason::MemoryOverflow)?;
```
xgreenx, line 710: It produces different results on the 32-bit and 64-bit platforms. Attacker could trigger either `MemoryOverflow` or `BlobNotFound`.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -661,6 +678,113 @@ where
...
+
+        let current_contract = current_contract(self.context, self.fp, self.memory)?;
+
+        let length = bytes::padded_len_usize(
+            length_unpadded
+                .try_into()
+                .map_err(|_| PanicReason::MemoryOverflow)?,
+        )
+        .map(|len| len as Word)
+        .unwrap_or(Word::MAX);
```
xgreenx, line 720:
```suggestion
        let length = bytes::padded_len_word(length_unpadded).unwrap_or(Word::MAX);
```

It produces different results on the 32-bit and 64-bit platforms.

fuel-vm/src/interpreter/blob.rs
xgreenx: Hmm, the logic here looks like the logic inside of the `copy_from_slice_zero_fill`. Maybe we can use it here instead(and the same for code copy)?

fuel-vm/src/interpreter/executors/instruction.rs
xgreenx: Let's remove it from the current PR

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -731,6 +736,57 @@ where
...
+
+        let blob_id = blob.blob_id();
+
+        debug_assert_eq!(
+            BlobId::compute(blob_data.as_ref()),
+            *blob_id,
+            "Tx has invalid BlobId",
+        );
+
+        storage
```
xgreenx, line 769: We need to return an error, if the same blob id already exists.

fuel-vm/src/storage/predicate.rs
xgreenx: It is a separate feature and not part of this PR. If we don't have task for this feature, let's create it)

fuel-vm/src/tests/alu.rs
```diff
@@ -71,35 +71,20 @@ fn reserved_register() {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum AluResult {
-    Success { value: Word, of: Word, err: Word },
-    MissingLog,
-    Revert,
-    Panic(PanicReason),
-    GenericFailure(u64),
+pub struct AluOk {
```
xgreenx, line 74: Do we need to touch it in this PR?=)

fuel-vm/src/tests/encoding.rs
```diff
@@ -661,4 +661,120 @@ fn transaction_canonical_serialization_deserialization() {
...
+            Policies::new()
+                .with_tip(Word::MAX >> 1)
+                .with_maturity((u32::MAX >> 3).into())
+                .with_witness_limit(Word::MAX >> 4)
+                .with_max_fee(Word::MAX >> 5),
+            vec![],
+            vec![],
+            vec![],
+        ),
+    ]);
```
xgreenx, line 779:
```suggestion
```

It is the same as another part

fuel-tx/src/tests/offset.rs
```diff
@@ -530,6 +530,46 @@ fn tx_offset_upload() {
...
+#[test]
+fn tx_offset_blob() {
+    let mut cases = TestedFields::default();
+    let number_cases = 100;
+
+    // The seed will define how the transaction factory will generate a new transaction.
+    // Different seeds might implicate on how many of the cases we cover - since we
+    // assert coverage for all scenarios with the boolean variables above, we need to
+    // pick a seed that, with low number of cases, will cover everything.
+    TransactionFactory::<_, Upload>::from_seed(1295)
```
xgreenx:
```suggestion
    TransactionFactory::<_, Blob>::from_seed(1295)
```

fuel-tx/src/tests/offset.rs
```diff
@@ -530,6 +530,46 @@ fn tx_offset_upload() {
...
+        .take(number_cases)
+        .for_each(|(tx, _)| {
+            let bytes = tx.to_bytes();
+            chargeable_transaction_parts(&tx, &bytes, &mut cases);
```
xgreenx, line 546: It would be nice to check getters for `blob_id`.

fuel-tx/src/tests/valid_cases/transaction/blob.rs
```diff
@@ -0,0 +1,629 @@
+#![allow(clippy::cast_possible_truncation)]
+#![allow(non_snake_case)]
+
+use super::*;
```
xgreenx, line 4: I don't see the test for the case when blob id mismatch

fuel-tx/src/transaction/validity/error.rs
xgreenx: Also, we want to allow at least `Output::Coin` for the base asset. I think we don't need this error and this check. The `TransactionInputContainsNonBaseAssetId` error should be enough.

fuel-tx/src/transaction/validity/error.rs
xgreenx: We need a test case for this error=)

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#783](https://github.com/FuelLabs/fuel-vm/pull/783): Remove unnecessary look up for old values by adding new methods to the `StorageMutate` trait.  The old `insert` and `remove` are now `replace` and `take`. The new `insert` and `remove` don't return a value.
 - [#783](https://github.com/FuelLabs/fuel-vm/pull/783): Renamed methods of `StorageWrite` trait from `write`, `replace`, `take` to `write_bytes`, `replace_bytes`, `take_bytes`.
 - [#788](https://github.com/FuelLabs/fuel-vm/pull/788): Fix truncating `sp` to `MEM_SIZE` in `grow_stack`, and allow empty writes to zero-length ranges at `$hp`.
+- [#780](https://github.com/FuelLabs/fuel-vm/pull/780): Added `Blob` transaction, and `BSIZ` and `BLDD` instructions. Also allows `LDC` to load blobs.
```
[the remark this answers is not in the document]
Dentosal: 6fcafd41

fuel-tx/src/tests/valid_cases/transaction/blob.rs
Dentosal: 6fcafd41

fuel-vm/src/tests/encoding.rs
Dentosal: 6fcafd41 (the one above was duplicated as well)

fuel-tx/src/transaction/validity/error.rs
Dentosal: 6fcafd41

fuel-vm/src/interpreter/blob.rs
Dentosal: 6fcafd41

fuel-vm/src/interpreter/blob.rs
Dentosal: Reading the spec, there's no need for the offset to be less than the contract size. Fixes done to other opcodes (CCP, LDC) are incorrect (against the spec). The spec doesn't mention panicing even If the offset is larger than the size. Is the spec wrong here? It would feel weird to special-case fitting into `u32`.

fuel-tx/src/transaction/validity/error.rs
Dentosal: ac2776ac changes the check to do that

fuel-tx/src/transaction/types/blob.rs
Dentosal: ac2776ac

fuel-vm/src/interpreter/executors/main.rs
Dentosal: Done in d38fc654

fuel-vm/src/interpreter/blob.rs
Dentosal: I fixed all those in 34107946.

fuel-vm/src/interpreter/blockchain.rs
Dentosal: 34107946

fuel-vm/src/interpreter/blob.rs
Dentosal: c723f3d3

fuel-tx/src/transaction/types/blob.rs
Dentosal: I don't think we can ever use the cached value for anything. The only use of the hash is to be a nonce.

fuel-vm/src/tests/alu.rs
Dentosal: The `RunResult` changes must be done here, and these are a natural consequence of those. It would be weird not to do them here.

fuel-tx/src/transaction/consensus_parameters/gas/default_gas_costs.rs
Dentosal: https://github.com/FuelLabs/fuel-core/issues/2049

fuel-vm/src/storage/predicate.rs
Dentosal: https://github.com/FuelLabs/fuel-vm/issues/798

fuel-tx/src/transaction/types/blob.rs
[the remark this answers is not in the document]
Dentosal: https://github.com/FuelLabs/fuel-vm/issues/800

fuel-tx/src/tests/offset.rs
Dentosal: 4fe7322e

fuel-tx/src/tests/offset.rs
```diff
@@ -530,6 +539,53 @@ fn tx_offset_upload() {
...
+    // The seed will define how the transaction factory will generate a new transaction.
+    // Different seeds might implicate on how many of the cases we cover - since we
+    // assert coverage for all scenarios with the boolean variables above, we need to
+    // pick a seed that, with low number of cases, will cover everything.
+    TransactionFactory::<_, Blob>::from_seed(1295)
+        .take(number_cases)
+        .for_each(|(tx, _)| {
+            let bytes = tx.to_bytes();
+
+            let i = *tx.bytecode_witness_index();
```
xgreenx, line 556: We need to read the value from the offset=)

fuel-tx/src/transaction/types/blob.rs
```diff
@@ -0,0 +1,250 @@
...
+                }
+                Output::Coin { asset_id, .. } => {
+                    if asset_id != consensus_params.base_asset_id() {
+                        Err(ValidityError::TransactionInputContainsNonBaseAssetId {
+                            index,
+                        })
+                    } else {
+                        Ok(())
+                    }
+                }
```
xgreenx, line 166: We don't need this branch because combination of the `TransactionOutputCoinAssetIdNotFound` and `TransactionInputContainsNonBaseAssetId` guarantees it=)

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -731,6 +736,61 @@ where
...
+        );
+
+        let old = storage
+            .storage_as_mut::<BlobData>()
+            .replace(blob_id, blob_data.as_ref())
+            .map_err(RuntimeError::Storage)?;
+
+        if old.is_some() {
+            return Err(InterpreterError::Panic(PanicReason::BlobIdAlreadyUploaded));
+        }
```
xgreenx, line 776: Maybe it will be cheaper to use `contains` + `insert` instead of `replace`. But it can be done in a separate PR/Issue, because other places have the same problem.

fuel-vm/src/tests/blockchain.rs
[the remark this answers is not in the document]
xgreenx: Hmm, I think we can remove it now, it seems old LDC for blobs will exist along with new LDCv2 that we plan to use for contracts. I don't like it, but because we need a backward compatibility with the testnetn, we can't drop it when we move do mainnet=(

fuel-vm/src/tests/blockchain.rs
```diff
@@ -2261,3 +2261,444 @@ fn various_ldc_issues_poc() {
...
+    let receipts = result.receipts();
+    let ret = receipts
+        .first()
+        .expect("A `RET` opcode was part of the program.");
+
+    assert_eq!(1, ret.val().expect("Return value"));
+}
+
+#[test]
+fn load_blob_code__fails_when_blob_offset_is_over_length() {
```
xgreenx, line 2652: Hmm, this test doesn't fail=) Maybe we need to change the name

fuel-vm/src/tests/blockchain.rs
```diff
@@ -2261,3 +2261,444 @@ fn various_ldc_issues_poc() {
...
+        op::xor(reg_b, reg_b, reg_b),          // r[b] := 0
+        op::ori(reg_b, reg_b, 12),             // r[b] += 12 (will be padded to 16)
+        op::ldc(reg_a, RegId::ZERO, reg_b, 1), // Load first two words from the contract
+    ];
+
+    ldcv2_reason_helper(load_blob, MemoryOverflow);
+}
+
+#[test]
+fn ldcv2_fails_when_blob_not_in_inputs() {
```
xgreenx, line 2587: We don't require the blob to be in inputs. What are we testing here?

Dentosal: It's test for nonexistent blob, but named incorrectly. I'll fix.

Dentosal: 02df199a

fuel-vm/src/interpreter/executors/main.rs
Dentosal: Maybe? Also should be profiled if we're doing anything for performance reasons.

fuel-tx/src/tests/offset.rs
Dentosal: Right. Fixed in 7ebac9e5.

fuel-vm/src/tests/blockchain.rs
Dentosal: Renamed in 7ebac9e5. I'd like to have a bit more exact test for this. This should be improved elsewhere as well, I'll do it as part of https://github.com/FuelLabs/fuel-vm/issues/746

MitchTurner (approved):

PR #781, opened 2024-06-21
Store the base asset id in the metadata

xgreenx:
Having the base asset id allows to get the base asset balance by just using metadata without consensus parameters.

### Before requesting review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [#781](https://github.com/FuelLabs/fuel-vm/pull/781): Added `base_asset_id` to checked metadata.
+
 ## [Version 0.54.1]
 
 ### Changed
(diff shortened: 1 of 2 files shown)


MitchTurner (commented):
Why do we need to get the asset id without consensus params? 

This seems like a harmless change, but it's also redundant, so I just want to understand the justification. It's not like the base asset will be different for txs in a block or likely ever change for the chain.

xgreenx:
It is possible to change the base asset id via consensus parameters.

While checked transactions live inside of the memory pool, consensus parameters could change. When we try to ask for a transaction about its base asset balance, it will return the wrong value. 

Another reason for the change is to make the `Checked` transaction self-contained. You can request the base balance without providing the consensus parameters.

Taking into account that each time to get the latest consensus parameters, we need to lock and clone the consensus parameters from the cache, it will not affect performance.

MitchTurner:
But we don't want to let a tx submitter just choose some arbitrary base asset. I guess we are going to still check that the base asset is correct for the block when we call `into_ready`, maybe?

xgreenx:
The TX submitter can't choose it. We cached the value from consensus parameters. The `Checked` transactions created within the node=)

In the same way, we could just case `base_asset_amount`. But having a base asset it is more flexible.

MitchTurner (approved):
Looks good assuming we implement the things we talked about offline.

PR #784, opened 2024-07-01
Optimized storage lookups for side nodes in the SMT

xgreenx:
During debugging high storage access in the testnet, I realized that we request left and right children during iteration over the SMT.

image

While we only need path nodes. When we construct the node back, we can use the same path node to decide who is the right child and who is the left child. In this case we reduce number of storage lookups in two times.

## Checklist

### Before requesting review

### DIFF
--- a/fuel-merkle/src/sparse/merkle_tree.rs
+++ b/fuel-merkle/src/sparse/merkle_tree.rs
@@ -229,20 +229,21 @@ where
     fn path_set(
         &self,
         leaf_key: &Bytes32,
-    ) -> Result<(Vec<Node>, Vec<Node>), MerkleTreeError<StorageError>> {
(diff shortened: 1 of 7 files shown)


fuel-merkle/src/sparse/merkle_tree.rs
```diff
@@ -524,14 +521,27 @@ where
         }
 
         // Merge side nodes
-        for side_node in side_nodes {
-            current_node = Node::create_node_on_path(path, &current_node, side_node);
+        for (side_node, old_parent) in
+            side_nodes.iter().zip(path_nodes.iter().skip(1 /* leaf */))
```
Dentosal, line 525: nit: Moving `path_nodes.iter().skip(1)` to a separate line with a let binding might make this more readable.

Dentosal (dismissed):

Dentosal (approved):

PR #787, opened 2024-07-01
Fixed charge functions to profile cost before charging

xgreenx:
Changed the behavior of the gas charge functions.

Form:
1. Charge
2. Profile

To :
1. Profile
2. Charge

Because the `profile` function expects that it is called before `$cgas` is updated.

image

Since we don't this code, it is not breaking behavior=)

### Before requesting review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - [#781](https://github.com/FuelLabs/fuel-vm/pull/781): Added `base_asset_id` to checked metadata.
 
+### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#781](https://github.com/FuelLabs/fuel-vm/pull/781): Added `base_asset_id` to checked metadata.
 
 ### Changed
+- [#787](https://github.com/FuelLabs/fuel-vm/pull/787): Fixed charge functions to profile cost before charging.
```
Dentosal, line 14: Technically a breaking change, but I don't think anybody cares about this

Dentosal (approved):
Nice simplification

PR #788, opened 2024-07-04
grow_stack and empty $hp range write fixes

Dentosal:
Fixes two minor bugs in the memory implementation:
- `grow_stack` was silently truncating the `sp` to `MEM_SIZE` even if it was above that
- Empty ranges at `$hp` would fail ownership check, even if they really should pass

## Checklist
 * The `grow_stack` case is not tested, as it's not really reachable externally.

### Before requesting review

### DIFF
--- a/fuel-vm/src/interpreter/memory.rs
+++ b/fuel-vm/src/interpreter/memory.rs
@@ -977,11 +977,10 @@ impl OwnershipRegisters {
 
     /// Empty range is owned iff the range.start is owned
     pub(crate) fn has_ownership_heap(&self, range: &Range<Word>) -> bool {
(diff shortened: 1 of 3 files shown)


fuel-vm/src/interpreter/memory.rs
```diff
@@ -972,11 +977,10 @@ impl OwnershipRegisters {
 
     /// Empty range is owned iff the range.start is owned
     pub(crate) fn has_ownership_heap(&self, range: &Range<Word>) -> bool {
-        // TODO implement fp->hp and (addr, size) validations
```
Dentosal, line 975: These were already implemented, so this comment was obsolete.

Voxelot (approved):

PR #804, opened 2024-08-21
Rewrite fuel_asm::impl_instructions! to a proc macro

Dentosal:
What it says on the tin. No functional changes, just refactoring.

The purpose of this PR is to allow more complex instruction definitions in the future. This should allow moving some validity checks to instruction parsing level, and to improve type safety.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog (None!!)
- [x] New behavior is reflected in tests (No new behavior!)

### Before requesting review


fuel-derive/src/asm.rs
```diff
@@ -0,0 +1,1006 @@
...
+                    .collect();
+
+                let pack_strict_arguments: TokenStream = args
+                    .iter()
+                    .enumerate()
+                    .map(|(i, arg)| {
+                        let name = &arg.name;
+                        if arg.is_imm() {
+                            quote! {
+                                packed_integer |= (#name.to_smallest_int() as u32);
```
MitchTurner, line 382: Places like this would be better with a single-use helper function.

fuel-asm/derive/src/codegen.rs
```diff
@@ -0,0 +1,507 @@
...
+                match value {
+                    #arms
+                    _ => Err(InvalidOpcode),
+                }
+            }
+        }
+    }
+}
+
+pub fn from_op(instructions: &InstructionList) -> TokenStream {
```
netrome, line 341: I like how most other codegen functions are named with the type they are implementing the behavior on first, e.g `op_try_from`, `op_debug_impl`, `instruction_try_from_bytes`. Therefore it would be nice to name this function in such a way that it leads with `op`.

I understand that `op_from` may sound like conversion methods that create opcodes from other types, so perhaps `op_into` would be a better name. Or just `op_conversions`.

I'm okay leaving this as-is as well. What are your thoughts?

fuel-asm/Cargo.toml
```diff
@@ -21,6 +22,7 @@ wasm-bindgen = { version = "0.2.88", optional = true }
 [dev-dependencies]
 bincode = { workspace = true }
 fuel-asm = { path = ".", features = ["serde"] }
+proptest = "1.5"
```
netrome, line 25: Nice with property based testing!

netrome (approved):
Nice stuff! While I still find macros hard to read in general, the code after the refactor is much easier to follow than the previous implementation of this macro. Thanks for making this improvement!

fuel-derive/src/asm.rs
```diff
@@ -0,0 +1,1006 @@
...
+                let pack_test_arguments: TokenStream = args
+                    .iter()
+                    .enumerate()
+                    .map(|(i, arg)| {
+                        let reg_name = Ident::new(&format!("reg{i}"), Span::call_site());
+                        match arg.typeinfo() {
+                            ArgType::Imm(bits) =>{
+                                let bits: u32 = bits.try_into().expect("Type size is checked");
+                                quote! {
+                                    packed_integer |= imm & ((#bits << 1u32) -1);
```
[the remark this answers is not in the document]
Dentosal: Just removed the whole `test_construct` instead and rewrote the test case using it.

fuel-derive/src/asm.rs
```diff
@@ -0,0 +1,1006 @@
...
+                        }
+                    })
+                    .collect();
+
+                let check_flexible_arguments: TokenStream = args
+                    .iter()
+                    .map(|arg| if arg.is_imm() {
+                        let name = &arg.name;
+                        let type_ = &arg.type_;
+                        quote! { #type_::new_checked(#name).expect("Immediate value overflows"), }
```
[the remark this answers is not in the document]
Dentosal: That would be a breaking change, leaving the change for a follow-up PR https://github.com/FuelLabs/fuel-vm/issues/807

fuel-derive/src/asm.rs
```diff
@@ -0,0 +1,1006 @@
...
+    }
+}
+
+/// Constructor functions and theirs shorthands
+fn make_constructors(instructions: &InstructionList) -> TokenStream {
+    instructions
+        .0
+        .iter()
+        .map(
+            |Instruction {
```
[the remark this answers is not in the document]
Dentosal: I've simplified the code quite a bit and each map has only a small block of code inside. Let me know if you still feel we need to have the closures as separate functions as well.

fuel-asm/derive/src/codegen.rs
Dentosal: That's a good point. Renamed to `op_conversions` in 43172b1c

fuel-derive/src/asm.rs
Dentosal: Moved most of these to smaller functions.

netrome (approved):
LGTM

AurelienFT:
@xgreenx Given the fact that we are now open to big refactor like this one : https://github.com/FuelLabs/fuel-vm/pull/904. Can we consider to merge this PR ?

xgreenx:
> Given the fact that we are now open to big refactor like this one : https://github.com/FuelLabs/fuel-vm/pull/904. Can we consider to merge this PR ?

The PR that you linked is pure refactoring, where we moved one part of the code to another part of the code. While in this PR, we re-implement the macro rule via the procedure macro. It is newly written logic.

We need to carefully review the implementation and compare the old expanded code with the new one to be sure that everything is correct.

Also, maybe we want to have an audit of this re-writing.

I still need to think about the main benefits of using procedural macro vs macro rule. I like the macro rule more because you can debug in your IDE there, but you can't do the same for the generated code from the procedure macro. But at the same time, the macro rule requires a lot of computation, and sometimes, my IDE freezes evaluating our macro rule for opcodes.

AurelienFT:
Okey, fair.

Dentosal:
> I still need to think about the main benefits of using procedural macro vs macro rule. I like the macro rule more because you can debug in your IDE there, but you can't do the same for the generated code from the procedure macro. But at the same time, the macro rule requires a lot of computation, and sometimes, my IDE freezes evaluating our macro rule for opcodes.

We could instead publish the code produced by expanding the macro, and not ship a proc macro at all. The produced code is reasonably small and shouldn't have any ot these issues.

I find the declarative macro to be unreadable and error-prone. As long as we don't change anything it's fine though.

Dentosal:
That all said, I'm closing this and https://github.com/FuelLabs/fuel-vm/pull/805 for now. We can reopen if we need to modify the macros in the future.

PR #808, opened 2024-08-27
DA compression: TxId → TxPointer

Dentosal:
Follow-up to #670.

Adds `TxId` → `TxPointer` compression to the DA compression pipeline. This shrinks `UtxoId` from 34 to 8 bytes.

Due to splitting types into multiple crates (`fuel-compression`, `fuel-types`, `fuel-tx`) this gets a bit ugly, as the latter two have to depend on `fuel-compression` to access the compression traits, which in turn have to access the types from both of the other crates. The current version just derefs the types to raw byte arays, but maybe moving all of these types to `fuel-types` would be better.

## Checklist

### Before requesting review


Dentosal:
Merged manually.

PR #810, opened 2024-09-02
add `Unknown` transaction variant

hal3e:
https://github.com/FuelLabs/fuels-rs/issues/1477

BREAKING: Added `Unknown` transaction variant.

## Checklist

### Before requesting review

- [ ] Someone else?


fuel-tx/src/transaction.rs
```diff
@@ -102,6 +103,7 @@ pub enum Transaction {
     Upgrade(Upgrade),
     Upload(Upload),
     Blob(Blob),
+    Unknown,
```
xgreenx, line 106: This kind of change should be done differently. From the protocol level, we don't have unknown transactions since it is impossible to do a state transition if the transaction is unknown.

A new variant should be handled by the code that may receive a new variant of the enum. Maybe you want to look into `fuel-core-client` since it is where you fetch transactions from the GraphQL.

xgreenx (changes requested):

fuel-tx/src/transaction.rs
hal3e: Thanks for the clarification.

PR #816, opened 2024-09-06
Simplifying tests for DA compression

xgreenx:
Small improvements for test maintenance


Dentosal (approved):
I don't like introducing syntax magic trait, but it's probably better than the alternative here.

PR #822, opened 2024-09-12
Return recipient as an owner for the message inputs

xgreenx:
## Checklist


AurelienFT (approved):

netrome (approved):

acerone85:
Is there an issue that explains what the bug is?

I had a look at the code and LGTM (won't approve as 1 codeowner approval should be required next per approval guidelines)

PR #824, opened 2024-09-12
Use `self` instead of `&self` during decompression

xgreenx:
Just move `self` during deserialization since in most cases we will do nothing with it.


netrome (approved):
Looks correct to me. However, it is not obvious to me what the benefit of doing this are. What do we gain by taking ownership/moving the compressed value as opposed to reading it over a reference?

The only obvious improvement I can see is for `PoliciesBits` where we skip a redundant dereferencing operation, but for `UtxoId` and `impl_substitutable_key` we need to create new references within the functions instead.

xgreenx:
It has more sense in the case of messages and coins(predicate data for example, where we just move vector instead of cloning it).

netrome:
> It has more sense in the case of messages and coins(predicate data for example, where we just move vector instead of cloning it).

Alright, makes sense. Thanks for clarifying!

rafal-ch (approved):
LGTM :+1:

Dentosal (dismissed):

netrome (approved):

Dentosal (approved):

PR #825, opened 2024-09-12
Deserialize arrays: avoid leaking partially allocated memory when failing

### LINKED ISSUE #811: Handle error in the cases with `MaybeUninit`
During decoding arrays, we use `MaybeUninit`.  If we get an error during this process, we will have a memory leak since `MaybeUninit` doesn't call drop on internal objects.

image

acerone85:
Closes #811

* Minor changes to the logic for deserializing arrays, using a [MaybeUninit<T>; N] instead of MaybeUninit<[T]; N>. 
* Dropping initialized values when failing and the array has been partially initialized

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/fuel-compression/src/key.rs
+++ b/fuel-compression/src/key.rs
@@ -44,7 +44,7 @@ impl TryFrom<u32> for RegistryKey {
     fn try_from(value: u32) -> Result<Self, Self::Error> {
         let v = value.to_be_bytes();
         if v[0] != 0 {
-            return Err("RegistryKey must be less than 2^24")
+            return Err("RegistryKey must be less than 2^24");
         }
 
         let mut bytes = [0u8; 3];
--- a/fuel-compression/src/key.rs
+++ b/fuel-compression/src/key.rs
@@ -44,7 +44,7 @@ impl TryFrom<u32> for RegistryKey {
     fn try_from(value: u32) -> Result<Self, Self::Error> {
         let v = value.to_be_bytes();
         if v[0] != 0 {
-            return Err("RegistryKey must be less than 2^24");
+            return Err("RegistryKey must be less than 2^24")
         }
 
         let mut bytes = [0u8; 3];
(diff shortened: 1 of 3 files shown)


fuel-types/src/canonical.rs
```diff
@@ -430,23 +430,40 @@ impl<const N: usize, T: Deserialize> Deserialize for [T; N] {
...
+                    }
                 }
-                // SAFETY: Point to the next element after every iteration.
-                // 		 We do this N times therefore this is safe.
-                ptr = unsafe { ptr.add(1) };
             }
+            let final_ptr = &mut uninit as *mut _ as *mut [T; N];
             // SAFETY: All array elements have been initialized above.
-            let init = unsafe { uninit.assume_init() };
+            let init = unsafe { final_ptr.read() };
```
acerone85, line 465: This is ugly, but `mem::transmute` cannot be used for arrays with generic length. 

See https://github.com/rust-lang/rust/issues/61956
(The solution advocated there also mentions calling `mem::forget(uninit)`, but I don't think in this case is necessary and apparently it causes a compilation error.

fuel-types/src/canonical.rs
```diff
@@ -430,23 +430,40 @@ impl<const N: usize, T: Deserialize> Deserialize for [T; N] {
         } else {
             // Spec doesn't say how to deserialize arrays with unaligned
             // primitives(as `u16`, `u32`, `usize`), so unpad them.
-            let mut uninit = <MaybeUninit<[T; N]>>::uninit();
+            let mut uninit: [MaybeUninit<T>; N] =
```
Dentosal, line 433: May I suggest this pattern instead: https://github.com/FuelLabs/fuel-vm/blob/dfab064ad084115d4d2cdd0191c28d71abe8679d/fuel-compression/src/impls.rs#L128-L154

acerone85: Done in 741007fd1

fuel-compression/src/key.rs
```diff
@@ -44,7 +44,7 @@ impl TryFrom<u32> for RegistryKey {
     fn try_from(value: u32) -> Result<Self, Self::Error> {
         let v = value.to_be_bytes();
         if v[0] != 0 {
-            return Err("RegistryKey must be less than 2^24");
+            return Err("RegistryKey must be less than 2^24")
```
xgreenx, line 47:
```suggestion
            return Err("RegistryKey must be less than 2^24");
```

xgreenx (approved):

AurelienFT (approved):

PR #829, opened 2024-09-17
Add `fn add_fee_input()` and make `fn add_random_fee_input()` truly random

rafal-ch:
Previously the `add_random_fee_input()` wasn't actually producing a random input, which was confusing. This PR:
1. adds the `add_fee_input()` which is an equivalent to the former version of `add_random_fee_input()` 
2. modifies the `add_random_fee_input()` to accept an `rng` and provide truly random fee input

For compatibility reasons, all tests that were previously using `add_random_fee_input()` were modified to use `add_fee_input()`.

This change should make the test code more explicit and make it easier to create random transactions in test code.

### Before requesting review
- [X] I have created follow-up issues caused by this PR and linked them here

Follow-up draft PR in `fuel-core`: https://github.com/FuelLabs/fuel-core/pull/2211

### DIFF
--- a/fuel-tx/src/builder.rs
+++ b/fuel-tx/src/builder.rs
@@ -68,6 +68,11 @@ use fuel_types::{
     Salt,
     Word,
 };
+#[cfg(feature = "rand")]
+use rand::{
+    rngs::StdRng,
+    Rng,
+};
 
 pub trait BuildableAloc
 where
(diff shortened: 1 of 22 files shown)


xgreenx (approved):

MitchTurner (approved):

fuel-tx/src/builder.rs
```diff
@@ -408,7 +413,18 @@ impl<Tx: Buildable> TransactionBuilder<Tx> {
...
+            SecretKey::random(rng),
+            rng.gen(),
+            u32::MAX as u64,
+            *self.params.base_asset_id(),
+            Default::default(),
+        )
+    }
+
+    #[cfg(feature = "rand")]
+    pub fn add_fee_input(&mut self) -> &mut Self {
```
acerone85, line 427: can this be simply 

```rust
self.add_fee_input(rand::rngs::StdRng::seed_from_u64(2322u64))
```
?

acerone85:
one minor comment, but otherwise LGTM

fuel-tx/src/builder.rs
rafal-ch: Yes, that would work. Since it's not critical, I'll consider updating this part with next PR to `fuel-vm`.

xgreenx:
@rafal-ch This change is breaking, so it should modify the changelog

rafal-ch:
> This change is breaking, so it should modify the changelog

I created a follow-up PR: https://github.com/FuelLabs/fuel-vm/pull/836

PR #833, opened 2024-09-23
Add `Display` and `Error` impls to custom error type `SettingBlockTransactionSizeLimitNotSupported`

rafal-ch:
This PR adds the `std::error::Error` implementations for the recently introduced `SettingBlockTransactionSizeLimitNotSupported` custom error type.

It also adds the derived `Display`.

This will make it easier for the consumers to properly use this error type.

### Before requesting review

### DIFF
--- a/fuel-tx/src/transaction/consensus_parameters.rs
+++ b/fuel-tx/src/transaction/consensus_parameters.rs
@@ -24,6 +24,14 @@ const MAX_SIZE: u64 = 110 * 1024;
 
 #[derive(Debug)]
 pub struct SettingBlockTransactionSizeLimitNotSupported;
+#[cfg(feature = "std")]
+impl std::fmt::Display for SettingBlockTransactionSizeLimitNotSupported {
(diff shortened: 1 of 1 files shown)


netrome (approved):

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -24,6 +24,14 @@ const MAX_SIZE: u64 = 110 * 1024;
 
 #[derive(Debug)]
 pub struct SettingBlockTransactionSizeLimitNotSupported;
+#[cfg(feature = "std")]
+impl std::fmt::Display for SettingBlockTransactionSizeLimitNotSupported {
```
rafal-ch, line 28: 1. use `core` instead of `std`
2. check if we can derive `Display` here

rafal-ch: Updated in https://github.com/FuelLabs/fuel-vm/pull/833/commits/1a1182436b235a27dce235171cadafe99b4e6dc8

xgreenx (approved):

netrome (approved):

PR #837, opened 2024-09-26
Rollback memory changes

xgreenx:
This PR change the function `diff` that allowed to get changes between two arbitrary VM instances to a `rollback_to` function that allow to pass from current VM state to a previous one.

This restriction allow us to manage the memory that grows between instances because the current system wasn't working with memory that could grow but only with fixed size memory.

## Checklist

### Before requesting review

- [ ] Someone else?


fuel-vm/src/interpreter/memory.rs
```diff
@@ -359,6 +343,120 @@ impl MemoryInstance {
...
+    global_start: usize,
+    data: Vec<u8>,
+}
+
+/// The container for the data used to rollback memory changes.
+#[derive(Debug, Clone)]
+pub struct MemoryRollbackData {
+    /// Desired stack pointer.
+    sp: usize,
+    /// Desired heap pointer. Desired heap pointer can't be less than the current one.
```
AurelienFT, line 454: can't be equal or more ?

fuel-vm/src/interpreter/memory.rs
```diff
@@ -359,6 +343,120 @@ impl MemoryInstance {
...
+
+        assert!(
+            hp >= self.hp,
+            "We only allow shrinking of the heap during rollback"
+        );
+
+        let stack_changes =
+            get_changes(&self.stack[..sp], &desired_memory_state.stack[..sp], 0);
+
+        let heap_start = hp
```
AurelienFT, line 368: This HP is the one from `desired_memory_state` and we use the `self.heap_offset()` to get heap start point. We should use HP from `self`, no ?

xgreenx: The heap can only grow, so `HP` register can only decrease. Here `desired_memory_state` is a memory before we did `grow_heap`, so the `HP` is higher than `self.HP`. We want to subtract `self.heap_offset()` from `desired_memory.HP` to get the offset from which we want to read `self.heap` because `heap` and `desired_heap` should have the same size and should correspond to the same memory location to actually calculate diff.

So the `heap_start` represent part of the memory which we want to skip from the diff, because it is part of the memory allocated by `grow_heap` function

fuel-vm/src/interpreter/memory.rs
xgreenx: Remember that heap only grows -> `HP` only decreases during `grow_heap`, so if we rollback to old state, `DESIRED_HP(or HP_BEFORE_GROW_HEAP) >= HP_AFTER_GROW_HEAP`

xgreenx (commented):
Approved=)

AurelienFT (approved):

Voxelot (approved):
Took me a while to wrap my head around all the heap pointer magic, but lgtm after writing a few of my own tests to sanity check.

PR #838, opened 2024-09-26
Implemented `AsRef<[u8]>` and `TryFrom<&[u8]>` for DA compression types

xgreenx:
Implemented `AsRef<[u8]>` and `TryFrom<&[u8]>` for DA compression types: ScriptCode, PredicateCode, RegistryKey. It allows to use `Raw` codec in the `fuel-core`

### Before requesting review

### DIFF
--- a/fuel-compression/src/key.rs
+++ b/fuel-compression/src/key.rs
@@ -38,6 +38,7 @@ impl RegistryKey {
         }
     }
 }
+
 impl TryFrom<u32> for RegistryKey {
     type Error = &'static str;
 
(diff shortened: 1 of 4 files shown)


Dentosal (approved):

fuel-compression/src/key.rs
```diff
@@ -53,6 +54,26 @@ impl TryFrom<u32> for RegistryKey {
...
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        if value.len() != Self::SIZE {
+            return Err("RegistryKey must be 3 bytes long");
+        }
+
+        let mut bytes = [0u8; 3];
+        bytes.copy_from_slice(value);
+        Ok(Self(bytes))
+    }
+}
```
netrome, line 69: Nit: How about using a dedicated error type here?
```suggestion
impl TryFrom<&[u8]> for RegistryKey {
    type Error = InvalidRegistryKeySize;

    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
        if value.len() != Self::SIZE {
            return Err(InvalidRegistryKeySize);
        }

        let mut bytes = [0u8; 3];
        bytes.copy_from_slice(value);
        Ok(Self(bytes))
    }
}

#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, derive_more::Display, derive_more::Error)]
#[display("RegistryKey must be 3 bytes long")]
struct InvalidRegistryKeySize;
```

netrome (approved):
LGTM. Just one nit suggestion.

Issue #832, opened 2024-09-22
Remove `serde` feature from the `fuel-tx` crate


xgreenx:

MitchTurner:
WHAT IS THIS!?!?!

PR #843, opened 2024-09-27
Remove `serde` feature from the `fuel-tx` crate

xgreenx:
Closes https://github.com/FuelLabs/fuel-vm/issues/832

After we introduced the `Upgrade` transaction and started to use postcard to encode the consensus parameters, we basically made the `serde` feature required. 

Part of our codebase already derived `serde::Seriliaze` and `serde::Deserialize` regardless of the `serde` feature. This PR removes the `serde` feature and enables it for the `fuel-tx` crate by default under the `alloc` feature.

## Checklist

### Before requesting review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           - command: check
             args: --target wasm32-unknown-unknown -p fuel-crypto --no-default-features
           - command: check
-            args: --target wasm32-unknown-unknown -p fuel-tx --features serde --no-default-features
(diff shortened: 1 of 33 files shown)


rymnc (approved):
LGTM, curious if this was tested with the network upgrader which we ran into issues with serde? :)

xgreenx:
No, I haven't tested it, but since we added assert into network ugprader itself, we will see if something is wrong=)

Dentosal (approved):

PR #846, opened 2024-10-01
feat: Add intentional panic to check if clusterfuzzlite finds it

netrome:


fuel-vm/src/interpreter/alu.rs
```diff
@@ -105,14 +105,8 @@ where
...
-    if let Ok(expo) = u32::try_from(c) {
-        Word::overflowing_pow(b, expo)
-    } else if b < 2 {
-        (b, false)
-    } else {
-        (0, true)
-    }
+pub(crate) fn exp(_b: Word, _c: Word) -> (Word, bool) {
+    panic!("Can you find me, mr. Fuzz?");
 }
```
xgreenx, line 110: Can we try something like?

```suggestion
     if let Ok(expo) = u32::try_from(c) {
         Word::pow(b, expo).unwrap()
     } else if b < 2 {
         (b, false)
     } else {
         (0, true)
     }
```

xgreenx (commented):
Looks like it works=) Jus curious about another use case=)

fuel-vm/src/interpreter/alu.rs
netrome: Sure thing, let's try it :)

netrome: Task failed successfully
image
image

xgreenx: Cool!=) I think we can close this PR

netrome: Done

PR #847, opened 2024-10-04
Optimize `interpreter::blockchain::{load_contract_code, code_copy}` to read contract starting from offset

### LINKED ISSUE #681: Use methods from `StorageRead` trait in the LDC and copy code opcodes
We need to remove `storage_contract` method and update all places to use `StorageRead::read` method instead.

image

acerone85:
[Link to related issue(s) here, if any]

Closes #681 

When loading the contract code into the stack, we use the new semantics of `StorageRead` from #863 to only fetch the portion of a contract starting from the specified offset. We also read the contract bytes directly into the stack.

A similar optimization is done for `code_copy`.

## Checklist

### Before requesting review

- [ ] Someone else?


fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -1112,13 +1132,18 @@ impl<'vm, S> CodeRootCtx<'vm, S> {
             self.gas_cost,
             len as u64,
         )?;
-        let root = self
-            .storage
-            .storage_contract(&contract_id)
+        // TODO: Can we write this to the stack and avoid allocation?
```
netrome, line 1135: Since we don't know `len` statically I don't think we can put this on the stack, unless we want to over-allocate if we have a upper bound of the contract size we can use.

```suggestion
```

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -994,9 +1004,19 @@ where
...
+        let mut contract_buffer: Vec<u8> = alloc::vec![0u8; contract_len as usize];
+        self.storage
+            .read_contract(&contract_id, &mut contract_buffer)
+            .transpose()
+            .ok_or(PanicReason::ContractNotFound)?
+            .map_err(RuntimeError::Storage)?;
+
+        if contract_buffer.len() != contract_len as usize {
+            return Err(PanicReason::ContractMismatch.into())
+        }
```
netrome, line 1018: This code is repeated. How about breaking it out to a helper function?

netrome (commented):
Looks alright to me. The TODO comment needs to be resolved before I can approve. There's also some code duplication that I think we could factor out, but that's not blocking from my end.

fuel-vm/src/interpreter/blockchain.rs
acerone85: Good shout: bd00e21fb

fuel-vm/src/interpreter/blockchain.rs
acerone85: I remember discussing this with @xgreenx during one of the weekly syncs. 
I think the only optimization we can do at this stage is when handling the LDC opcode (so not in this function). In that case,  If we request the code for a contract starting at offset 0 then we can import directly into the stack. In fact, we can do this even if we request to load the contract at a greater offset than 0, but the logic becomes slightly complicated and error-prone. See https://github.com/FuelLabs/fuel-vm/pull/847/commits/ab0cc7bae7af3f97d4c8a9f2942a5834f7df8399 for how it was done.

netrome: Ah I see, I misunderstood this and thought we talked about the stack for the rust code, not within the VM :sweat_smile:.

fuel-vm/src/interpreter/blockchain.rs
netrome: Nice, thanks for updating!

netrome (approved):
Nice stuff!

netrome:
Poke me when you've fixed the clippy errors and I'll re-approve

netrome (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed 
+- [#847](https://github.com/FuelLabs/fuel-vm/pull/847): Remove `contract_read`, and change `load_contract_code`, `code_copy` and `code_root`  to explicitly load the contract code in a buffer. Also check for mismatches between contract size stored and actual size of contract in those functions.
```
MitchTurner, line 11: I'm confused by this. There is no function called `contract_read` that is removed in these changes. And you added a call to `read_contract`, which sounds like the same thing?

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -870,6 +875,28 @@ where
...
+    S: InterpreterStorage,
+{
+    let mut contract_buffer: Vec<u8> = alloc::vec![0u8; contract_len];
+    storage
+        .read_contract(&contract_id, &mut contract_buffer)
+        .transpose()
+        .ok_or(PanicReason::ContractNotFound)?
+        .map_err(RuntimeError::Storage)?;
+
+    if contract_buffer.len() != contract_len {
```
MitchTurner, line 893: If it's already allocated to to `contracct_len` size, will the `len` call ever not be `contract_len`?

MitchTurner (commented):
I'm not a VM guy, so maybe I could get you to walk me through these changes.

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -636,8 +637,12 @@ where
             self.gas_cost,
             charge_len,
         )?;
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract_bytes = contract.as_ref().as_ref();
+
+        let contract_buffer: Vec<u8> = load_contract_code_from_storage(
```
xgreenx, line 641: After reviewing this change, I see that we basically have the same code as before and allocate vectors in all cases.

So basically, this change doesn't make a lot of sense=D

The main problem is why we can't do the optimal implementation and copy directly into the memory - contract offset(and blob offset, since we have the same problem for blobs `LDC` with `mode == 1` and `BLDD` opcode).

I think we need to extend the `StorageRead` trait if we want to fix this issue in a proper way. A new method will do the same as the `read` or change the `read` method itself to something like:

```
fn read(&self, key: &Type::Key, offset: usize, buf: &mut [u8])
        -> Result<Option<usize>, Self::Error>
```

Could you change this method to work with `offset` and update all code to not allocate vector, please?

Also, it would be nice if you do the same for blobs as well.

fuel-vm/src/interpreter/blockchain.rs
acerone85: Yes, this is a leftover from refactoring, most probably this check should not be there. 
(Btw, I noticed that we have a slightly different implementation of the same function: https://github.com/FuelLabs/fuel-vm/blob/a6abe47f14a91f7498d27ebf956a882ecb52aa44/fuel-vm/src/interpreter/flow.rs#L631, so maybe we should factor out the common code and have only one function instead)

fuel-vm/src/interpreter/blockchain.rs
acerone85: Yes, the original idea was to have only one allocation. The only actual improvement in this PR as is now is fixing <MemoryStorage as StorageRead<ContractsRawCode>>::read`, which was panicking before due to slices of been different size being copied. 

Happy to extend the trait to read. I think it would be better to add a new function `read_from_offset`, and then change `read` to be a provided method with offset 0

acerone85: Sure, just to make sure: if we opt for modifying the `StorageRead::read` function, then there will be several places where the implementation needs to be updated, both in `fuel-vm` and `fuel-core`; also it is not clear to me that we can read from an arbitrary offset easily in all the implementations of the `StorageRead` trait. But I will try and see how far can I go before getting into trouble

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -994,9 +1002,8 @@ where
         self.memory.write(self.owner, dst_addr, length)?;
         self.input_contracts.check(&contract_id)?;
 
-        let contract = super::contract::contract(self.storage, &contract_id)?;
-        let contract_bytes = contract.as_ref().as_ref();
-        let contract_len = contract_bytes.len();
+        let contract_bytes = self.memory.write(self.owner, dst_addr, length)?;
+        let contract_len = contract_size(&self.storage, &contract_id)?;
         let charge_len = core::cmp::max(contract_len as u64, length);
```
acerone85, line 1007: We never read more than `length` bytes with the new version. Should `charge_len` be changed accordingly to reflect this?

acerone85:
With the new version the function `code_root` still uses the old `StorageInspect::get<RawContractCode>` function. This seems to be okay to me, as the function will return a borrowed reference to the whole contract, which will then be moved into chunks into a `MerkleTree` structure to compute the code root. In this case, using `StorageRead::read<RawContractCode>` would require allocating more space to store the contract code, which can be avoided. Do you agree with this?

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -650,15 +648,25 @@ where
...
+        if contract_offset >= contract_len.into() {
+            contract_buffer[..].fill(0);
+        } else {
+            // The conract offset is within the contract bounds, which
+            // is guaranteed to be at most `u32::MAX`. As such, it is safe
+            // to convert `contract_offset` to `usize`, which is required
+            // to invoke `self.storage.read_contract`. Furthermore,
+            // the call to `self.storage.read_contract` cannot fail with
+            // error `StorageError::`
+            #[allow(clippy::cast_possible_truncation)]
```
xgreenx, line 663: I think it is better to remove it and `contract_len.into()` above, and convert `contract_offset` into the `u32` at the begin of the function explicitly returning an error. In this case you don't need to have a comment

fuel-vm/src/interpreter/blockchain.rs
```diff
@@ -650,15 +648,25 @@ where
...
+            self.storage
+                .read_contract(&contract_id, contract_offset as usize, contract_buffer)
```
xgreenx, line 665: I think the `fuel-vm` should be responsible for extracting subsection from `contract_buffer` and passing it to the `read_contract` function.

Plus, it should be responsible for filling `contract_buffer[(length - (contract_len - contract_offset))..].fill(0);` as well.

Because it is requirement from the specification and we don't want to leak business logic to the storage.

The same is related to another place in the PR.

xgreenx (changes requested):

xgreenx (approved):
@Dentosal Could you also review this PR please?=)

Dentosal (approved):

fuel-vm/src/interpreter/blockchain.rs
xgreenx: I also prefer to keep the current charging strategy, even if we overcharge user=) The `length` in most cases should be not more than `contract_len`.

PR #848, opened 2024-10-04
LDC support in predicates

Voxelot:
Enables predicates to load immutable blobs of code from storage. This helps reduce bloating blockspace with large predicates.

Core changes:

- `$PC` can now jump outside of predicate memory range
- `ldc`, `bldd`, and `bsiz` instructions are allowed inside of predicates (but only for blobs, no contracts)
- verifying or estimating predicates now requires access to blob storage

### DIFF
--- a/fuel-asm/src/lib.rs
+++ b/fuel-asm/src/lib.rs
@@ -993,9 +993,9 @@ fn check_predicate_allowed() {
     for byte in 0..u8::MAX {
         if let Ok(repr) = Opcode::try_from(byte) {
             let should_allow = match repr {
-                BAL | BHEI | BHSH | BURN | CALL | CB | CCP | CROO | CSIZ | LDC | LOG
-                | LOGD | MINT | RETD | RVRT | SMO | SCWQ | SRW | SRWQ | SWW | SWWQ
-                | TIME | TR | TRO | ECAL | BSIZ | BLDD => false,
+                BAL | BHEI | BHSH | BURN | CALL | CB | CCP | CROO | CSIZ | LOG | LOGD
+                | MINT | RETD | RVRT | SMO | SCWQ | SRW | SRWQ | SWW | SWWQ | TIME
+                | TR | TRO | ECAL => false,
                 _ => true,
             };
             assert_eq!(should_allow, repr.is_predicate_allowed());
--- a/fuel-asm/src/lib.rs
+++ b/fuel-asm/src/lib.rs
@@ -701,7 +701,7 @@ impl Opcode {
             | K256 | S256 | NOOP | FLAG | ADDI | ANDI | DIVI | EXPI | MODI | MULI
             | MLDV | ORI | SLLI | SRLI | SUBI | XORI | JNEI | LB | LW | SB | SW
             | MCPI | MCLI | GM | MOVI | JNZI | JI | JMP | JNE | JMPF | JMPB | JNZF
-            | JNZB | JNEF | JNEB | CFEI | CFSI | CFE | CFS | GTF => true,
+            | JNZB | JNEF | JNEB | CFEI | CFSI | CFE | CFS | GTF | LDC => true,
             _ => false,
         }
     }
--- a/fuel-asm/src/lib.rs
+++ b/fuel-asm/src/lib.rs
@@ -701,7 +701,9 @@ impl Opcode {
             | K256 | S256 | NOOP | FLAG | ADDI | ANDI | DIVI | EXPI | MODI | MULI
             | MLDV | ORI | SLLI | SRLI | SUBI | XORI | JNEI | LB | LW | SB | SW
             | MCPI | MCLI | GM | MOVI | JNZI | JI | JMP | JNE | JMPF | JMPB | JNZF
-            | JNZB | JNEF | JNEB | CFEI | CFSI | CFE | CFS | GTF | LDC => true,
+            | JNZB | JNEF | JNEB | CFEI | CFSI | CFE | CFS | GTF | LDC | BSIZ | BLDD => {
+                true
+            }
             _ => false,
         }
     }
--- a/fuel-tx/src/transaction/id.rs
+++ b/fuel-tx/src/transaction/id.rs
@@ -108,6 +108,7 @@ where
                 }) if owner == &pk => Some(*witness_index as usize),
                 _ => None,
             })
+            .sorted()
             .dedup()
             .collect_vec();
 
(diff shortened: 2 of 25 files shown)


fuel-asm/src/lib.rs
```diff
@@ -701,7 +701,9 @@ impl Opcode {
             | K256 | S256 | NOOP | FLAG | ADDI | ANDI | DIVI | EXPI | MODI | MULI
             | MLDV | ORI | SLLI | SRLI | SUBI | XORI | JNEI | LB | LW | SB | SW
             | MCPI | MCLI | GM | MOVI | JNZI | JI | JMP | JNE | JMPF | JMPB | JNZF
-            | JNZB | JNEF | JNEB | CFEI | CFSI | CFE | CFS | GTF => true,
+            | JNZB | JNEF | JNEB | CFEI | CFSI | CFE | CFS | GTF | LDC | BSIZ | BLDD => {
+                true
```
acerone85, line 705: test needs updating (https://github.com/FuelLabs/fuel-vm/blob/39eb13e176c84b3e82187dbe50168587910b4b28/fuel-asm/src/lib.rs#L694)

fuel-vm/Cargo.toml
```diff
@@ -77,35 +77,35 @@ tokio-rayon = "2.1.0"
...
-    "itertools/use_std",
+  "alloc",
+  "fuel-crypto/std",
+  "fuel-types/std",
+  "fuel-asm/std",
+  "fuel-tx/std",
+  "itertools/use_std",
 ]
-alloc = ["fuel-asm/alloc", "fuel-tx/alloc", "fuel-tx/alloc"]
+alloc = ["fuel-asm/alloc", "fuel-tx/alloc", "fuel-crypto/alloc"]
```
acerone85, line 87: I don't have enough context to understand this line. Why are we switching the alloc feature?

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -257,6 +281,7 @@ where
                 let tx = kind.tx().clone();
                 let my_params = params.clone();
                 let mut memory = pool.get_new().await;
+                let my_storage = storage.clone();
```
acerone85, line 284: jut checking, in practice this is going to result in a shallow copy (e.g. the actual storage is hidden behind an Arc)?

fuel-vm/src/storage/predicate.rs
```diff
@@ -63,246 +92,281 @@ impl From<StorageUnavailable> for RuntimeError<StorageUnavailable> {
...
+    Type: Mappable,
+    D: StorageRead<BlobData>,
+{
+    type Error = PredicateStorageError;
 
     fn get(
         &self,
         _key: &Type::Key,
-    ) -> Result<Option<Cow<'_, Type::OwnedValue>>, StorageUnavailable> {
-        Err(StorageUnavailable)
```
acerone85, line 73: Is `StorageUnavailable still used`? Its definition has not been removed

fuel-vm/src/storage/predicate.rs
```diff
@@ -63,246 +92,281 @@ impl From<StorageUnavailable> for RuntimeError<StorageUnavailable> {
...
         _key: &<ContractsState as Mappable>::Key,
-    ) -> Result<Option<usize>, StorageUnavailable> {
-        Err(StorageUnavailable)
+    ) -> Result<Option<usize>, Self::Error> {
+        Err(Self::Error::UnsupportedStorageOperation)
     }
 }
 
-impl StorageRead<ContractsState> for PredicateStorage {
+impl<D> StorageRead<ContractsState> for PredicateStorage<D>
```
acerone85, line 204: question for my own understanding: in what cases can these operations be called?

acerone85 (commented):
My understanding is that after this PR the contract code will be loaded from a blob, hence we keep predicate stateless. Just to check that I am getting this correctly?

fuel-tx/src/transaction/id.rs
```diff
@@ -108,6 +108,7 @@ where
                 }) if owner == &pk => Some(*witness_index as usize),
                 _ => None,
             })
+            .sorted()
```
AurelienFT, line 111: personal question: what is the reason behind this ?

AurelienFT (approved):

fuel-vm/src/interpreter/executors/main.rs
xgreenx: Removed

fuel-vm/src/storage/predicate.rs
xgreenx: Removed

fuel-vm/src/storage/predicate.rs
xgreenx: We allow only blobs in the predicate. Any other access to the storage should return an error

fuel-tx/src/transaction/id.rs
xgreenx: Just notice going over the code that we call `dedup`, but it not we expect=) So I added `dedup` to have behavior that we want

xgreenx:
> My understanding is that after this PR the contract code will be loaded from a blob, hence we keep predicate stateless. Just to check that I am getting this correctly?

No. In this change we break the idea that predicates are stateless. This PR enables access to the blobs in the predicate context. Since blobs are monotonic it shouldn't invalidate the predicate. If blob doesn't exist, we return a panic. If blob exists, it will not change later.

xgreenx (approved):

fuel-tx/src/transaction/id.rs
rafal-ch: Yeah, `dedup()` works on "[consecutive identical elements](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.dedup)". This bit me many times in the past :)
Good catch.

fuel-vm/src/storage/predicate.rs
```diff
@@ -30,279 +34,420 @@ use fuel_types::{
...
+}
 
-impl From<StorageUnavailable> for InterpreterError<StorageUnavailable> {
-    fn from(val: StorageUnavailable) -> Self {
-        let rt: RuntimeError<StorageUnavailable> = val.into();
+/// Errors that happen when using predicate storage
+#[derive(Debug, Clone)]
+pub enum PredicateStorageError {
+    /// Storage operation is unavailable in predicate context.
+    UnsupportedStorageOperation,
```
acerone85, line 71: nit, and definitely not for this PR. But this could be also a `StorageError`.

acerone85:
I think this has been tested already extensively, it would be nice to have a test that checks that loading blobs in a predicate works as expected. That said, it would be nice to have tests for it. Not blocking.

acerone85 (approved):

Voxelot:
> I think this has been tested already extensively, it would be nice to have a test that checks that loading blobs in a predicate works as expected. That said, it would be nice to have tests for it. Not blocking.

The rust-sdk has tests with their new custom blob loader to verify this all works.

PR #871, opened 2024-11-16
Add expiration policy

AurelienFT:
https://github.com/FuelLabs/fuel-specs/issues/593

Spec PR : https://github.com/FuelLabs/fuel-specs/pull/616

## Checklist

### Before requesting review

- [ ] Someone else?


fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -123,6 +123,56 @@ fn maturity() {
     assert_eq!(ValidityError::TransactionMaturity, err);
 }
 
+#[test]
+fn expiration() {
```
MitchTurner, line 127: I know you're just basing this on the old tests, but can you add the SUT and G/W/T to this test.

This is a good example of why we should include those. I'm having a hard time understanding what the two `assert`s are for, so this should _probably_ be 2 tests as well.

fuel-tx/src/tests/valid_cases/transaction/upgrade.rs
```diff
@@ -91,6 +93,23 @@ fn maturity() {
     assert_eq!(Err(ValidityError::TransactionMaturity), result);
 }
 
+#[test]
+fn expiration() {
```
MitchTurner, line 97: I think we need a better name here.

fuel-tx/src/tests/valid_cases/transaction/upload.rs
```diff
@@ -168,6 +171,23 @@ fn maturity() {
     assert_eq!(Err(ValidityError::TransactionMaturity), result);
 }
 
+#[test]
+fn expiration() {
```
MitchTurner, line 175: And here.

fuel-vm/src/tests/validation.rs
```diff
@@ -67,6 +67,36 @@ fn transaction_can_be_executed_after_maturity() {
     assert!(result.is_ok());
 }
 
+#[test]
+fn transaction_can_be_executed_before_expiration() {
```
MitchTurner, line 71: SUT and G/W/T here too :)

fuel-tx/src/transaction.rs
```diff
@@ -832,6 +832,27 @@ pub mod field {
...
+    }
+
+    impl<T: Policies + ?Sized> Expiration for T {
+        #[inline(always)]
+        fn expiration(&self) -> BlockHeight {
+            self.policies()
+                .get(PolicyType::Expiration)
+                .map(|value| u32::try_from(value).unwrap_or(u32::MAX).into())
+                .unwrap_or_default()
+        }
```
MitchTurner, line 847: I noticed that there are a few places you added:
```rs
.with_expiration(u32::MAX.into());
```

I feel like that shouldn't be necessary. Perhaps we should something like:
```suggestion
        fn expiration(&self) -> BlockHeight {
            self.policies()
                .get(PolicyType::Expiration)
                .and_then(|value| u32::try_from(value).ok()))
                .unwrap_or(u32::MAX)
                .into()
        }
```
Or something like that? i.e. move the `unwrap_or` up one level?

MitchTurner (commented):
About time we added the expiration :) Both maturity and expiration are important for predicate writing.

fuel-tx/src/tests/valid_cases/transaction.rs
Voxelot: Yes or at least we should have some comments throughout the test explaining what is being tested and what the expected result is.

Voxelot:
We might want to beef up the tests around da_compression to cover policy compression/decompression, as right now it is only testing randomly generated txs without any policies.

fuel-tx/src/transaction.rs
AurelienFT: Yeah I agree the default behavior is better if it doesn't constraint at all. Updated and updated all tests where it was not necessary.

fuel-vm/src/tests/validation.rs
AurelienFT: Done

fuel-tx/src/tests/valid_cases/transaction/upload.rs
AurelienFT: Same as above. Done

fuel-tx/src/tests/valid_cases/transaction/upgrade.rs
AurelienFT: Same as above. Done

fuel-tx/src/tests/valid_cases/transaction.rs
AurelienFT: Yeah I just wanted to follow the previous "convention" but okey I will go for the new improved way. Done.

fuel-tx/src/transaction.rs
MitchTurner: It doesn't specify in the specs, BTW. So we should probably clarify there.

AurelienFT: Will ask to our sync

MitchTurner (approved):
LGTM!

fuel-tx/src/tests/bytes.rs
```diff
@@ -383,6 +383,7 @@ fn transaction_serde_serialization_deserialization() {
             Policies::new()
                 .with_tip(Word::MAX >> 1)
                 .with_maturity((u32::MAX >> 3).into())
+                .with_expiration((u32::MAX >> 2).into())
```
acerone85, line 386:
```suggestion
                .with_expiration((u32::MAX >> 2).into())
                .with_maturity((u32::MAX >> 3).into())
```

fuel-tx/src/transaction/policies.rs
```diff
@@ -305,6 +328,11 @@ impl Distribution<Policies> for Standard {
             policies.set(PolicyType::Maturity, Some(maturity as u64));
         }
 
+        if policies.get(PolicyType::Expiration).is_some() {
```
rafal-ch, line 331: nit: I think we can use `is_set()` instead of `get(...).is_some()`

fuel-vm/src/tests/validation.rs
```diff
@@ -67,6 +67,40 @@ fn transaction_can_be_executed_after_maturity() {
...
 }
 
+#[test]
+fn transaction__execution__works_before_expiration() {
+    let arb_max_fee = 1;
+
+    let rng = &mut StdRng::seed_from_u64(2322u64);
+
+    // Given
+    const EXPIRATION: BlockHeight = BlockHeight::new(2);
```
rafal-ch, line 77:
```suggestion
    const EXPIRATION: BlockHeight = BlockHeight::new(1);
```

to clearly show that transactions expire only when the current block height is **higher** than the specified expiry block (if the match, the transaction is still ok).

rafal-ch: For the record, I'd also add a similar test `fn transaction__execution__doesn_not_work_after_expiration()` and use, for example:
```
    const EXPIRATION: BlockHeight = BlockHeight::new(9);
    const BLOCK_HEIGHT: BlockHeight = BlockHeight::new(10);
```

xgreenx (commented):
I think we can also update the `into_ready` function to verify the expiration policy as well. Because this function exists to verify stuff that can change after verification.

AurelienFT:
@xgreenx Do we agree that if I add it to the `into_ready` function it will require everyone to update to add the `block_height` parameters which creates a big breaking change ?

fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -123,6 +123,87 @@ fn maturity() {
...
+        .add_fee_input()
+        .finalize()
+        // When
+        .check(block_height, &test_params())
+        // Then
+        .expect("Failed to validate script");
+}
+
+#[test]
+fn script__check__invalid_expiration_policy() {
```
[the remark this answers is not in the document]
AurelienFT: The goal is to check that the error variant is correct but yeah it's possible to make it less different. Updated

xgreenx:
`into_ready` only used in the `fuel-vm`, so I think we can do that=)

xgreenx (dismissed):

fuel-tx/src/tests/valid_cases/transaction/blob.rs
```diff
@@ -69,6 +70,22 @@ fn check__fails_if_maturity_not_met() {
...
+    let failing_block_height = block_height.pred().unwrap();
+    let tx = valid_blob_transaction()
+        .expiration(failing_block_height)
+        .finalize_as_transaction();
+
+    // When
+    let result = tx.check(block_height, &test_params());
+
+    // Then
+    assert_eq!(Err(ValidityError::TransactionExpiration), result);
```
MitchTurner, line 86: Why do we check the happy and sad paths for `script` and `create`, but not for `blob`, `upgrade`, and `upload`?

AurelienFT: It follows the same pattern as the others policies, the correct cas is only tested in `valid_cases.rs` file.

AurelienFT: Okeyyyy, added :)

MitchTurner (approved):

xgreenx (approved):

PR #872, opened 2024-11-18
Add owner policy

AurelienFT:
https://github.com/FuelLabs/fuel-specs/issues/617

## Checklist

### Before requesting review

- [ ] Someone else?


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - [871](https://github.com/FuelLabs/fuel-vm/pull/871): Add `expiration` policy that prevent a transaction to be inserted after a given block height.
+- [872](https://github.com/FuelLabs/fuel-vm/pull/872): Add `owner` policy to specify a primary owner of the transaction. We use an index of an input to identify the owner.
```
acerone85, line 12:
```suggestion
- [872](https://github.com/FuelLabs/fuel-vm/pull/872): Add `owner` policy to specify a primary owner of the transaction. We use the index of an input to identify the owner.
```
IIUC?

fuel-asm/src/args.rs
```diff
@@ -248,6 +248,9 @@ crate::enum_try_from! {
 
         /// Set `$rA` to `tx.policies[count_ones(0b11111 & tx.policyTypes) - 1].expiration`
         PolicyExpiration = 0x505,
+
+        /// Set `$rA` to `tx.policies[count_ones(0b111111 & tx.policyTypes) - 1].owner`
```
acerone85, line 252: this applies also to comments for other variants, but I could not find `tx.policyTypes` anywhere, so not sure how I should interpret `0b111111 & tx.policyTypes`

acerone85: found it, NVM :)

fuel-tx/src/tests/valid_cases/transaction.rs
```diff
@@ -395,6 +395,111 @@ fn create__check__no_max_fee_fails() {
...
+    let err = TransactionBuilder::script(generate_bytes(rng), generate_bytes(rng))
+        .add_fee_input()
+        .owner(1)
+        .finalize()
+        .check(block_height, &test_params())
+        .expect_err("Expected erroneous transaction");
+
+    // Then
+    assert_eq!(ValidityError::TransactionOwnerIndexDoesntExist, err);
+}
```
acerone85, line 447: following on this test, I suggest to add another one: 

```rust
fn script__check__set_owner_bad_input_type() {
    // Given
    let rng = &mut StdRng::seed_from_u64(8586);
    let block_height = 1000.into();

    // When
    let err = TransactionBuilder::script(generate_bytes(rng), generate_bytes(rng))
        .add_fee_input()
        .add_contract_input()
        .owner(1)
        .finalize()
        .check(block_height, &test_params())
        .expect_err("Expected erroneous transaction");

    // Then
    assert_eq!(ValidityError::TransactionOwnerInputHasNoOwner, err);
}
```

It requires you to add a method `TransactionBuilder::add_contract_input()`, but this can be easily done: 

```rust
    pub fn add_contract_input(&mut self) -> &mut Self {
        let contract_input = Input::Contract(Contract::default());
        self.tx.add_input(contract_input);
        self
    }
```

(Btw, if you want to be pedantic you could also test with other inputs, and rewrite `add_contract_input` to take a `SeedableRng` in input).

fuel-vm/src/tests/metadata.rs
```diff
@@ -622,6 +624,12 @@ fn get_transaction_fields() {
         op::eq(0x10, 0x10, 0x11),
         op::and(0x20, 0x20, 0x10),
 
+        op::movi(0x19, 0x00),
+        op::movi(0x11, owner_idx as Immediate18),
```
acerone85, line 628: question for me to understand: why `Immediate18`? By looking at the specs and the [implementation](https://github.com/FuelLabs/fuel-vm/blob/9079849648355542a9a3adb80ed5d62a3abe283b/fuel-asm/src/op.rs#L38) of `GTF`, I would expect `Immediate12` here?

fuel-vm/src/tests/metadata.rs
```diff
@@ -622,6 +624,12 @@ fn get_transaction_fields() {
...
+        op::gtf_args(0x10, 0x19, GTFArgs::PolicyOwner),
```
acerone85, line 629: Another question for me to understand here. 
I guess the `PolicyOwner` field is taken from the transaction included in the frame pointer. What is the value of the register `0x19` here used for? From the specs i can infer that it is related to the part of the field that we want to read, but it is not clear if this is the start or end offset or something else. 

Anyway, not related to this PR (also probably this value is ignored since I'd expect the owner value to be fixed size).

acerone85:
Another question for me to understand: 
What is the motivation behind this change? I.e. why is it important for transactions to have an explicit user, and how it is going to be used by applications?

fuel-vm/src/tests/metadata.rs
AurelienFT: I don't know I just followed the existing pattern. Maybe @Dentosal knows

fuel-vm/src/tests/metadata.rs
AurelienFT: Same as above I jsut copy pasted the existing. cc @Dentosal

fuel-tx/src/tests/valid_cases/transaction.rs
AurelienFT: I added your test and reuse some code already existing to create the contract because adding this in the public interface used by all Rust project inside a PR that has no meanign with that, I'm not confident.

fuel-tx/src/transaction/validity.rs
```diff
@@ -377,6 +378,18 @@ where
...
+        let owner = usize::try_from(owner)
+            .map_err(|_| ValidityError::TransactionOwnerIndexDoesntExist)?;
+        if owner >= tx.inputs().len() {
+            Err(ValidityError::TransactionOwnerIndexDoesntExist)?
+        }
+        // SAFETY: `owner` is guaranteed to be a valid index because it was checked above.
+        if !&tx.inputs()[owner].input_owner().is_some() {
+            Err(ValidityError::TransactionOwnerInputHasNoOwner)?
+        }
+    }
```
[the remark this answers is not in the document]
AurelienFT: Done

AurelienFT:
> Another question for me to understand: What is the motivation behind this change? I.e. why is it important for transactions to have an explicit user, and how it is going to be used by applications?

The motivation is explained here : https://github.com/FuelLabs/fuel-specs/issues/617 but in other word the smart contract sometimes wants to be able to know who is the caller (partially because dev are use to EVM `msg.sender` and also to have a common standard of who is the owner is needed by contract). This will allow contracts to have the language when speaking about owners instead of the current anarchy.

acerone85 (approved):

fuel-tx/src/transaction/validity/error.rs
```diff
@@ -133,6 +133,8 @@ pub enum ValidityError {
     TransactionInputsMax,
     TransactionOutputsMax,
     TransactionWitnessesMax,
+    TransactionOwnerIndexDoesntExist,
```
rafal-ch, line 136: a nit:
```suggestion
    TransactionOwnerIndexOutOfBounds
```

fuel-tx/src/transaction/validity.rs
```diff
@@ -321,6 +322,24 @@ where
...
+    if let Some(owner) = tx.owner() {
+        let owner = usize::try_from(owner)
+            .map_err(|_| ValidityError::TransactionOwnerIndexDoesntExist)?;
+        if owner >= tx.inputs().len() {
+            Err(ValidityError::TransactionOwnerIndexDoesntExist)?
+        }
+        // SAFETY: `owner` is guaranteed to be a valid index because it was checked above.
+        if !&tx.inputs()[owner].input_owner().is_some() {
+            Err(ValidityError::TransactionOwnerInputHasNoOwner)?
+        }
```
rafal-ch, line 338: can we avoid manual safety check with something like:
```
        if tx
            .inputs()
            .get(owner)
            .and_then(|input| input.input_owner())
            .is_none()
        {
            return Err(ValidityError::TransactionOwnerInputHasNoOwner);
        }
```
?

fuel-tx/src/transaction/validity.rs
```diff
@@ -321,6 +322,24 @@ where
     Ok(())
 }
 
+pub(crate) fn check_owner<T>(tx: &T) -> Result<(), ValidityError>
```
rafal-ch, line 325: Do you think adding some UTs to this function makes sense?

AurelienFT: I don't really think so it's well tested with the other tests and it's not a practice we currently have for validation rules. I don't find this one more sensible or difficult to understand.

fuel-tx/src/transaction/validity.rs
AurelienFT: Ok I will change that. (not the whole code to mentionned because otherwise we will loose the two different errors)

fuel-tx/src/transaction/validity/error.rs
AurelienFT: I'm ok with that. Changed.

xgreenx:
I marked this PR as a Draft for now until we resolved https://forum.fuel.network/t/proposal-for-note-inputs/7094 proposal=)

MitchTurner:
I'm going to close for now since it looks like we have other routes we might take.

PR #874, opened 2024-11-28
cleanup for add zk opcodes PR

Dentosal:
* Use `MemoryOverflow` instead of `ArithmeticOverflow` for memory ptr overflows
* Use `From<PanicReason>` impl of `PanicOrBug` to simplify code
* Simplify some memory access patterns


AurelienFT:
I will check to fix it.

AurelienFT:
The coordinates reading was inverted for a point.

PR #881, opened 2024-12-09
Execution trace recording

### LINKED ISSUE #867: Execution traces
Instead of running the interactive debugger, store full execution trace. This means storing state delta for each instruction as well.

Dentosal:
Closes #867. Core PR https://github.com/FuelLabs/fuel-core/pull/2491.

Introduces a mechanism for recording an *execution trace*, which can be used to inspect VM states after execution without having to use a debugger. Most importantly it allows reading registers and memory at certain points of the exection, which is quite useful for tooling like indexers.

## Checklist
- [ ] Breaking changes recorded in changelog
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (no spec changes)

### Before requesting review

- [ ] Indexer team [in this Slack thread](https://fuellabs.slack.com/archives/C01L1HUJSL8/p1733244266212099)


fuel-vm/src/interpreter/executors/instruction.rs
```diff
@@ -80,8 +80,15 @@ where
             }
         }
 
-        self.instruction_inner(raw.into())
-            .map_err(|e| InterpreterError::from_runtime(e, raw.into()))
+        let result = self
```
xgreenx, line 83: I think it will be more beneficial if, on the creation of the interpreter, we will pass some object that implements traits like `Tracer` or something.

```rust
trait Tracer<VM> {
    fn trace_instruction_before_execution(
        &mut self,
        instruction: &Intstuction,
        vm: &mut VM,
    );

    fn trace_instruction_after_execution(
        &mut self,
        instruction: &Intstuction,
        vm: &mut VM,
    );
}
```

In this case, we will have more flexibility in the implementation and can define logic to get traces on the `fuel-core` side. In the case of tooling, they can maybe create a local debugger. In the case of indexers, they can index something inside of these functions. 

Of course, this solution means that we need to have getters for all internal fields, including `MemoryInstance`.

fuel-vm/src/interpreter/memory.rs
```diff
@@ -403,6 +418,37 @@ impl MemoryInstance {
...
     }
 
+    /// Diff of from `self` to `new` memory state.
+    /// Panics if new instance is not possible to reach from the current one,
+    /// for instance if it has smaller stack or heap allocation.
+    pub fn diff_patches(&self, new: &MemoryInstance) -> Vec<MemorySliceChange> {
+        let mut changes = Vec::new();
+        let mut current_change: Option<MemorySliceChange> = None;
+
+        for i in 0..MEM_SIZE {
```
xgreenx, line 428: Ohh, it will be very expensive to do 10^7 iterations per each instruction=D It would be nice to have something more performant=)

If we go for now with `Tracer` approach, we can decide later in the `fuel-core` how do we want to do that in an optimal way

xgreenx (commented):
I think most of the code from this PR can live in the `fuel-core` and there we can gather tracer via the `Tracer` trait

fuel-vm/src/interpreter/memory.rs
Dentosal: I'm very aware of this. I have lots of ideas on how to optimize this, and I'm going to write some benchmarks to see what approach is the right one.

fuel-vm/src/interpreter/executors/instruction.rs
Dentosal: I think nobody is actually going to use this, and it's an unnecessary complicaton. But it would be nice to be wrong about this, and it's not that much extra complication.

Fortunately we already support almost all fields this needs, as ECAL support is also done like this.

Dentosal:
Closing in favor of storage access recording and debugger -based approach, that doesn't require VM changes.

Issue #812, opened 2024-09-05
GTF support for new tx types


Dentosal:
From Slack:

> I don't see a GTF opcode defined to get the witnessesCount, witnesses, outputs, outputsCount, inputs, and inputsCount for TransactionUpgrade, TransactionUpload, and TransactionBlob in the [specs](https://docs.fuel.network/docs/nightly/specs/fuel-vm/instruction-set/#gtf-get-transaction-fields) as well as transaction specific types such as subsectionIndex, subsectionsNumber, root, witnessIndex, proofSetCount, and proofSet in TransactionUpload.
>
> This is blocking https://github.com/FuelLabs/sway/pull/6500

xgreenx:
Also, it seems that we have a different getter for each type of transaction. I think we just need to have one getter for all types. And if the transaction doesn't support it, return an error like in the case of Mint transaction.

<img width="501" alt="image" src="https://github.com/user-attachments/assets/72e3f151-b7b1-4b8e-acdd-133d0f4cdc99">

<img width="482" alt="image" src="https://github.com/user-attachments/assets/31439733-636f-4351-ab14-e0a2d6f7cace">

PR #882, opened 2024-12-10
Add new tx specific GTFArgs and make new generic GTFArgs to replace duplication

AurelienFT:
https://github.com/FuelLabs/fuel-vm/issues/812

Add new transactions specific `GTFArgs`, creates new `GTFArgs` to deprecate duplicated one between Script and Create and make them generic for all transactions implementing `ExecutableTransaction`.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/fuel-asm/src/args.rs
+++ b/fuel-asm/src/args.rs
@@ -1,3 +1,5 @@
+#![allow(deprecated)]
+
 pub mod wideint;
 
 /// 12-bits immediate value type
@@ -64,12 +66,15 @@ crate::enum_try_from! {
         ScriptDataLength = 0x004,
 
         /// Set `$rA` to `tx.inputsCount`
+        #[deprecated(since = "0.60.0", note = "Use the generic `TxInputsCount` instead")]
         ScriptInputsCount = 0x005,
 
         /// Set `$rA` to `tx.outputsCount`
+        #[deprecated(since = "0.60.0", note = "Use the generic `TxOutputsCount` instead")]
         ScriptOutputsCount = 0x006,
 
         /// Set `$rA` to `tx.witnessesCount`
+        #[deprecated(since = "0.60.0", note = "Use the generic `TxWitnessesCount` instead")]
         ScriptWitnessesCount = 0x007,
 
         /// Set `$rA` to `Memory address of tx.script`
@@ -79,12 +84,15 @@ crate::enum_try_from! {
         ScriptData = 0x00A,
 
         /// Set `$rA` to `Memory address of tx.inputs[$rB]`
+        #[deprecated(since = "0.60.0", note = "Use the generic `TxInputAtIndex` instead")]
         ScriptInputAtIndex = 0x00B,
 
         /// Set `$rA` to `Memory address of t.outputs[$rB]`
+        #[deprecated(since = "0.60.0", note = "Use the generic `TxOutputAtIndex` instead")]
         ScriptOutputAtIndex = 0x00C,
 
         /// Set `$rA` to `Memory address of tx.witnesses[$rB]`
+        #[deprecated(since = "0.60.0", note = "Use the generic `TxWitnessAtIndex` instead")]
         ScriptWitnessAtIndex = 0x00D,
 
         /// Set `$rA` to size of the transaction in memory, in bytes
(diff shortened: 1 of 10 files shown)


fuel-asm/src/args.rs
```diff
@@ -248,6 +262,51 @@ crate::enum_try_from! {
...
+        UploadProofSetCount = 0x604,
+
+        /// Set `$rA` to `Memory address of tx.proofSet[$rB]`
+        UploadProofSetAtIndex = 0x605,
+
+        /// Set `$rA` to `Memory address of tx.id`
+        BlobId = 0x700,
+
+        /// Set `$rA` to `tx.witnessIndex`
+        BlobWitnessIndex = 0x701,
```
xgreenx, line 288: I think we need better comments here=)

fuel-vm/src/tests/metadata.rs
```diff
@@ -551,19 +573,6 @@ fn get_transaction_fields() {
...
-    // TODO GTFArgs::CreateBytecodeWitnessIndex
-    // TODO GTFArgs::CreateStorageSlotsCount
-    // TODO GTFArgs::CreateSalt
-    // TODO GTFArgs::CreateStorageSlotAtIndex
-    // TODO GTFArgs::OutputContractCreatedContractId
-    // TODO GTFArgs::OutputContractCreatedStateRoot
-
-    // blocked by https://github.com/FuelLabs/fuel-vm/issues/59
-    // TODO GTFArgs::InputCoinTxPointer
-    // TODO GTFArgs::InputContractTxPointer
```
xgreenx, line 565: Why do we remove these todos?=D

fuel-tx/src/transaction/repr.rs
```diff
@@ -33,3 +38,19 @@ impl From<&Transaction> for TransactionRepr {
         }
     }
 }
+
+impl TryFrom<Word> for TransactionRepr {
```
xgreenx, line 42: I don't see where do we use it. Could you point please?=)

fuel-vm/src/tests/metadata.rs
AurelienFT: Because I added all the tests to tests them all in the PR héhé

fuel-tx/src/transaction/repr.rs
AurelienFT: It's used here : https://github.com/FuelLabs/fuel-vm/blob/4618cf67567d312a74d17328f72940efaf217e18/fuel-vm/src/interpreter/metadata.rs#L648
Which is a internal type that allow more clean matching for tx specific GTF here : https://github.com/FuelLabs/fuel-vm/blob/4618cf67567d312a74d17328f72940efaf217e18/fuel-vm/src/interpreter/metadata.rs#L540

fuel-asm/src/args.rs
AurelienFT: I don't really see what information should be explained here. I tried to follow the current information it's not a memory address or anything. Maybe adding that is blob specific ? But the name seems straight forward. Maybe explaning what is the witness index but in this case we should explain what is each field fetch by the other GTF

fuel-tx/src/transaction/repr.rs
xgreenx: Okay, I see. I think we don't need to put unreachable use cases(`InvalidTransactionType`) into a panic reason. I think we should use `Bug` in the case if `Mint` transaction presents in VM, but it is not possible right now, because it doesn't implement `ExecutableTransaction` trait .

fuel-vm/src/tests/metadata.rs
xgreenx: I don't see where we test `CreateBytecodeLength`, `CreateBytecodeWitnessIndex`, `CreateSalt`, `CreateStorageSlotAtIndex`, `OutputContractCreatedContractId`, `OutputContractCreatedStateRoot`, `InputCoinTxPointer`, `InputContractTxPointer` GTF opcodes. Could you point to corresponding tests please?

AurelienFT: It starts here : https://github.com/FuelLabs/fuel-vm/blob/067d836a337472be2d8b9b8daba9d99c9aa8e784/fuel-vm/src/tests/metadata.rs#L1025 and there is a test for each gtf.

Side node : InputContractTxPointer and InputCoinTxPointer do not exist anymore.

fuel-tx/src/transaction/repr.rs
AurelienFT: I'm not sure I understand, I do nothing for now and open an issue ?

fuel-vm/src/tests/metadata.rs
xgreenx: My bad, search in browser was not working=( Thank you!

fuel-tx/src/transaction/repr.rs
xgreenx: Removed in https://github.com/FuelLabs/fuel-vm/pull/882/commits/a01e2a9b4d008065cf2010e2b294414cb83ac109

AurelienFT: Okey ty

rymnc (approved):

PR #889, opened 2025-01-15
bugfix: Produce correct receipts with debugger breakpoints enabled

Dentosal:
While working on execution traces, I noticed that the receipts are produced in incorrect oder when the debugger is enabled. The PR moves the end-of-execution receipt generation to `run_program`, so it's also done when called using `Interpreter::resume`.

This is technically a breaking change, but it's a bugfix so it doesn't necessarily warrant a breakng-version bump.

## Checklist

### Before requesting review
- [x] I have created follow-up issues caused by this PR and linked them here (none!)

### DIFF
--- a/fuel-vm/src/interpreter/executors/main.rs
+++ b/fuel-vm/src/interpreter/executors/main.rs
@@ -944,82 +944,9 @@ where
             )?;
             ProgramState::Return(1)
         } else {
-            let gas_limit;
-            let is_empty_script;
-            if let Some(script) = self.transaction().as_script() {
-                gas_limit = *script.script_gas_limit();
-                is_empty_script = script.script().is_empty();
-            } else {
-                unreachable!(
-                    "Only `Script` transactions can be executed inside of the VM"
-                )
-            }
-
-            // TODO set tree balance
-
             // `Interpreter` supports only `Create` and `Script` transactions. It is not
             // `Create` -> it is `Script`.
-            let program = if !is_empty_script {
-                self.run_program()
-            } else {
-                // Return `1` as successful execution.
-                let return_val = 1;
-                self.ret(return_val)?;
-                Ok(ProgramState::Return(return_val))
-            };
-
-            let gas_used = gas_limit
-                .checked_sub(self.remaining_gas())
-                .ok_or_else(|| Bug::new(BugVariant::GlobalGasUnderflow))?;
-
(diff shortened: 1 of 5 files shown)


fuel-vm/src/tests/crypto.rs
Dentosal, line 1: These changes are unrelated, but the CI wont pass without them. The import would have to be moved behind a feature, but I found it cleaner to just inline them.

fuel-vm/src/tests/debugger.rs
```diff
@@ -0,0 +1,72 @@
...
+        IntoChecked,
+    },
+    state::ProgramState,
+};
+
+#[test]
+fn receipts_are_produced_correctly_with_stepping() {
+    let script = vec![
+        op::movi(0x20, 1234),
+        op::log(0x20, RegId::ZERO, RegId::ZERO, RegId::ZERO),
```
AurelienFT, line 30: Can we add more than one log to be sure it's ordered even more than just : "all before the return one"

AurelienFT (approved):

fuel-vm/src/tests/debugger.rs
Dentosal: This test case is enought to demonstrate the bug. The issue was that the script result receipt would be created right after the first call to `transact` returns, i.e. before the log.

fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -1029,29 +956,94 @@ where
...
+        let (result, state) = if script.script().is_empty() {
+            // Empty script is special-cased to simply return `1` as successful execution.
+            let return_val = 1;
+            self.ret(return_val)?;
+            (
+                ScriptExecutionResult::Success,
+                ProgramState::Return(return_val),
+            )
+        } else {
+            // TODO set tree balance
```
netrome, line 973: I know this is not introduced in this PR so it's not blocking, but do we have an issue for this?

netrome (approved):

xgreenx (approved):

fuel-vm/src/interpreter/executors/main.rs
Dentosal: Nothing that I know of. I'm not even sure what it's actually about.

PR #897, opened 2025-01-20
Return `TxId` when error in checking of transactions

AurelienFT:
When we try to turn a `Tx` in a `Checked<Tx>`, errors can happens, but we always start by computing the `TxId` and this can never fails. If something else, fails afterwards it's super useful to have the `TxId` to link the error with the transaction associated.

Currently, in the existing code calling `into_checked()`, if we have an error we are computing the `TxId` ourself again to link it to the `CheckError` which creates a double computation.

## Breaking change

The `TxId` is now returned with the error.

Before : 
```rust
fn into_checked(
        self,
        block_height: BlockHeight,
        consensus_params: &ConsensusParameters,
```

Now :
```rust
fn into_checked(
        self,
        block_height: BlockHeight,
        consensus_params: &ConsensusParameters,
    ) -> Result<Checked<Self>, (TxId, CheckError)>
```

## Breaking change notes

I didn't placed this behind a feature flag because I think it's always valuable, however if we want to avoid breaking for now and have this information under a special feature I'm ok with this. Inputs from reviewers would be super valuable on that.

## Checklist

### Before requesting review

### DIFF
--- a/fuel-tx/src/transaction/metadata.rs
+++ b/fuel-tx/src/transaction/metadata.rs
@@ -11,6 +11,8 @@ use crate::{
     ValidityError,
 };
 
+use super::TxId;
+
 /// Entity support metadata computation to cache results.
 pub trait Cacheable {
     /// The cache is already computed.
@@ -19,7 +21,7 @@ pub trait Cacheable {
     fn is_computed(&self) -> bool;
 
     /// Computes the cache for the entity.
-    fn precompute(&mut self, chain_id: &ChainId) -> Result<(), ValidityError>;
+    fn precompute(&mut self, chain_id: &ChainId) -> Result<(), (TxId, ValidityError)>;
 }
 
 impl Cacheable for super::Transaction {
@@ -34,7 +36,7 @@ impl Cacheable for super::Transaction {
         }
     }
 
-    fn precompute(&mut self, chain_id: &ChainId) -> Result<(), ValidityError> {
+    fn precompute(&mut self, chain_id: &ChainId) -> Result<(), (TxId, ValidityError)> {
         match self {
             Self::Script(tx) => tx.precompute(chain_id),
             Self::Create(tx) => tx.precompute(chain_id),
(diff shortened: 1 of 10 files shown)


xgreenx (commented):
It would be nice to make this change non breaking. I tried to play with generics, but it still produces breaking code in some cases: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5f70213c55d0cf6e17abf2cd3e5f17fe


Alternative solution will be to add one more function like `into_checked_basic_with_metadata` and this one will return metadata in the case of error.

AurelienFT:
I improved your playground code to be more close to reality and I have something that is breaking only because you need to specify the error type. I still consider this breaking but I would like to have your thoughts. Link : https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d256580f356615cd7ff3979bcd2e3706

xgreenx:
Yeah, it is why I'm saying it is not good solution because it is still breaking=D So because of that I'm more in favor of the `into_checked_basic_with_metadata`

AurelienFT:
> Yeah, it is why I'm saying it is not good solution because it is still breaking=D So because of that I'm more in favor of the `into_checked_basic_with_metadata`

Okey will do that but it will be `into_checked_basic_with_id` because it's only the id that we have the metadata creation is finished after some error could have been triggered

xgreenx (commented):
I just got an idea of how we can do it better. Instead of adding new methods, we can add a new type `Immutable<Tx>`, where `Immutable` only calculates the metadata but doesn't have any validation logic. Any `Immutable` can be converted into checked without calculation of the metadata.

In this case, you can have an id before the validation, and for actual validation, you can leave for STF. In this case, splitting also will be faster

AurelienFT:
> I just got an idea of how we can do it better. Instead of adding new methods, we can add a new type `Immutable<Tx>`, where `Immutable` only calculates the metadata but doesn't have any validation logic. Any `Immutable` can be converted into checked without calculation of the metadata.
> 
> In this case, you can have an id before the validation, and for actual validation, you can leave for STF. In this case, splitting also will be faster

Okey but the problem is that creating metadata can throw error too so the signature to create a new `Immutable` would be something like `from_tx(tx: Tx) -> Result<Self, (TxId, ValidityError)`.

I understand why it's faster. So will probably do.

AurelienFT:
I tried to implement something but in any case you will need to compute the `CheckedMetadata` and to pass to a check<Tx> you need to have a Tx only not a Immutable<Tx>.

I tried a PR : https://github.com/FuelLabs/fuel-vm/pull/906 but it's not compiling because of the Checked that wants a `Immutable<Tx>`. I think it's possible to do something but it will require a lot more boilerplate. And I want that we first discuss about the real motivations and is it really useful (cf previous message)

Voxelot:
Do we still need this PR? 

cc @AurelienFT

PR #898, opened 2025-01-20
Better usage of STD features in our dependencies

AurelienFT:
We spotted that there is some dependencies where we didn't used their STD features in our STD contexts. Leveraging these features allow us to increase our performances in STD contexts.

## Checklist

### Before requesting review

### DIFF
--- a/fuel-crypto/src/secp256/backend/k1/secp256k1.rs
+++ b/fuel-crypto/src/secp256/backend/k1/secp256k1.rs
@@ -28,8 +28,9 @@ use rand::{
     RngCore,
 };
 
+static CONTEXT: OnceLock<Secp256k1<secp256k1::All>> = OnceLock::new();
+
 fn get_context() -> &'static Secp256k1<secp256k1::All> {
-    static CONTEXT: OnceLock<Secp256k1<secp256k1::All>> = OnceLock::new();
     CONTEXT.get_or_init(Secp256k1::new)
 }
 
--- a/fuel-crypto/src/secp256/backend/k1/secp256k1.rs
+++ b/fuel-crypto/src/secp256/backend/k1/secp256k1.rs
@@ -28,9 +28,8 @@ use rand::{
     RngCore,
 };
 
-static CONTEXT: OnceLock<Secp256k1<secp256k1::All>> = OnceLock::new();
-
 fn get_context() -> &'static Secp256k1<secp256k1::All> {
+    static CONTEXT: OnceLock<Secp256k1<secp256k1::All>> = OnceLock::new();
     CONTEXT.get_or_init(Secp256k1::new)
 }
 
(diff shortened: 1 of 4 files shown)


fuel-crypto/src/secp256/backend/k1/secp256k1.rs
```diff
@@ -27,8 +28,9 @@ use rand::{
     RngCore,
 };
 
-lazy_static::lazy_static! {
-    static ref CONTEXT: Secp256k1<secp256k1::All> = Secp256k1::new();
+fn get_context() -> &'static Secp256k1<secp256k1::All> {
+    static CONTEXT: OnceLock<Secp256k1<secp256k1::All>> = OnceLock::new();
```
rymnc, line 32: hmm, shouldn't this be a global instead?

AurelienFT: You can't create a global `Secp256k1<secp256k1::All>` because it has no `const` constructructors

rymnc: using a OnceCell, I believe you can

AurelienFT: Yes but each time we will have to call `get_or_init` on every invocations. Is there any performance improvement to use a global otherwise, I find it more readable to be able to call directly `get_context()` everywhere instead of `CONTEXT.get_or_init(Secp256k1::new())`

rymnc: I'm just saying this is how it should be ->
```rs
static CONTEXT: OnceLock<Secp256k1<secp256k1::All>> = OnceLock::new();

fn get_context() -> &'static Secp256k1<secp256k1::All>> {
    CONTEXT.get_or_init(Secp256k1::new)
}

```

otherwise you're creating the OnceLock every time `get_context` is called, no?

AurelienFT: Me -> 🤡  :) Changed. Ty for patience xd

xgreenx: Any static is global and initialized once. I put inside of the function to make it accessible only from the context of this function=)


https://users.rust-lang.org/t/statics-inside-functions-are-fun/103041

AurelienFT: Yes. 🤡

xgreenx (approved):

fuel-crypto/src/secp256/backend/k1/secp256k1.rs
rymnc: that's neat lol didn't know about it

rymnc (approved):

PR #899, opened 2025-01-20
Add a feature that propose `u32` `tx_pointer` in `fuel-tx` and `fuel-vm`

AurelienFT:
Having `u32` `tx_pointer` is mendatory to make our system scales and support more than `u16::MAX` tx in a block.

However, changing this in the current network is a hard breaking change and is not needed for now. That's why we decided to add this change to the VM behind a feature flag if we want to use in other system or be able to enable it in our principal one too.

## Checklist

### Before requesting review


rymnc (dismissed):
LGTM, but in the `malleable_fields_do_not_affect_*` tests if we used u16::MAX and u32::MAX to indicate change in size vs adding some zeroes OR using rng to generate those values, would be nice, but is out of scope for this PR :)

xgreenx:
It is a big breaking change for the type system. We need to have proper testing and double check that serialization and deserialization will be not affected for mainnet/testnet/devnet blocks. It would be nice try to use extreme values and see when postcard serialization/deserialiation will be different(like from 1 to u16::MAX). 

Also maybe Rollup team uses bitcode, if yes, then it also break compressed blocks for them

AurelienFT:
@xgreenx I didn't want to activate this feature on Ignition is it something we want to do in the near future ? For xxx, it will be a new network and will be in the specs.

xgreenx:
We can activate it on Ignition, if tests show that old database types are compatible with it.

AurelienFT:
I have added a test to showcase that we have a difference on the postcard serialization format (it's expected because postcard use varint encoding). I tried our `.to_bytes()` which doesn't break because it's not using varint and was padded to a 8 bytes value so `u32` fits.

I don't really understand how we could activate this on Ignition but if we want to make more tests to activate this for Ignition they should be done on the `fuel-core` repository

fuel-tx/src/tx_pointer.rs
```diff
@@ -70,12 +83,28 @@ impl fmt::Display for TxPointer {
...
         write!(f, "{:08x}{:04x}", self.block_height, self.tx_index)
     }
 }
 
+#[cfg(feature = "u32-tx-pointer")]
+impl fmt::UpperHex for TxPointer {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{:08X}{:08X}", self.block_height, self.tx_index)
+    }
+}
```
xgreenx, line 105: It would be nice if we can make it to use old representation in the case of small number and big one in the case of big numbers=D In this case it will be backward compatible.

xgreenx (commented):
Along with this change would be nice to introduce "block_max_transactions" parameter to `ConsensusParameters`. It can be a separate change, we just need these two things released together.

fuel-tx/src/tx_pointer.rs
AurelienFT: Done. I also updated `from_str` to read old version then.

rymnc (approved):
lgtm, would be nice for some fuzzing/proptest for this new change

xgreenx:
> Because we use postcard in P2P and Database, the variant encoding of integers should help us to deserialize old types of transactions, because we know that tx pointer is not more than u16. We just need to have a test that verifies that(that we can decode the old transaction type into a new one with any old tx pointer).

It should be a compatibility test where we use the previous release of the `fuel-tx` to serialize the transaction and the latest version to deserialize it. You can set `TxPointer` in the loop form `0..u16::MAX` to see how it affects serialization and deserialization. To compare old Tx and new Tx, you can compare them as strings, or JSONs

AurelienFT:
Thanks for the detailed explanation @xgreenx. Done :)

version-compatibility/src/fuel_tx.rs
```diff
@@ -25,3 +25,32 @@ fn release_0_58_2_can_deserialize_latest() {
     // Then
     let _ = latest_tx.expect("Deserialization failed");
 }
+
+#[test]
+fn latest_can_deserialize_previous_tx_pointer() {
+    for idx in 0..u16::MAX {
```
xgreenx, line 31:
```suggestion
    for idx in 0..=u16::MAX {
```

Would it work if we used `u16::MAX`?

xgreenx (dismissed):

version-compatibility/src/fuel_tx.rs
```diff
@@ -26,6 +26,35 @@ fn release_0_58_2_can_deserialize_latest() {
     let _ = latest_tx.expect("Deserialization failed");
 }
 
+#[test]
+fn latest_can_deserialize_previous_tx_pointer() {
+    for idx in 0..=u16::MAX {
+        // Given
+        let tx_pointer = fuel_tx_0_59_1::TxPointer::new(1u32.into(), idx);
+        let expected = latest_fuel_tx::TxPointer::new(1u32.into(), idx.into());
```
xgreenx, line 34: May I ask you to do the same for script transaction please?=D Or `Mint` transaction for simplicity

AurelienFT: I tried it requires a lot of machinery to specify a `TxPointer` inside the transaction. At the top of the file there is the test `latest_can_deserialize_0_58_2` that do this with a test tx. `TxPointer` is set to 0 but if the deserialization wasn't working then it would flag as a error for the number of bytes. Why do you think a test for this is still valuable ?

xgreenx: Because based on the value of `idx` the result serialization can be different. It why test with `0` tx pointer will not catch the problem

AurelienFT: Found a way to do it without too much machinery. I hope it's fine for you :)

xgreenx (approved):

xgreenx (approved):

Dentosal (approved):
The code LGTM, but we should probably have this feature toggle properly documented somewhere. Also the spec should probably reflect this option somehow? I'm fine with leaving those as follow-ups, though.

PR #900, opened 2025-01-20
Compare nonce instead of message id of inputs to catch duplicates

AurelienFT:
Comparing the `nonce` of two messages inputs if enough to decide if they are duplicated or not. Previously we were computing the `message_id` now we avoid this computation and only compare the `nonce`.

It requires us to change the error variant to include `nonce` instead of `message_id`. Because the type of the content has change it's already breaking and so we can change the name also to match this type.

#### Before 
```rust
enum ValidityError {
  DuplicateMessageInputId {
        message_id: MessageId,
  },
  ...
}
```

#### After
```rust
enum ValidityError {
  DuplicateInputNonce {
        nonce: Nonce,
  },
  ...
}
```

## Checklist

### Before requesting review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Breaking
+- [900](https://github.com/FuelLabs/fuel-vm/pull/900): Change the error variant `DuplicateMessageInputId` to `DuplicateInputNonce` which now contains a nonce instead of `MessageId` for performance improvements.
+
 ### Fixed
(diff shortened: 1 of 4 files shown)


xgreenx (dismissed):

rafal-ch (approved):
LGTM :+1:

rafal-ch:
Out of curiosity, do you know what's the purpose of `MessageId`?

AurelienFT:
> Out of curiosity, do you know what's the purpose of `MessageId`?

I think it's use to distinguish each message and have a unique id for all of them across the network such as txid or coin id. Detailed spec if needed : https://github.com/FuelLabs/fuel-specs/blob/cca2e8cd2e475263c909cb902620b5ff143713ec/src/identifiers/utxo-id.md#message-id

xgreenx (approved):

PR #901, opened 2025-01-21
impl `Display` for `Receipt`

rafal-ch:
This PR implements `Display` for `Receipt` for easier logging.

Once this is merged and released, there will be a follow-up PR to the `fuel-core` to get rid of [this](https://github.com/FuelLabs/fuel-core/blob/1a294e2892b937341f022f25738dd7bc47604d66/crates/fuel-core/src/graphql_api/indexation/asset_metadata.rs#L97-L156) boilerplate.

Follow-up PR: https://github.com/FuelLabs/fuel-core/pull/2616

## Checklist
- [X] Breaking changes are clearly marked as such in the PR description and changelog

### Before requesting review


fuel-tx/src/receipt.rs
```diff
@@ -156,6 +156,51 @@ pub enum Receipt {
     },
 }
 
+impl core::fmt::Display for Receipt {
```
xgreenx, line 159: Why can't we derive the implementation with all fields?=D

rafal-ch: Regarding all fields - I thought it may be just a bit too much details, which, when needed, could be retrieved by `Debug`. I prefer `Display` to be more concise, targeted on humans, not developers :)

I have no strong opinions here, so if you prefer to have all fields, I can just stick `derive_more::Display` without any custom `#[display(fmt =` attributes.

Let me know what you think.

fuel-tx/src/receipt.rs
```diff
@@ -156,6 +156,51 @@ pub enum Receipt {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Receipt::Call { id, .. } => write!(f, "Call({})", id),
+            Receipt::Return { id, .. } => write!(f, "Return({})", id),
+            Receipt::ReturnData { id, .. } => write!(f, "ReturnData({})", id),
+            Receipt::Panic { id, .. } => write!(f, "Panic({})", id),
```
rymnc, line 165: do you think we can display the panic reason in a meaningful way?

https://github.com/FuelLabs/fuel-vm/blob/2fc9c5d21d1f1ac1c4ec2e0878566c15f4f653db/fuel-tx/src/receipt.rs#L65

rafal-ch: Yes, since `PanicReason` is a simple `repr(u8)` enum, we can just debug-print it to get an output like:
```
Panic(id=0000000000000000000000000000000000000000000000000000000000000000, reason=OutOfGas)
```

Updated in https://github.com/FuelLabs/fuel-vm/pull/901/commits/7e04606ea5561615b445fd907f73d5e8cfc4c203

rymnc: nice, ty

xgreenx (approved):

Issue #888, opened 2025-01-14
Add gas price change percent to the consensus parameters


rymnc:
ref: https://github.com/FuelLabs/fuel-core/pull/2547/files#r1913972358

PR #905, opened 2025-01-27
Add `block_max_transactions` to `ConsensusParameters`

AurelienFT:
Request from : https://github.com/FuelLabs/fuel-vm/pull/899#pullrequestreview-2574624691

## Checklist

### Before requesting review

### DIFF
--- a/fuel-tx/src/transaction/consensus_parameters.rs
+++ b/fuel-tx/src/transaction/consensus_parameters.rs
@@ -28,12 +28,19 @@ pub struct SettingBlockTransactionSizeLimitNotSupported;
 #[cfg(feature = "std")]
 impl std::error::Error for SettingBlockTransactionSizeLimitNotSupported {}
 
+#[derive(Debug, derive_more::Display)]
+#[display("setting block max transactions is not supported")]
+pub struct SettingBlockMaxTransactionsNotSupported;
+#[cfg(feature = "std")]
+impl std::error::Error for SettingBlockMaxTransactionsNotSupported {}
+
 /// A versioned set of consensus parameters.
 #[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
 pub enum ConsensusParameters {
     /// Version 1 of the consensus parameters
     V1(ConsensusParametersV1),
     V2(ConsensusParametersV2),
+    V3(ConsensusParametersV3),
 }
 
 #[cfg(feature = "test-helpers")]
@@ -68,9 +75,10 @@ impl ConsensusParameters {
         base_asset_id: AssetId,
         block_gas_limit: u64,
         block_transaction_size_limit: u64,
+        block_max_transactions: u64,
         privileged_address: Address,
     ) -> Self {
-        Self::V2(ConsensusParametersV2 {
+        Self::V3(ConsensusParametersV3 {
             tx_params,
             predicate_params,
             script_params,
@@ -81,6 +89,7 @@ impl ConsensusParameters {
             base_asset_id,
             block_gas_limit,
             block_transaction_size_limit,
+            block_max_transactions,
             privileged_address,
         })
     }
@@ -90,6 +99,7 @@ impl ConsensusParameters {
         match self {
             Self::V1(params) => &params.tx_params,
             Self::V2(params) => &params.tx_params,
+            Self::V3(params) => &params.tx_params,
         }
     }
 
@@ -98,6 +108,7 @@ impl ConsensusParameters {
         match self {
             Self::V1(params) => &params.predicate_params,
             Self::V2(params) => &params.predicate_params,
+            Self::V3(params) => &params.predicate_params,
         }
     }
 
@@ -106,6 +117,7 @@ impl ConsensusParameters {
         match self {
             Self::V1(params) => &params.script_params,
             Self::V2(params) => &params.script_params,
+            Self::V3(params) => &params.script_params,
         }
     }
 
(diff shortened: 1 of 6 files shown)


AurelienFT:
@xgreenx I saw https://github.com/FuelLabs/fuel-vm/issues/888 what do you think about adding this new parameter inside this ConsensusParametersV3 version ?

xgreenx:
Yes, we also can add this one to consensus parameters=)

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -166,6 +195,20 @@ impl ConsensusParameters {
...
+    }
+
+    /// Get the block max transactions
+    pub fn block_max_transactions(&self) -> u64 {
+        match self {
+            Self::V1(_) | Self::V2(_) => {
+                // In V1 and V2 there was no limit on the transaction size. For the sake
+                // of backwards compatibility we allow for a largest limit
+                // possible.
+                u64::MAX
```
xgreenx, line 209:
```suggestion
                u16::MAX as u64
```

fuel-tx/src/transaction/consensus_parameters/gas_price_metadata.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub new_exec_gas_price: u64,
+    pub min_exec_gas_price: u64,
+    pub exec_gas_price_change_percent: u16,
+    pub l2_block_fullness_threshold_percent: u8,
+    // TODO:We don't need this after we implement
+    // https://github.com/FuelLabs/fuel-core/issues/2481
+    pub gas_price_factor: NonZeroU64,
+    pub min_da_gas_price: u64,
+    pub max_da_gas_price: u64,
+    pub max_da_gas_price_change_percent: u16,
```
xgreenx, line 14: We only need this field, all other fields shouldn't be part of the consensus

fuel-tx/src/transaction/consensus_parameters/gas_price_metadata.rs
```diff
@@ -0,0 +1,42 @@
+use core::num::NonZeroU64;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct GasPriceMetadata {
```
xgreenx, line 4: Maybe you want make it an enum

fuel-tx/src/transaction/consensus_parameters/gas_price_metadata.rs
rymnc: i want to disagree with this. while keeping just this one field is nice to ensure compatibility with all other operators, we have to make sure that all parameters with which the block producer determines the gas price (p, d, gas price factor etc) is included in the consensus parameters for transparency sake

fuel-tx/src/transaction/consensus_parameters/gas_price_metadata.rs
AurelienFT: Done

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -388,6 +486,72 @@ impl From<ConsensusParametersV2> for ConsensusParameters {
...
+    pub chain_id: ChainId,
+    pub gas_costs: GasCosts,
+    pub base_asset_id: AssetId,
+    pub block_gas_limit: u64,
+    pub block_transaction_size_limit: u64,
+    pub block_max_transactions: u64,
+    /// The privileged address(user or predicate) that can perform permissioned
+    /// operations(like upgrading the network).
+    pub privileged_address: Address,
+    gas_price_metadata: GasPriceMetadata,
```
xgreenx, line 509: Why is it private?

fuel-tx/src/transaction/consensus_parameters/gas_price_metadata.rs
xgreenx: We don't know what algorithm we want to use for the gas price on consensus level. Right now, only one entity - the block producer, decides how the gas price is calculated. On the consensus level, we just agree that gas price shouldn't increase more than X %.

It is done for flexibility and to avoid setting the algorithm in stone.

rymnc: if that is the case then perhaps we should version this field in the consensus parameters. what do you think?

fuel-tx/src/transaction/consensus_parameters.rs
AurelienFT: Error, changed.

fuel-tx/src/transaction/consensus_parameters.rs
```diff
@@ -388,6 +486,72 @@ impl From<ConsensusParametersV2> for ConsensusParameters {
...
+    /// The privileged address(user or predicate) that can perform permissioned
+    /// operations(like upgrading the network).
+    pub privileged_address: Address,
+    pub gas_price_metadata: GasPriceMetadata,
+}
+
+#[cfg(feature = "test-helpers")]
+impl ConsensusParametersV3 {
+    const DEFAULT_BLOCK_MAX_TRANSACTIONS: u64 = u64::MAX;
+    const DEFAULT_BLOCK_TRANSACTION_SIZE_LIMIT: u64 = 126 * 1024;
```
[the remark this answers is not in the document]
AurelienFT: it's a copy paste from your test-helpers for consensus parameter v2 I beleive it's just an arbitrary value.

PR #908, opened 2025-02-07
Bugfix: more receipt generation issues with debugger

Dentosal:
While #889 fixed some receipt generation issues it missed regenerating receipts_root after the debugger is done. Also the in-memory outputs were updated too early.

## Checklist

### Before requesting review

### DIFF
--- a/fuel-vm/src/interpreter/executors/main.rs
+++ b/fuel-vm/src/interpreter/executors/main.rs
@@ -958,7 +958,7 @@ where
     pub(crate) fn run_program(
         &mut self,
     ) -> Result<ProgramState, InterpreterError<S::DataError>> {
-        let Some(script) = self.transaction().as_script() else {
+        let Some(script) = self.tx.as_script() else {
             unreachable!("Only `Script` transactions can be executed inside of the VM")
         };
         let gas_limit = *script.script_gas_limit();
@@ -1045,7 +1045,11 @@ where
             gas_price,
         )?;
         self.update_transaction_outputs()?;
-        *self.tx.as_script_mut().unwrap().receipts_root_mut() = self.receipts.root();
+
+        let Some(script) = self.tx.as_script_mut() else {
+            unreachable!("This is checked to hold in the beginning of this function");
+        };
+        *script.receipts_root_mut() = self.receipts.root();
 
         Ok(state)
     }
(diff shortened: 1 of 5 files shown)


fuel-vm/src/interpreter/executors/main.rs
```diff
@@ -1042,16 +1044,11 @@ where
             &self.balances,
             gas_price,
         )?;
+        self.update_transaction_outputs()?;
+        *self.tx.as_script_mut().unwrap().receipts_root_mut() = self.receipts.root();
```
xgreenx, line 1048: Whiel it is true that we run only script, I still prefer to sue `if let Some(script)`=)

fuel-vm/src/interpreter/internal.rs
```diff
@@ -89,8 +89,8 @@ pub(crate) fn update_memory_output<Tx: ExecutableTransaction>(
         .ok_or(PanicReason::OutputNotFound)?;
     let mut mem = memory.write_noownerchecks(range.start, range.len())?;
     let output = tx
-        .outputs_mut()
-        .get_mut(idx)
+        .outputs()
+        .get(idx)
         .expect("Invalid output index; checked above");
     output
```
xgreenx, line 95: I remember why it was mut. Because `encode` before require mutable reference...) After refactoring of encoding and decoding now it is not an issue

Dentosal:
Taking this back to draft until https://github.com/FuelLabs/fuel-core/pull/2661 is merged.

fuel-vm/src/interpreter/executors/main.rs
Dentosal: At the top of the function we error if `as_script()` cannot be used. We only access it again like this to satisfy borrow checker.

MitchTurner: I agree. I don't think we should `unwrap` anywhere in our prod code. I prefer the type-safe approach too--`expect` is more acceptable than `unwrap` but it's fragile still.

Dentosal: Fixed in 690a381e1.

xgreenx (approved):

netrome (approved):

Issue #886, opened 2025-01-06
Expose non-normalized signatures from fuel-crypto


Dentosal:
See https://github.com/FuelLabs/fuel-core/pull/1922#discussion_r1901912701

PR #909, opened 2025-02-11
Add the `remove_recovery_id()` to `Signature`

rafal-ch:
This PR adds the `remove_recovery_id()` function to the `Signature`. Recovery ID must be removed from the signature in some use cases and this is done manually in `fuel-core`, for example [here](https://github.com/FuelLabs/fuel-core/blob/master/crates/services/shared-sequencer/src/lib.rs#L258-L260). Therefore, there will be a follow-up PR for the `fuel-core` created.

Closes: https://github.com/FuelLabs/fuel-vm/issues/886

## Checklist
- [X] New behavior is reflected in tests

### Before requesting review

### DIFF
--- a/fuel-crypto/src/secp256/signature.rs
+++ b/fuel-crypto/src/secp256/signature.rs
@@ -1,4 +1,7 @@
-use super::backend::k1;
+use super::{
+    backend::k1,
+    signature_format::decode_signature,
+};
 use crate::{
     Error,
     Message,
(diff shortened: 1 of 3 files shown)


fuel-crypto/src/secp256/signature.rs
```diff
@@ -137,3 +146,23 @@ impl Signature {
...
+mod tests {
+    use crate::Signature;
+    use test_case::test_case;
+
+    #[test_case(0x00 => 0x00)]
+    #[test_case(0x7E => 0x7E)]
+    #[test_case(0x7F => 0x7F)]
+    #[test_case(0x80 => 0x00)]
+    #[test_case(0xFF => 0x7F)]
+    fn removes_recovery_id(s_byte: u8) -> u8 {
```
Dentosal, line 160: This test feels like it has wrong layer of abstraction, but I cannot figure out how to test this on a higher level

Dentosal (approved):

rymnc (approved):

PR #910, opened 2025-02-12
Add fix serailization/deserialization of Policies in compressed transactions to be backward compatible

AurelienFT:
We changed the `compress` `decompress` implementation to not be the default one which would lead to use the default serde implementation for each field. Instead we use a custom implementation that will use the custom serde implementations we have for `Policies`

## Checklist

### Before requesting review


xgreenx (approved):

rymnc (approved):
can confirm this fix works. :)

PR #915, opened 2025-02-19
Dry run mode: continue-on-error handling for "contract not in inputs" 

Dentosal:
~~Requires #918~~

Work towards #691

Adds a way to continue vm execution after each error instead of terminating immediately. This can be used to collect multiple execution errors at once, which is useful if you don't know in advance what storage, inputs or outputs the tx is going to need access to.

Names are still subject to bikeshedding.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [ ] If performance characteristic of an instruction change, update gas costs as well or make a follow-up PR for that
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

Issue #691, opened 2024-03-05
Support dry run mode in the FuelVM


xgreenx:
## Problem overview

The FuelVM has many rules regarding transaction validity and requirements for using specific opcodes. Each opcode that touches the contract requires this contract's existence in the transactions' inputs. Another example is the `tro` opcode that requires the existence of the `Variable` output.

All those rules make the usage harder and require dry-running transactions many times on the SDK side to fulfill all rules.

## Solution

The FuelVM may support dry run mode, in which all validity rules will not abort the execution immediately. Instead, we will keep a record of what rules are failing and provide feedback at the end of the execution with all problems.

xgreenx:
We don't have plans to work on it before mainnet since it may bring vulnerabilities.

miladhatami1393:
Good job

MitchTurner:
I think we need to be clearer about what the implementation would look like. And it also could be useful to get some feedback from SDK people about what features they would most like.

PR #918, opened 2025-02-25
continue-on-error foundations: refactor verification to a separate trait

Dentosal:
Work towards #691

Follow-up: #915

Moves some parts of tx verification logic to a separate trait so it can be replaced. This initial PR only moves "is contract in inputs" check. More rules can be migrated as they are needed.

This is a breaking change, as we need to add a new type parameter to the VM. While the default value keeps the same behavior, existing trait implementations might still break.

The new trait is fully sealed: it cannot be implemented outside fuel-vm, and it's methods cannot be called outside it. This means adding or removing methods is not a breaking change in the future.

In addition, an internal `InputContracts` type is removed and it's field flattened on the containing structs.

## Checklist
- [x] New behavior is reflected in tests: no behavior changes!

### Before requesting review


fuel-vm/src/interpreter/contract.rs
```diff
@@ -174,36 +175,44 @@ where
...
+        self,
         result: &mut Word,
         b: Word,
         c: Word,
-    ) -> IoResult<(), S::Error>
+    ) -> IoResult<(), <S as StorageInspect<ContractsAssets>>::Error>
     where
         S: ContractsAssetsStorage,
+        S: InterpreterStorage,
+        V: Verifier<S>,
```
xgreenx, line 197:
```suggestion
    ) -> IoResult<(), S::Error>
    where
        S: ContractsAssetsStorage,
        V: Verifier<S>,
```

fuel-vm/src/interpreter/contract.rs
```diff
@@ -212,37 +221,45 @@ struct TransferCtx<'vm, S, Tx> {
...
         recipient_contract_id_offset: Word,
         transfer_amount: Word,
         asset_id_offset: Word,
-    ) -> IoResult<(), S::Error>
+    ) -> IoResult<(), <S as StorageInspect<ContractsAssets>>::Error>
     where
         Tx: ExecutableTransaction,
         S: ContractsAssetsStorage,
+        S: InterpreterStorage,
+        V: Verifier<S>,
```
xgreenx, line 251:
```suggestion
    ) -> IoResult<(), S::Error>
    where
        Tx: ExecutableTransaction,
        S: ContractsAssetsStorage,
        V: Verifier<S>,
```

fuel-vm/src/verification.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Do not allow outside implementations for the Verifier, so that it's not a breaking
+/// change to modify it.
+trait Seal {}
+
+/// What to do when verification fails.
+#[allow(private_bounds)] // For selaed trait
+pub trait Verifier<S>
+where
+    Self: Sized + Seal,
+    S: InterpreterStorage,
```
xgreenx, line 26: Do we need `S` here at all? Based on the use cases that we discussed, it looks like we don't need it anymore=)

fuel-vm/src/interpreter/flow.rs
```diff
@@ -449,22 +452,23 @@ struct PrepareCallCtx<'vm, S> {
...
-{
+impl<S, V> PrepareCallCtx<'_, S, V> {
     fn prepare_call(mut self) -> IoResult<(), S::DataError>
     where
         S: StorageSize<ContractsRawCode>
             + ContractsAssetsStorage
             + StorageRead<ContractsRawCode>
             + StorageAsRef,
+        S: InterpreterStorage,
+        V: Verifier,
```
xgreenx, line 471:
```suggestion
        S: InterpreterStorage,
        V: Verifier,
```

fuel-vm/src/verification.rs
```diff
@@ -0,0 +1,61 @@
+//! Stategies for verifying the correctness of the VM execution.
+//! The default strategy, [`Panic`], simply panics on failed verification.
```
xgreenx, line 2:
```suggestion
//! The default strategy, [`Normal`], simply returns an error on failed verification.
```

xgreenx (commented):
Looks good to me=)

xgreenx (dismissed):

xgreenx (approved):

rymnc (approved):

PR #923, opened 2025-03-06
Narrow integer load/store operations

Dentosal:
Spec PR: https://github.com/FuelLabs/fuel-specs/pull/628

Adds `LQW` `LHW` `SQW` `SHW` instructions to store and load quarter and half words efficiently. They follow the same conventions as their pre-existiing counterparts `LB` `LW` `SB` `SW`. The new instructions have same cost as their full-word counterparts.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog: None!
- [ ] If performance characteristic of an instruction change, update gas costs as well or make a follow-up PR for that: using gas costs of existing instructions as a baseline
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior: https://github.com/FuelLabs/fuel-specs/pull/628

### Before requesting review


fuel-vm/src/interpreter/executors/opcodes_impl.rs
```diff
@@ -1637,7 +1637,45 @@ where
...
+    S: InterpreterStorage,
+    Tx: ExecutableTransaction,
+    Ecal: EcalHandler,
+    V: Verifier,
+{
+    fn execute(
+        self,
+        interpreter: &mut Interpreter<M, S, Tx, Ecal, V>,
+    ) -> IoResult<ExecuteState, S::DataError> {
+        interpreter.gas_charge(interpreter.gas_costs().lw())?;
```
AurelienFT, line 1657: It's fine to use the same gas_costs (the same value ok, but talking about the exact same field), it can't cause any issue elsewhere to not have gas defined for this opcode ? (Just double checking)

fuel-vm/src/tests/alu.rs
```diff
@@ -110,7 +110,9 @@ impl AluResultForFlags {
         }
     }
 
-    /// The operation wraps around if wrapping is enabled, otherwise it panics
+    /// The operation wraps around if wrapping is enabled, otherwise it panics.
+    /// Computes the wrapping value for u64 oveflow. For narrow ints, use
+    /// `wrapping_ok_narrow`.
```
AurelienFT, line 115: Does `wrapping_ok_narrow` exists ? I don't find it in codebase.

AurelienFT (approved):
Two little personal questions but if there is no issue raised then I'm ok with the PR. I like the new tests :p

xgreenx (commented):
@Dentosal Could you post benchmark results please? Just to show the impact of the change

Dentosal:
> @Dentosal Could you post benchmark results please? Just to show the impact of the change

This isn't really done for performance or code size reasons, but to help the compiler team with implementing the trickier details of stdlib implementation. I have a modified sway compiler in a way that shows minimum impact, but we can get way more out of these if we do advanced packing of the data section as well. (i.e. substring search, and ordering it in a way that produces the best substrings). Also this should make some optimizations with store ops possible, but those haven't been explored yet. I've benchmarked with the minimum impact version here:

The benchmark results for `mira-v1-core` (release build):

latest sway `8e162c966` | after changes | reduction
--|--|--
89.384 KB | 88.764 KB | 0.7%

fuel-vm/src/interpreter/executors/opcodes_impl.rs
Dentosal: My understanding is yes, but maybe @xgreenx knows about some issues?

AurelienFT (approved):

AurelienFT (dismissed):

fuel-asm/src/op.rs
```diff
@@ -312,7 +342,12 @@ const _: () = {
...
     /// Construct a `WQDV` instruction from its arguments.
-    pub fn wqdv_args(ra: u8, rb: u8, rc: u8, args: DivArgs) -> typescript::Instruction {
+    pub fn wqdv_args(
+        ra: u8,
+        rb: u8,
+        rc: u8,
+        args: wideint::DivArgs,
+    ) -> typescript::Instruction {
         crate::op::wqdv_args(ra, rb, rc, args).into()
     }
```
Voxelot, line 352: We should also add exports for the new instructions here to support them in the TS-SDK.

fuel-vm/src/interpreter/memory/allocation_tests.rs
```diff
@@ -150,141 +150,6 @@ fn test_stack_pointer_overflow(
     Ok(())
 }
 
-#[test_case(20, 20 => Ok(()); "Can load a byte")]
```
Voxelot, line 153: Why were these tests removed?

fuel-vm/src/tests/memory.rs
```diff
@@ -102,6 +103,125 @@ fn test_lb() {
     assert_eq!(1, result);
 }
 
+#[test_case(op::sb, 0, 0, 0 => RunResult::Success([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]))]
```
Voxelot, line 106: Please consider grouping some of these test cases by labels or comments to make it easier to understand the intention and coverage of these tests.

Voxelot:
From my side the only blocker is adding proper support for the WASM constructors of these new instructions.

fuel-vm/src/interpreter/memory/allocation_tests.rs
Dentosal: They were testing the internal implementation details, instead of the actual functionality. There is [ongoing effort](https://github.com/FuelLabs/fuel-vm/issues/746) to switch away from these anyway, so this was the natural moment to do that. The instructions are already tested in `fuel-vm/src/tests/memory.rs`.

fuel-vm/src/tests/memory.rs
Dentosal: Done in b65d73ae9

fuel-asm/src/op.rs
Dentosal: I don't think the exports are missing? They're automatically exported by 

```
pub use super::_op::*;
```

(and I see them in rustdoc generated by `cargo doc -p fuel-asm --features typescript`)

Voxelot: Ah I see it now, looks like these are just special helpers for a select few opcodes but not required for all of them 👍🏻

Voxelot (approved):

AurelienFT (approved):

PR #925, opened 2025-03-14
Jump-and-link instruction

### LINKED ISSUE #857: New function call/return helper opcodes
~~`$is` being the basis of jumps is something we should get rid of. I'd like to introduce two new opcodes:~~
* ~~`JMPA $reg imm` : jump to absolute address in register plus immediate~~
* ~~`SRAJ $reta reg/imm` store return address and jump to "function"~~

See Spec PR: https://github.com/FuelLabs/fuel-specs/pull/630

Dentosal:
Closes #857. See the spec PR for full description: https://github.com/FuelLabs/fuel-specs/pull/630.

Adds `JAL $rA $rB imm` instruction that sets `$ra = $pc + 4` and then jump setting `$pc = $rB + imm * 4`. Allows `$rA` to be `$zero` in which case the value is discarded instead.

Open questions:
* Do we want to do overflow checking when setting `$rA`? Since we're always within valid memory before the jump, worst that can happen that the panic is delayed until trying to use that address.
* Do we want to support discard-writing in other opcodes as well? Seems like it's rarely useful in other cases.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog: none!
- [x] New behavior is reflected in tests: yep!
- [x] If performance characteristic of an instruction change, update gas costs as well or make a follow-up PR for that: using costs of the jmp opcode for this, should be about the same
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior: https://github.com/FuelLabs/fuel-specs/pull/630

### Before requesting review


fuel-vm/src/interpreter/executors/opcodes_impl.rs
```diff
@@ -1358,6 +1360,37 @@ where
...
+        interpreter: &mut Interpreter<M, S, Tx, Ecal, V>,
+    ) -> IoResult<ExecuteState, S::DataError> {
+        interpreter.gas_charge(interpreter.gas_costs().jmp())?;
+        let (reg_ret_addr, reg_target, offset) = self.unpack();
+
+        // Storing return address to zero register discards it instead
+        // While we use saturating_add here, the PC shoudln't ever have values above the
+        // memory size anyway
+        let ret_addr =
+            interpreter.registers[RegId::PC].saturating_add(Instruction::SIZE as u64);
```
netrome, line 1382: Are we certain the interpreter will always have the `RegId::PC` register available? How do we maintain that invariant?

netrome (dismissed):
This should be reviewed by someone who understands the VM better than I, but I don't spot any obvious errors at least.

fuel-vm/src/interpreter/executors/opcodes_impl.rs
Dentosal: Yep, we're certain. There are hundreds of tests in this repo which all rely on that fact. The field is defined as`registers: [Word; VM_REGISTER_COUNT]`, and PC along with every other system register has an index below `VM_REGISTER_COUNT`.

netrome: That's good, thanks for the pointer! :pray:

fuel-vm/src/tests/jump_and_link.rs
```diff
@@ -0,0 +1,167 @@
...
+        op::subi(reg_tmp, reg_tmp, 1),
+        op::log(reg_tmp, RegId::ZERO, RegId::ZERO, RegId::ZERO),
+        op::ret(RegId::ONE),
+    ];
+
+    let receipts = run_script(script);
+    assert_success(&receipts);
+
+    if let Some(Receipt::Log { ra, .. }) = receipts.first() {
+        assert_eq!(*ra, skip, "Expected correct number of skipped instructions");
```
xgreenx, line 95: Shouldn't `ra` be `2` here since we skipped 3 instructions?

fuel-vm/src/tests/jump_and_link.rs
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+        panic!("Expected a log receipt");
+    };
+}
+
+#[test]
+fn jump_and_link__recursive_fibonacci() {
```
xgreenx, line 102: I will skip this function from the review=D

Looks like it works, so very cool that you managed to write it with just instructions=)

xgreenx (dismissed):

fuel-vm/src/tests/jump_and_link.rs
Dentosal: No. We compute  `5-1-1 = 3`. If that feels confusing, the test could be rewritten to use `addi` instead. Let me know if you'd prefer that.

fuel-asm/src/lib.rs
```diff
@@ -308,6 +308,8 @@ impl_instructions! {
     0x97 POPL popl [bitmask: Imm24]
     "Pop a bitmask-selected set of registers in range 40..64 to the stack."
     0x98 POPH poph [bitmask: Imm24]
+    "Store return address and jump to an absolute address."
+    0x99 JAL jal [ret_addr: RegId target: RegId offset: Imm12]
```
Voxelot, line 312: We might want something like:

```
0x99 JAL jal [ret_addr: RegId target: Imm12]
0x9A JALR jalr [ret_addr: RegId target: RegId offset: Imm12]
```

fuel-vm/src/interpreter/flow/jump_tests.rs
```diff
@@ -8,7 +8,7 @@ use crate::error::PanicOrBug;
 #[test_case(0, 80, 10 => Ok(40); "jump backwards")]
 #[test_case(0, 40, VM_MAX_RAM => Err(PanicOrBug::Panic(PanicReason::MemoryOverflow)); "jump too far forward")]
 fn test_absolute_jump(is: Word, mut pc: Word, j: Word) -> SimpleResult<Word> {
-    JumpArgs::new(JumpMode::Absolute)
+    JumpArgs::new(JumpMode::RelativeIS)
```
Voxelot, line 11: Should we add test cases in this file for `JumpMode::Assign`?

fuel-vm/src/interpreter/flow.rs
```diff
@@ -253,11 +253,13 @@ pub(crate) fn revert(
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum JumpMode {
-    /// `$pc = $is + address`
-    Absolute,
-    /// `$pc = $pc + address`
+    /// `$pc = reg + (imm * instruction_size)`
```
Voxelot, line 256:
```suggestion
    /// `$pc = (reg + imm) * instruction_size`
```

fuel-vm/src/tests/jump_and_link.rs
```diff
@@ -0,0 +1,167 @@
...
+#[test]
+fn jump_and_link__immediate_count_is_instructions() {
+    let reg_return_addr = RegId::new(0x11);
+    let reg_tmp = RegId::new(0x12);
+
+    let skip = 3; // Jump over the next 3 instructions
+
+    let script = vec![
+        op::movi(reg_tmp, 5),
+        op::jal(reg_return_addr, RegId::PC, (skip + 1) as _), /* Zero would mean
```
Voxelot, line 79: Should we disallow imm == 0 if rB == PC?

Voxelot (approved):

Voxelot:
> Do we want to support discard-writing in other opcodes as well? Seems like it's rarely useful in other cases.

Maybe we should analyze some sway compiler output examples to see if there are many cases of registers being allocated for capturing results and never getting used afterwards?

Dentosal:
> Maybe we should analyze some sway compiler output examples to see if there are many cases of registers being allocated for capturing results and never getting used afterwards? This should likely be done as a separate cleanup PR anyways.

I did quickly check some compiler output, and it seems either super rare or nonexisting. And yep definitely a separate PR in any case.

fuel-vm/src/tests/jump_and_link.rs
Dentosal: I don't see any point in  disallowing that. You're just going to run out of gas.

fuel-asm/src/lib.rs
Dentosal: I don't see much reason to do this. The form above with only immediate wouldn't be too useful, since you can already do that with setting target register to `Reg::ZERO`. Having a wider immediate might help in some huge programs, but only a little bit.

fuel-vm/src/interpreter/flow/jump_tests.rs
Dentosal: Good point. Added in bb9d2161d.

Voxelot (dismissed):

fuel-vm/src/interpreter/flow/jump_tests.rs
Voxelot: thanks!

fuel-asm/src/lib.rs
Voxelot: makes sense, seems like a very minor/niche optimization

xgreenx (dismissed):

Voxelot (approved):

xgreenx (approved):

PR #933, opened 2025-04-08
Don't update balance if change is zero

### LINKED ISSUE #866: When `amount` is zero do not update the balance
The current code always increases and decreases the balance even if we have no updates. It means storage read and write and spends a lot of resources for nothing.

image

Dentosal:
Closes #866.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog: not breaking, I hope. Changes observable behavior though.
- [ ] New behavior is reflected in tests: nope! it wasn't tested before, it won't be tested after!
- [ ] If performance characteristic of an instruction change, update gas costs as well or make a follow-up PR for that: the worst case performance remains unchanged
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior: no changes needed

### Before requesting review


fuel-vm/src/interpreter/contract.rs
```diff
@@ -409,26 +409,29 @@ where
...
 /// Increase the asset balance for a contract.
-/// Returns new balance, and a boolean indicating if a new entry was created.
+/// A boolean indicating if a new entry was created.
 pub fn balance_increase<S>(
     storage: &mut S,
     contract: &ContractId,
     asset_id: &AssetId,
     amount: Word,
-) -> IoResult<(Word, bool), S::Error>
+) -> IoResult<bool, S::Error>
```
netrome, line 418: Since we change the public interface, I suppose this is a breaking change. Do we have any idea of potential downstream impact of this?

netrome (changes requested):
Looks good but i think this counts as a breaking change and should be marked as such in the changelog.

Also it would be nice to see a test case verifying this behavior (not a hard blocker imo though).

rymnc (approved):

xgreenx (approved):

PR #934, opened 2025-04-09
Include input_index of the predicate that failed verification

Dentosal:
@luizstacio posted on slack:
> It would be good to have the input index of the predicate that failed on this error: Invalid transaction data: `PredicateVerificationFailed(Panic(PredicateReturnedNonOne))`  Right now, with many predicates on a single TX, it is difficult for me to say which one is returning false.

The PR adds `input_index` field to `PredicateVerificationFailed` variants that need use it.

This breaks serde deserialization of old `PredicateVerificationFailed` into the new type, but I don't think we depend on that anywhere. Maybe @xgreenx knows about this one?

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/fuel-vm/src/checked_transaction.rs
+++ b/fuel-vm/src/checked_transaction.rs
@@ -288,7 +288,13 @@ pub enum CheckError {
     /// The transaction doesn't pass validity rules.
     Validity(ValidityError),
     /// The predicate verification failed.
-    PredicateVerificationFailed(PredicateVerificationFailed),
+    PredicateVerificationFailed {
+        /// Input index of the failed predicate.
+        /// `TransactionExceedsTotalGasAllowance` sets this to `0`.
+        input_index: u16,
+        /// The reason of the failure.
+        reason: PredicateVerificationFailed,
+    },
     /// The max fee used during checking was lower than calculated during `Immutable`
     /// conversion
     InsufficientMaxFee {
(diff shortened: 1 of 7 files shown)


fuel-vm/src/checked_transaction.rs
```diff
@@ -288,7 +288,13 @@ pub enum CheckError {
     /// The transaction doesn't pass validity rules.
     Validity(ValidityError),
     /// The predicate verification failed.
-    PredicateVerificationFailed(PredicateVerificationFailed),
+    PredicateVerificationFailed {
+        /// Input index of the failed predicate.
+        /// `TransactionExceedsTotalGasAllowance` sets this to `0`.
+        input_index: usize,
```
AurelienFT, line 294: If predicate index can only be u16 maybe we could use u16::MAX + 1 for general error. Or have an option. Because, currently 0 can be general or predicate 0 which is two different things

Dentosal: That actually just felt too ugly. Instead I just rewrote the whole thing in 223e24819. to include the fields directly into  `PredicateVerificationFailed` variants.

fuel-vm/src/error.rs
```diff
@@ -310,20 +339,20 @@ impl From<Bug> for PredicateVerificationFailed {
...
-    }
-}
+// impl From<PanicOrBug> for PredicateVerificationFailed {
+//     fn from(err: PanicOrBug) -> Self {
+//         match err {
+//             PanicOrBug::Panic(reason) => Self::from(reason),
+//             PanicOrBug::Bug(bug) => Self::Bug(bug),
+//         }
+//     }
+// }
```
xgreenx, line 355: I guess we can remove it=D

fuel-vm/src/error.rs
```diff
@@ -310,20 +339,20 @@ impl From<Bug> for PredicateVerificationFailed {
...
-        match err {
-            PanicOrBug::Panic(reason) => Self::from(reason),
-            PanicOrBug::Bug(bug) => Self::Bug(bug),
-        }
-    }
-}
+// impl From<PanicOrBug> for PredicateVerificationFailed {
+//     fn from(err: PanicOrBug) -> Self {
+//         match err {
+//             PanicOrBug::Panic(reason) => Self::from(reason),
```
AurelienFT, line 351: Dead code ?

fuel-vm/src/error.rs
Dentosal: Yep =D afdcd4e13

xgreenx (approved):

AurelienFT (approved):

PR #938, opened 2025-04-14
feat: Allow empty values in sparse merkle trees

### LINKED ISSUE #937: Sparse merkle tree treats empty values as deletions in update method
### Problem
When attempting to insert empty values in a sparse merkle tree, the tree treats it as a deletion instead of an insertion. 

```rust
    pub fn update(
        &mut self,
        key: MerkleTreeKey,
        data: &[u8],
    ) -> Result<(), MerkleTreeError<StorageError>> {
        if data.is_empty() {
            // If the data is empty, this signifies a delete operation for the
            // given key.
            self.delete(key)?;
            return Ok(())
        }
```

This poses a problem for the proof system where we want to merkleize the `ProcessedTransactions` table (among others), which uses empty values `()`. When attempting to generate inclusion proofs for keys, the SMT implementation returns exclusion proofs.

After discussing this problem with @xgreenx we agreed that the merkle tree implementation should accept insertions of empty values as insertions and not deletions.

### Definition of Done
Sparse merkle trees can hold empty values.

### Note
While we're at it, we should probably rename the `update` function to `insert`.

netrome:
closes #937 

This PR removes the problematic check for empty values in the SMT `update` function (and the equivalent filtering in `from_set`). This makes SMTs able to commit to empty values, which we need for the proof system for the `ProcessedTransactions` table.

Additionally, to better reflect the behavior I renamed the `update` method to `insert`.

### DIFF
--- a/.changes/breaking/938.md
+++ b/.changes/breaking/938.md
@@ -1,0 +1,1 @@
+Rename `update` method to `insert` on sparse merkle tree, and allow inserting empty values.
--- a/fuel-merkle/src/sparse/in_memory.rs
+++ b/fuel-merkle/src/sparse/in_memory.rs
@@ -189,7 +189,7 @@ impl MerkleTree {
     }
 
     pub fn update(&mut self, key: MerkleTreeKey, data: &[u8]) {
-        let _ = self.tree.update(key, data);
+        let _ = self.tree.insert(key, data);
     }
 
     pub fn delete(&mut self, key: MerkleTreeKey) {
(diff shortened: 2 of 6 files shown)


netrome:
Note: Before merging, I want to integrate the change in `fuel-core` to ensure we don't hit any unexpected problems with this change.

netrome:
> Note: Before merging, I want to integrate the change in `fuel-core` to ensure we don't hit any unexpected problems with this change.

Confirmed now, we should be good to go.

AurelienFT (approved):
LGTM

rymnc (approved):

Dentosal (approved):

PR #945, opened 2025-04-24
Enable logs in the predicates via the Sway's `__dbg!` macro

xgreenx:
TODO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [Version 0.60.2]
+
 ### Added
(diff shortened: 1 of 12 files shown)


Dentosal:
Do we really want to support this on the VM side? I feel like the custom ECAL provider should be on fuel-core side.

xgreenx:
Yeah, it would be nice if we could come up with an idea on how to make the running of predicates customizable from outside.

Main problem is that we want to have a separate ECAL instance for each predicate in order to aggregate all logs for one predicates in one tracing message.


But maybe we could introduce new generic to the `check_predicates` method, or add a new method with new generic to not break API for downstream crates.

Dentosal:
closing in favor of #963

PR #948, opened 2025-04-30
Chargeable Transaction V2

MitchTurner:
[Link to related issue(s) here, if any]

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/fuel-tx/src/transaction/types/input.rs
+++ b/fuel-tx/src/transaction/types/input.rs
@@ -245,6 +245,26 @@ pub enum Input {
     MessageDataPredicate(MessageDataPredicate),
 }
 
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    strum_macros::EnumCount,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+#[cfg_attr(
+    feature = "da-compression",
+    derive(fuel_compression::Compress, fuel_compression::Decompress)
+)]
+pub enum InputV2 {
+    Coin(CoinV2),
+    Message(MessageV2),
+    Contract(Contract),
+}
+
 impl Default for Input {
     fn default() -> Self {
         Self::contract(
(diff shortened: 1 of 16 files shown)


fuel-tx/src/transaction/types/input.rs
```diff
@@ -236,13 +236,44 @@ impl AsFieldFmt for PredicateCode {
...
+)]
+pub enum InputV2 {
+    Coin(CoinV2),
+    Message(MessageV2),
+    Contract(Contract),
+}
+
+impl From<InputV2> for InputRepr {
+    fn from(_value: InputV2) -> Self {
+        todo!()
```
Dentosal, line 275: Is this todo! intentionally left here?

fuel-tx/src/transaction/types/input.rs
```diff
@@ -236,13 +236,44 @@ impl AsFieldFmt for PredicateCode {
...
+    //   up this code significantly.
+    //   https://github.com/FuelLabs/fuel-vm/issues/956
     CoinSigned(CoinSigned),
     CoinPredicate(CoinPredicate),
     Contract(Contract),
     MessageCoinSigned(MessageCoinSigned),
     MessageCoinPredicate(MessageCoinPredicate),
     MessageDataSigned(MessageDataSigned),
     MessageDataPredicate(MessageDataPredicate),
+    InputV2(InputV2),
```
Dentosal, line 250: We discussed about flattening these to aid with canonical serialization. Is there a reason to have a separate type here? How does it affect serialization size?

MitchTurner:
> The feature-flag protection is not extended to all things here, for instance the Input::InputV2 variant isn't covered. I'm not sure how much issue this actually is, but it's already a breaking change as it allows deserializing transactions that weren't valid before. This means we cannot release these changes even when feature-flag protected. I'm not sure if this blocks us from merging into master.

I think I've corrected a lot of these spots in https://github.com/FuelLabs/fuel-vm/pull/960, so maybe we merge them before merging into `master.

> There are a lot of todo!()s left, but I assume these are mostly intentional, and they seem to be feature-flag protected.
Assuming they are all behind the feature flag, then I don't care. If I missed some spots (very possible) then I will fix those.

PR #952, opened 2025-05-06
hotfix: Check changelog job should only run on PRs

netrome:
Problem discovered during last release, where this job causes CI to fail:

https://github.com/FuelLabs/fuel-vm/actions/runs/14853766345

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,12 @@ jobs:
 
       - name: Check for a file with the PR number in a sub-directory of the .changes directory" or "no changelog" label.
         run: |
+          # Only check the changelog if we're on a PR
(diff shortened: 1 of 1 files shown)


rymnc:
this won't work, it'll cause the verifications-complete workflow to be skipped because check-changelog will be skipped. you have to do it the same way we did it in fuel-core

netrome:
> this won't work, it'll cause the verifications-complete workflow to be skipped because check-changelog will be skipped. you have to do it the same way we did it in fuel-core

Oh I see, will update.

netrome:
> this won't work, it'll cause the verifications-complete workflow to be skipped because check-changelog will be skipped. you have to do it the same way we did it in fuel-core

Updated now - thanks for spotting.

AurelienFT (approved):
oopsie

Dentosal (approved):

PR #959, opened 2025-05-16
feat: Add `Borrow` trait bounds to owned types in `Mappable` trait

### LINKED ISSUE #958: Add `Borrow` trait bound for owned keys and values in `Mappable` trait
### Background
So I've had to work around this limitation a few times now and think we can improve the `Mappable` trait.

### Context
Currently, `Mappable` is defined as:
```rust
pub trait Mappable {
    /// The key type is used during interaction with the storage. In most cases, it is the
    /// same as `Self::OwnedKey`.
    type Key: ?Sized + ToOwned;
    /// The owned type of the `Key` retrieving from the storage.
    type OwnedKey: From<<Self::Key as ToOwned>::Owned> + Clone;
    /// The value type is used while setting the value to the storage. In most cases, it
    /// is the same as `Self::OwnedValue`, but it is without restriction and can be
    /// used for performance optimizations.
    type Value: ?Sized + ToOwned;
    /// The owned type of the `Value` retrieving from the storage.
    type OwnedValue: From<<Self::Value as ToOwned>::Owned> + Clone;
}
```
The `From<<Self::Key as ToOwned>::Owned>` trait bound ensures we can transform a `Key` to an `OwnedKey`, and the equivalent for values. However, in situations where we have an `OwnedKey` (or Value) we have to make additional assumptions in order to reference the `Value` type. It would be better to add an additional `Borrow` trait bound on the owned types to ensure we are able to get value references when needed.

### Definition of Done
It is possible to get references to the non-owned type from the owned types.

netrome:
closes #958 

This PR adds a `Borrow` trait bound to the owned types in the `Mappable` trait. This should improve ergonomics in generic code where we want to use the non-owned value types from the owned ones.

Since this is a limitation in the public interface of a pretty fundamental type, we should be cautious to merge it. However, since we already have an inverse `ToOwned` bound (albeit indirectly with an intermediate `From` conversion), it should be safe to assume that all existing instances are of types where they either already implement the required `Borrow` or should implement it but haven't already.

I have verified that the latest `fuel-core` compiles with this additional bound.

### DIFF
--- a/.changes/breaking/959.md
+++ b/.changes/breaking/959.md
@@ -1,0 +1,1 @@
+Add Borrow trait bounds to owned types in Mappable trait.
--- a/fuel-storage/src/lib.rs
+++ b/fuel-storage/src/lib.rs
@@ -13,6 +13,8 @@ mod impls;
 
 extern crate alloc;
 
+use core::borrow::Borrow;
+
 use alloc::{
     borrow::{
         Cow,
@@ -46,13 +48,13 @@ pub trait Mappable {
     /// same as `Self::OwnedKey`.
     type Key: ?Sized + ToOwned;
     /// The owned type of the `Key` retrieving from the storage.
-    type OwnedKey: From<<Self::Key as ToOwned>::Owned> + Clone;
+    type OwnedKey: From<<Self::Key as ToOwned>::Owned> + Borrow<Self::Key> + Clone;
     /// The value type is used while setting the value to the storage. In most cases, it
     /// is the same as `Self::OwnedValue`, but it is without restriction and can be
     /// used for performance optimizations.
     type Value: ?Sized + ToOwned;
     /// The owned type of the `Value` retrieving from the storage.
-    type OwnedValue: From<<Self::Value as ToOwned>::Owned> + Clone;
+    type OwnedValue: From<<Self::Value as ToOwned>::Owned> + Borrow<Self::Value> + Clone;
 }
 
 /// Base read storage trait for Fuel infrastructure.
--- a/fuel-tx/src/contract.rs
+++ b/fuel-tx/src/contract.rs
@@ -21,7 +21,10 @@ use fuel_types::{
 };
 
 use alloc::vec::Vec;
-use core::iter;
+use core::{
+    borrow::Borrow,
+    iter,
+};
 
 /// The target size of Merkle tree leaves in bytes. Contract code will will be divided
 /// into chunks of this size and pushed to the Merkle tree.
@@ -147,6 +150,12 @@ impl From<&mut [u8]> for Contract {
 impl From<Contract> for Vec<u8> {
     fn from(c: Contract) -> Vec<u8> {
         c.0
+    }
+}
+
+impl Borrow<[u8]> for Contract {
+    fn borrow(&self) -> &[u8] {
+        self.0.borrow()
     }
 }
 
--- a/fuel-vm/src/storage/blob_data.rs
+++ b/fuel-vm/src/storage/blob_data.rs
@@ -1,3 +1,5 @@
+use core::borrow::Borrow;
+
 use fuel_storage::Mappable;
 use fuel_types::{
     BlobId,
@@ -53,6 +55,12 @@ impl From<&mut [u8]> for BlobBytes {
 impl From<BlobBytes> for Vec<u8> {
     fn from(c: BlobBytes) -> Vec<u8> {
         c.0
+    }
+}
+
+impl Borrow<[u8]> for BlobBytes {
+    fn borrow(&self) -> &[u8] {
+        &self.0
     }
 }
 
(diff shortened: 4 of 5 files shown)


netrome:
Verified against latest fuel-core. It compiles with this without any required modifications to any types.

AurelienFT (approved):

MitchTurner:
In general, I'm hesitant to put trait bounds on the interface side and I tend to put it on the consumer/`impl` side. i.e.
```rs
impl<T> Foobar<T> 
    where
        T: Mappable,
        <T as Mappable>::OwnedValue: Borrow<<T as Mappable>::Value> 
```

But as you point out, 
> However, since we already have an inverse ToOwned bound (albeit indirectly with an intermediate From conversion), it should be safe to assume that all existing instances are of types where they either already implement the required Borrow or should implement it but haven't already.

, I think it's safe to add this--it's not really changing the implementations.

I think this qualifies as a #3 below. 

---
While we're on the topic, I'm going to synthesize some of my thoughts:

I think there are 3 main reasons one might put constraints on the trait side:
1. Declutter the consumer signatures. e.g. If you are going to have `Foo + Bar + Serialize + DeserializeOwned`, bundling those together to clean up code.
2. Enforcing domain level concepts at the interface level. i.e. you know how the domain code is going to be used and you are preemptively enforcing that something meets requirements of what the implementations are going to need.
3. Anticipating actual requirements for the trait. If a type is used in an `async` function, it will often need the `Send` and/or `Sync`.

I think 1 should be used very carefully. It adds a lot of indirection, so I'd prefer the bundled trait to be as close to the implementation as possible.

I think 2 should be avoided. Traits should be very specific to the their requirements and the more we try to anticipate the implementation details the more we couple concepts together in confusing ways. So if a function needs a method from one trait but not another, we should only constrain for the trait method it needs.

I think 3 is fine, but we should ask ourselves if we are really doing #1 or #2. The line between these can be fuzzy.

I think the tradeoff in general of pushing the constraints closer to the `impl`emnation, is a type can have many `impl` blocks scattered around the code base. This results in a new practical problem of organizing the `impl` blocks in a coherent way--which I know no standard for.

netrome:
> In general, I'm hesitant to put trait bounds on the interface side and I tend to put it on the consumer/`impl` side. i.e.
> 
> ```rust
> impl<T> Foobar<T> 
>     where
>         T: Mappable,
>         <T as Mappable>::OwnedValue: Borrow<<T as Mappable>::Value> 
> ```
> 
> But as you point out,
> 
> > However, since we already have an inverse ToOwned bound (albeit indirectly with an intermediate From conversion), it should be safe to assume that all existing instances are of types where they either already implement the required Borrow or should implement it but haven't already.
> 
> , I think it's safe to add this--it's not really changing the implementations.
> 
> _I think this qualifies as a # 3 below._
> 
> While we're on the topic, I'm going to synthesize some of my thoughts:
> 
> I think there are 3 main reasons one might put constraints on the trait side:
> 
>     1. Declutter the consumer signatures. e.g. If you are going to have `Foo + Bar + Serialize + DeserializeOwned`, bundling those together to clean up code.
> 
>     2. Enforcing domain level concepts at the interface level. i.e. you know how the domain code is going to be used and you are preemptively enforcing that something meets requirements of what the implementations are going to need.
> 
>     3. Anticipating actual requirements for the trait. If a type is used in an `async` function, it will often need the `Send` and/or `Sync`.
> 
> 
> I think 1 should be used very carefully. It adds a lot of indirection, so I'd prefer the bundled trait to be as close to the implementation as possible. If it is used liberally around the codebase, the trait should be named very precisely!
> 
> I think 2 should be avoided. Traits should be very specific to the their requirements and the more we try to anticipate the implementation details the more we couple concepts together in confusing ways. So if a function needs a method from one trait but not another, we should only constrain for the trait method it needs.
> 
> I think 3 is fine, but we should ask ourselves if we are really doing # 1 or # 2. The line between these can be fuzzy.
> 
> I think the tradeoff in general of pushing the constraints closer to the `impl`emnation, is a type can have many `impl` blocks scattered around the code base. This results in a new practical problem of organizing the `impl` blocks in a coherent way--which I know no standard for.

You're raising a really good point, and I agree interface level trait bounds should be added with caution. In this particular scenario, I do believe we are in the third case you outline. I.e. I believe **all legitimate implementations** of Mappable **should uphold this property**.

Note that this is a belief, not necessarily an absolute truth. I have two main reasons for my conviction however:

1. The fact that we already have the `ToOwned` bound in the interface indicates that the reciprocal `Borrow` makes sense.
2. This property holds for all known tables. *Note however, that I had to implement `Borrow` for some newtypes* - I think the lack of the `Borrow` implementation on those is an oversight in the original implementation - which we could catch by enforcing this trait bound directly on the interface level. Without that bound, we wouldn't discover this missing bound until hitting this limitation in consumer code.

That said, the main benefit of this PR is the implementation of the missing `Borrows`. With these in place, I wouldn't have any issues skipping the bound on the `Mappable` type and keep this to consumer code for the sake of being cautious. In this scenario however, I would question the existing `From<ToOwned::Owned>` bound for the same reasons and perhaps relax that requirement on the trait.

xgreenx (approved):

PR #963, opened 2025-07-21
Allow setting ECAL for predicate exection

Dentosal:
See https://github.com/FuelLabs/fuel-vm/pull/945#issuecomment-2829271997

Adds `ECAL` state parameter to some predicate-related APIs. Allows `ECAL` instruction during predicate execution. Exposes `Interpreter::context` that can be used by the `ECAL` handler to determine if it's inside a predicate.

## Checklist

### Before requesting review

### DIFF
--- a/fuel-vm/src/checked_transaction.rs
+++ b/fuel-vm/src/checked_transaction.rs
@@ -340,7 +340,7 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
             consensus_params,
             memory,
             storage,
-            NotSupportedEcal::default(),
+            NotSupportedEcal,
         )
     }
 
@@ -482,13 +482,7 @@ pub trait EstimatePredicates: Sized {
         memory: impl Memory,
         storage: &impl PredicateStorageRequirements,
     ) -> Result<(), CheckError> {
-        Self::estimate_predicates_ecal(
-            self,
-            params,
-            memory,
-            storage,
-            NotSupportedEcal::default(),
-        )
+        Self::estimate_predicates_ecal(self, params, memory, storage, NotSupportedEcal)
     }
 
     /// Estimates predicates of the transaction in parallel.
@@ -503,7 +497,7 @@ pub trait EstimatePredicates: Sized {
             params,
             pool,
             storage,
-            NotSupportedEcal::default(),
+            NotSupportedEcal,
         )
         .await
     }
@@ -2139,7 +2133,7 @@ mod tests {
             )
             .unwrap()
             // Sets Checks::Predicates
-            .check_predicates(&check_predicate_params, MemoryInstance::new(), &EmptyStorage, NotSupportedEcal::default())
+            .check_predicates(&check_predicate_params, MemoryInstance::new(), &EmptyStorage, NotSupportedEcal)
             .unwrap();
         assert!(
             checked
--- a/fuel-vm/src/checked_transaction.rs
+++ b/fuel-vm/src/checked_transaction.rs
@@ -44,8 +44,10 @@ pub use types::*;
 use crate::{
     error::PredicateVerificationFailed,
     interpreter::{
+        EcalHandler,
         Memory,
         MemoryInstance,
+        NotSupportedEcal,
     },
     pool::VmMemoryPool,
     prelude::*,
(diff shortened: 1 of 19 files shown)


fuel-vm/src/checked_transaction.rs
```diff
@@ -332,11 +334,54 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
...
+            block_height,
+            consensus_params,
+            memory,
+            storage,
+            NotSupportedEcal,
+        )
+    }
+
+    /// Returns transaction that passed all `Checks`.
+    fn into_checked_ecal<Ecal>(
```
MitchTurner, line 348: nit: Do we want these `_ecal` methods to be on a separate ext trait? for both the `checked` and `estimate` traits? Maybe it doesn't matter.

fuel-vm/src/tests/predicate.rs
```diff
@@ -292,8 +306,40 @@ async fn predicate() {
...
+        .await
+    );
+    assert!(
+        !execute_predicate(predicate.iter().copied(), wrong_data, 0, NotSupportedEcal)
+            .await
+    );
+}
+
+#[tokio::test]
+async fn predicate_allows_ecal() {
```
MitchTurner, line 325: Are we missing a test showing that a predicate will fail if it using ecal and ecal isn't supported?

Also maybe a test that shows that the ecal is executed as well? `NoopEcal` shows that the vm won't panic, but it doesn't show that it's executing the ecal.

Dentosal: Sure. Fixed in e5ba1d7d7.

fuel-vm/src/checked_transaction.rs
MitchTurner: It was just a subtle interface segregation example, since the `__ecal` methods won't be used by the same consumers. i.e. it is coupling the `checked` and `estimate` interfaces with debug features.

It's like how you don't want to include test code in production. I wouldn't want to include test methods on a trait without a `#[cfg(test)]` or `#[cfg(feature=test-helpers)]`. Since this is a runtime feature we don't have that luxary. 

Again this is a "nit", so I'm not going to push too hard on this.

> I'd rather consider just removing the non _ecal methods (and not appending _ecal from the name).

Yeah. Good point. And at that point we might as well have a default `impl` for the `NotSupportedEcal`. 

I think this is noop.

fuel-vm/src/tests/predicate.rs
```diff
@@ -292,8 +306,60 @@ async fn predicate() {
...
+        op::ret(0x10),
+    ];
+    assert!(
+        !execute_predicate(predicate.iter().copied(), vec![], 0, NotSupportedEcal).await
+    );
+}
+
+#[tokio::test]
+async fn predicate_allows_ecal() {
+    let accept_data = 1 as Word;
```
MitchTurner, line 337:
```suggestion
    // given
    let accept_data = 1 as Word;
```

fuel-vm/src/tests/predicate.rs
```diff
@@ -292,8 +306,60 @@ async fn predicate() {
...
+    let accept_data = accept_data.to_be_bytes().to_vec();
+
+    let decline_data = 0 as Word;
+    let decline_data = decline_data.to_be_bytes().to_vec();
+
+    let predicate = [
+        // Load data
+        op::gtf_args(0x11, RegId::ZERO, GTFArgs::InputCoinPredicateData),
+        op::lw(0x10, 0x11, 0),
+        // Incremnet by one
```
rymnc, line 347:
```suggestion
        // Increment by one
```

rymnc (approved):
generally lgtm, i'm concerned with the large amount of generics on these functions, any way we can convert into a builder pattern with one generic at a time? will prevent usage of so many turbofishes everywhere in the code (can be discussed later / follow up PR)

fuel-vm/src/checked_transaction.rs
```diff
@@ -332,11 +334,54 @@ pub trait IntoChecked: FormatValidityChecks + Sized {
...
+        self,
+        block_height: BlockHeight,
+        consensus_params: &ConsensusParameters,
+        ecal_state: Ecal,
```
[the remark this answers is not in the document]
Dentosal: The type determines the handler. The value is the state of that handler.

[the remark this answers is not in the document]
MitchTurner: Shouldn't the constraint be EcalState then?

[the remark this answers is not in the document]
MitchTurner: >as Rust traits are usually verb-named if possible

I'm not bought into this philosophy. This makes sense for EXT type traits like `Clone`, `Read`, `Deref`, but the purpose of this code is to abstract logic out of the business code and we should make it eagerly available to the reader what that logic should be. I think `EcalHandler` is the correct choice here.

>No, the constraint is named correctly. It's something that can handle ECALs. 

I agree. I think `handler` makes sense. I just don't agree with calling a ECAL handler `ecal_state`. There exists `impl EcalHandler`s that don't have any state. That's why I prefer `ecal_handler` here. If you think that is misleading, then that's when we should consider changing the trait name from `EcalHandler` to something that matches the _expected logic_ better (I have no reason to believe it is wrong).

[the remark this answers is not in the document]
Dentosal: The logic here is that this is the state the ecal handler is initialized to.

In any case, I think it's fine either way. I renamed the parameter to `ecal_handler` in 6138bbf04.

rymnc (approved):

PR #979, opened 2025-11-21
feat: Add DeserializeForwardCompatible trait for forward-compatible d…

its-saeed:
### Problem
When old nodes receive transactions with new policy types, they cannot preserve unknown policies during deserialization/modification/reserialization, breaking transaction signatures.

### Solution
Implement `DeserializeForwardCompatible` and `SerializeForwardCompatible` traits that:
- Capture unknown policy fields in metadata during deserialization
- Preserve unknown policies when reserializing
- Enable roundtrip serialization

### Key Changes
- **`fuel-types/src/canonical.rs`**: New traits for forward-compatible serialization
- **`fuel-tx/src/transaction/policies.rs`**: Implementation for `Policies` struct
- Unknown policies stored as `Vec<(bit_position, value)>` in metadata
- Values encoded in strict bit order (0-31) for signature compatibility

## Checklist

### Before requesting review

### DIFF
--- a/fuel-types/src/canonical.rs
+++ b/fuel-types/src/canonical.rs
@@ -1037,5 +1037,4 @@ mod tests {
         assert_eq!(obj.flag, 0); // Masked to known bits
         assert!(metadata.has_extra_data);
     }
-
 }
--- a/fuel-tx/src/transaction.rs
+++ b/fuel-tx/src/transaction.rs
@@ -771,9 +771,7 @@ impl Transaction {
             Self::Upload(tx) => {
                 tx.to_bytes_forward_compatible(&metadata.policies_metadata)
             }
-            Self::Blob(tx) => {
-                tx.to_bytes_forward_compatible(&metadata.policies_metadata)
-            }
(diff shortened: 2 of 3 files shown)


fuel-types/src/canonical.rs
```diff
@@ -182,6 +291,90 @@ pub trait Deserialize: Sized {
...
+///     fn decode_static_forward_compatible<I: Input + ?Sized>(
+///         buffer: &mut I,
+///     ) -> Result<(Self, Self::Metadata), Error> {
+///         let known_field = u32::decode(buffer)?;
+///         let metadata = MyMetadata { had_unknown_data: false };
+///         Ok((Self { known_field }, metadata))
+///     }
+/// }
+/// ```
+pub trait DeserializeForwardCompatible: Sized {
```
xgreenx: The definition looks good, but I don't see the implementation for all types supported by canonical deserialization. So after this change, we should be able to do `Transaction::from_bytes_backward/forward_compatible`

PR #981, opened 2025-12-19
fix: skip unknown policies during deserialization

### LINKED ISSUE #978: During deserialization of the Policies in the case of unknown policies skip them, instead of failing deserialization

anuragchvn-blip:
Fixes #978

## Problem
Deserialization fails when encountering unknown policy bits, breaking forward compatibility with future policy versions.

## Solution
- Changed `from_bits()` to `from_bits_truncate()` in canonical deserialization
- Added `unknown_policies` HashMap to store unknown policy values
- Added `raw_bits` field to preserve original bits for serialization
- Decode and re-serialize unknown policies in correct bit order

## Changes
- `Policies` struct now stores unknown policies and raw bits
- `decode_static`: uses `from_bits_truncate()` instead of `from_bits()`
- `decode_dynamic`: reads values for both known and unknown policies
- `encode_dynamic`: writes values for all policies in bit order
- Manual `Hash` implementation (HashMap doesn't derive Hash)

## Tests
- All 285 existing tests pass
- Added 6 new tests covering:
  - Deserialization with unknown bits
  - Round-trip preservation of unknown policies
  - Mixed known/unknown policy handling
  - Serialization order maintenance
  - Size calculation accuracy

### DIFF
--- a/fuel-tx/src/transaction/policies.rs
+++ b/fuel-tx/src/transaction/policies.rs
@@ -15,6 +15,7 @@ use fuel_types::{
         Serialize,
     },
 };
+use hashbrown::HashMap;
 
 #[cfg(feature = "random")]
 use rand::{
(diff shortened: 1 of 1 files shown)


xgreenx:
Thank you for your PR, but we've already start on this issue in this PR: https://github.com/FuelLabs/fuel-vm/pull/979

PR #1007, opened 2026-04-27
Charge for reads in SCLR properly

Dentosal:
`SCLR` could insert unlimited `None` entries into the slot cache for free. This could be used as a DoS attack. This PR makes `SCLR` charge for the storage cache slots properly.

## Checklist
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior

### Before requesting review

### DIFF
--- a/fuel-vm/src/interpreter/executors/opcodes_impl.rs
+++ b/fuel-vm/src/interpreter/executors/opcodes_impl.rs
@@ -2808,25 +2808,15 @@ where
         let num_slots = crate::convert::to_usize(interpreter.registers[r_num_slots])
             .ok_or(PanicReason::TooManySlots)?;
         let contract_id = interpreter.internal_contract()?;
-        // Charge cold-access gas for slots not already in the cache. Without
-        // this, SCLR could flood the cache with `None` entries for slots that
-        // were never read, bypassing the cold-read fee that normally gates new
-        // cache allocations and enabling a memory-exhaustion attack.
-        for key in key_range(start_key, num_slots) {
-            let key = key.ok_or(PanicReason::TooManySlots)?;
-            if !interpreter
-                .storage_slot_cache
-                .contains_key(&(contract_id, key))
-            {
-                interpreter.dependent_gas_charge(
-                    interpreter
-                        .gas_costs()
-                        .storage_read_cold()
-                        .map_err(PanicReason::from)?,
-                    0, // slot is absent — no data bytes, only the base fee applies
-                )?;
-            }
-        }
+        // Charge gas for setting storage cache slots to `None`
+        interpreter.gas_charge(
+            interpreter
+                .gas_costs()
+                .storage_read_hot()
+                .map_err(PanicReason::from)?
+                .base()
+                .saturating_mul(num_slots as u64),
+        )?;
         interpreter.storage_clear_slot_range(contract_id, start_key, num_slots)?;
         let (SystemRegisters { pc, .. }, _) = split_registers(&mut interpreter.registers);
         inc_pc(pc);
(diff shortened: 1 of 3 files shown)


MitchTurner (commented):
The code lgtm, but I handed it to Codex and it gave me two pieces of feedback.
> The fix makes SCLR charge for touching uncached slots, but it always charges as if the slot were empty. If the slot actually has data in it, a normal cold read costs more than that. So the fix is incomplete: some cold clears are still too cheap.

> The new code starts charging while it walks the key range. If the range later turns out to be invalid because it overflows, the instruction now fails after already spending some gas. Before this change, it failed first and charged nothing.

fuel-vm/src/interpreter/executors/opcodes_impl.rs
```diff
@@ -2808,6 +2808,25 @@ where
         let num_slots = crate::convert::to_usize(interpreter.registers[r_num_slots])
             .ok_or(PanicReason::TooManySlots)?;
         let contract_id = interpreter.internal_contract()?;
+        // Charge cold-access gas for slots not already in the cache. Without
+        // this, SCLR could flood the cache with `None` entries for slots that
+        // were never read, bypassing the cold-read fee that normally gates new
+        // cache allocations and enabling a memory-exhaustion attack.
+        for key in key_range(start_key, num_slots) {
+            let key = key.ok_or(PanicReason::TooManySlots)?;
+            if !interpreter
```
xgreenx: It is not clear to me why we want to charge for a cold read, only if the item doesn't exist. By doing this check(contains_key), you now require us to look into the real storage, which will trigger a cold read. But you do that read for each key. So if you want to call `contains_key`, you should charge cold or hot for any value.

But I am not even sure why we need to call `contains_key` in the first place. I think we can just clean all of them and charge for a hot read to ensure the attacker can't abuse the free loop.

Dentosal: I was mostly mirroring what `SCWQ` does and now that you point it out, charging for a cold read makes no sense here, as we never read the underlying storage. 

> It is not clear to me why we want to charge for a cold read, only if the item doesn't exist.

We need to charge for inserting `None` into the storage cache. But cold read is certainly wrong thing to charge for here, when the real thing is a memory allocation.

> By doing this check(contains_key), you now require us to look into the real storage, which will trigger a cold read.

We never look into the real storage here.

> But I am not even sure why we need to call `contains_key` in the first place. I think we can just clean all of them and charge for a hot read to ensure the attacker can't abuse the free loop.

If by clean you mean removing them from the storage cache, this is a breaking change. We currently guarantee that when the contents of a slot are known, reading from it is always a hot read. The state of a slot after clearing it is known to be `None` and not absent. Thus, we need to cache the cleans here.

We could indeed relax this requirement instead.

xgreenx: If we remove values, we don't care whether they existed before. We just place `None` into the in memory transaction. Fro the gas point of view, it is similar to hot look up, so I think we just should carge hot read price for each key which we remove.

Dentosal: Like this? b735aa86a

xgreenx: Yep. I also though that the ide a of the change was to check that `let key = key.ok_or(PanicReason::TooManySlots)?;` will not overflow. Or is it related to another bug?=D I forgot, sorry

Dentosal: That's for another bug. The fix is already in https://github.com/FuelLabs/fuel-vm/pull/1004.

fuel-vm/src/interpreter/executors/opcodes_impl.rs
```diff
@@ -2808,6 +2808,15 @@ where
...
         let contract_id = interpreter.internal_contract()?;
+        // Charge gas for setting storage cache slots to `None`
+        interpreter.gas_charge(
+            interpreter
+                .gas_costs()
+                .storage_read_hot()
+                .map_err(PanicReason::from)?
+                .base()
+                .saturating_mul(num_slots as u64),
+        )?;
```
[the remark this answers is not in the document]
Dentosal: 94b67a10e

xgreenx (approved):

Dentosal:
This seems unnecessary, we already charge for this.

